PR #1, opened 2024-11-14
add skeleton api crate

neonphog:
Consider before approving:

- are we happy with the crate layout?
- are we happy with the `kitsune2_` crate naming convention?
- are we happy with the concept itself of having a basic types + trait definition crate separate from the `kitsune2` crate?


crates/kitsune2_api/src/lib.rs
jost-s, line 1: Why are traits in a separate crate? So that people can implement their own kitsune2 without needing the kitsune2 crate?

jost-s (commented):
Fine with `kitsune2_`

crates/kitsune2_api/Cargo.toml
ThetaSinner, line 1: I'd really like it if we didn't name the directories after the crate. Could this be an `api` directory, and then I'm happy with the crate name 👍🏻

crates/kitsune2_api/src/lib.rs
ThetaSinner: If we get this right, only the equivalent of `holochain_p2p` will need to depend on the implementation and anything else in Holochain that needs to call Kitsune will be able to just depend on this crate.

That might mean putting the types in here too, which seems reasonable I think

PR #2, opened 2024-11-14
Add basic pr checks

ThetaSinner:


ThetaSinner:
This won't work until there are Rust crates. PR #1 needs to merge first

ThetaSinner:
Don't want to fail fast, and only run docs once on Linux

neonphog (approved):

PR #3, opened 2024-11-14
simplified hash types for kitsune2

neonphog:
The kitsune2 spike repo had a dyn trait for these Ids, but that proved difficult to work with. Here, I'm suggesting instead that we go for newtypes around the `Bytes` type which has already solved the shallow clone problem, so we shouldn't have to worry about memory usage if we are using clones of these in hash maps for example.

Another awkwardness with the holochain/kitsune1 interface was debugging in kitsune1 printed out things in hex and without the prefix that holochain uses, so it was difficult to compare the results.

Here, I've gone with a default of printing out the bytes in base64, and also allowing whatever crate is implementing kitsune2 to override the debug/display functions, so in holochain integration, we can have it print out with the prefix and location suffix, or whatever we choose.

Finally, I've altered the location calculation algorithm to something that is far less computationally problematic, and which would be much easier for clients to implement directly. I don't believe there are any downsides to this change, but please correct me if there are!

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+max_width = 80
--- a/crates/api/src/id.rs
+++ b/crates/api/src/id.rs
@@ -1,0 +1,169 @@
+//! Types dealing with data identity or hashing.
+
+macro_rules! imp_deref {
+    ($i:ty, $t:ty) => {
+        impl std::ops::Deref for $i {
+            type Target = $t;
+
+            fn deref(&self) -> &Self::Target {
+                &self.0
+            }
+        }
+    };
+}
+
+/// Base data identity type meant for newtyping.
+/// You probably want [AgentId] or [OpId].
+///
+/// In Kitsune2 these bytes should ONLY be the actual hash bytes
+/// or public key of the identity being tracked, without
+/// prefix or suffix.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Id(pub bytes::Bytes);
+
+imp_deref!(Id, bytes::Bytes);
+
+impl Id {
+    /// Get the location u32 based off this Id.
+    //
+    // Holochain previously would re-hash the hash, and then
+    // xor to shrink down to a u32. This extra step is not needed
+    // and does not provide any benefit. One extra hash step does
+    // not prevent location farming, and if the original hash was
+    // distributed well enough, re-hashing it again doesn't improve
+    // distribution.
+    pub fn loc(&self) -> u32 {
+        let mut out = [0_u8; 4];
+        let mut i = 0;
+        for c in &self.0 {
+            out[i] ^= c;
+            i += 1;
+            if i > 3 {
+                i = 0;
+            }
+        }
+        u32::from_le_bytes(out)
+    }
+}
+
+/// The function signature for Id display overrides.
+pub type DisplayCb =
+    fn(&bytes::Bytes, &mut std::fmt::Formatter<'_>) -> std::fmt::Result;
+
+/// The default display function encodes the Id as base64.
+/// This makes debugging so much easier than rust's default of decimal array.
+fn default_display(
+    b: &bytes::Bytes,
+    f: &mut std::fmt::Formatter<'_>,
+) -> std::fmt::Result {
+    use base64::prelude::*;
+    f.write_str(&BASE64_URL_SAFE_NO_PAD.encode(b))
+}
+
+#[inline(always)]
+fn display(
+    b: &bytes::Bytes,
+    f: &mut std::fmt::Formatter<'_>,
+    l: &std::sync::OnceLock<DisplayCb>,
+) -> std::fmt::Result {
+    l.get_or_init(|| default_display)(b, f)
+}
+
+static AGENT_DISP: std::sync::OnceLock<DisplayCb> = std::sync::OnceLock::new();
+
+/// Identifies an agent to be tracked as part of a Kitsune space.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct AgentId(pub Id);
+
+imp_deref!(AgentId, Id);
+
+impl std::fmt::Display for AgentId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        display(&self.0 .0, f, &AGENT_DISP)
+    }
+}
+
+impl std::fmt::Debug for AgentId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        display(&self.0 .0, f, &AGENT_DISP)
+    }
+}
+
+impl AgentId {
+    /// Set the display/debug implementation for AgentId for the duration
+    /// of this process. Note, if anything was printed earlier, the
+    /// default impl will have been set and cannot be changed.
+    /// Returns false if the default was unable to be set.
+    pub fn set_global_display_callback(cb: DisplayCb) -> bool {
+        match AGENT_DISP.set(cb) {
+            Ok(_) => true,
+            Err(_) => false,
+        }
+    }
+}
+
+static SPACE_DISP: std::sync::OnceLock<DisplayCb> = std::sync::OnceLock::new();
+
+/// Identifies a space to be tracked by Kitsune.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SpaceId(pub Id);
+
+imp_deref!(SpaceId, Id);
+
+impl std::fmt::Display for SpaceId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        display(&self.0 .0, f, &SPACE_DISP)
+    }
+}
+
+impl std::fmt::Debug for SpaceId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        display(&self.0 .0, f, &SPACE_DISP)
+    }
+}
+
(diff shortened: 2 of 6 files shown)


pdaoust:
This feels like a massive change (and I presume it'll trickle upward and Holochain will adopt it for some sorts of `EntryId` and `ActionId`), but from a developer experience point of view, I'm thrilled. (I've also been adopting "identifier" in my language rather than "hash", because it's annoying to explain to people that the hash function for `AgentPubKey` is just the identity function!)

neonphog:
> This feels like a massive change (and I presume it'll trickle upward and Holochain will adopt it for some sorts of EntryId and ActionId), but from a developer experience point of view, I'm thrilled.

While I'd love to do some retooling on the holochain side too, I imagine the first integration is going to be less invasive. It will be fairly straight-forward to translate between these types and the existing Hash types in holochain.

> (I've also been adopting "identifier" in my language rather than "hash", because it's annoying to explain to people that the hash function for AgentPubKey is just the identity function!)

I prefer the explanation that the algorithm to derive a keypair from seed entropy is very akin to hashing it : )

jost-s (commented):
So great that hash display is configurable. Is it necessary to lock this once it's set? Just in case it's simpler to not lock it, not that I see a use case where you'd want to change it while running a test or Holochain.

.rustfmt.toml
```diff
@@ -0,0 +1 @@
+max_width = 80
```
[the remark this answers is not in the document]
neonphog: I know this may not be popular, and I'm fine to remove it.

But, I always code in 80 character width terminals... and it's very annoying having wrapping lines...

crates/api/src/id.rs
```diff
@@ -0,0 +1,160 @@
...
+/// Base data identity type meant for newtyping.
+/// You probably want [AgentId] or [OpId].
+///
+/// In Kitsune2 these bytes should ONLY be the actual hash bytes
+/// or public key of the identity being tracked, without
+/// prefix or suffix.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Id(pub bytes::Bytes);
+
+imp_deref!(Id, bytes::Bytes);
```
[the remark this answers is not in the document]
neonphog: I don't understand the sudden dislike of deref in the rust community. But setting that entirely aside, I think newtypes (in addition to smart pointers) are exactly what deref is for. You want the functions for your inner type to be exposed.

neonphog:
> Is it necessary to lock this once it's set?

This is the simplest, most efficient way to have a thread-safe setter. If we want to re-set it we'd need a mutex which will slow down the call time.

> not that I see a use case where you'd want to change it while running a test or Holochain

Actually, whether we allow re-setting this or not, tests are a case where we CAN NOT do it... since multiple tests run in the same process, they would stamp on each other's implementations.

.rustfmt.toml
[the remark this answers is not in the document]
jost-s: As long as the formatter does it, I don't care. =)

crates/api/src/id.rs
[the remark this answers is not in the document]
jost-s: I don't have a history with it, so I'm asking with openness. delegate seems wordier and more difficult to read, and deref is more practical, I agree.

crates/api/src/id.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl Id {
+    /// Get the location u32 based off this Id.
```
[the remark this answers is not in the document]
neonphog: done: https://github.com/holochain/kitsune2/pull/3/commits/22ede7c53f8db5aafcd72979fe914f4f49feeffe

jost-s (changes requested):
Thanks for the comment. Now that I look at it again, I think `loc` needs some unit tests.

neonphog:
> Thanks for the comment. Now that I look at it again, I think loc needs some unit tests.

done: https://github.com/holochain/kitsune2/pull/3/commits/0ff84ab4abc544af5fc0ab49b5ef86a6c3336aa3

jost-s (dismissed):
Interesting, I would have thought the locations would progress a bit more linear and predictable.

neonphog:
> Interesting, I would have thought the locations would progress a bit more linear and predictable.

Yeah, it's a little weird looking at it with the fixture data I added in. The theory is that the starting hashes should be indistinguishable from noise at the bit level, and the same should be true for the bits at the u32 location level too.

neonphog:
Sorry to do this after you already approved @jost-s, but I needed to add some from and serde impls for the Ids, to make the agent PR work, but it seemed weird to do them there, so I did it here.

The possibly controversial bit is using base64 as the default encoding for the Ids. I certainly need this for the json version of the agent info encoding... but we could go either way for any message pack encoding we'd do.

My thoughts on this are that the hashes are not very big so the base64 doesn't add much overhead, and we don't accidentally fall into that silly serde trap where it encodes as an array of bytes instead of a byte array...

mattyg:
> The possibly controversial bit is using base64 as the default encoding for the Ids. I certainly need this for the json version of the agent info encoding... but we could go either way for any message pack encoding we'd do.

Why not just send an array of numbers for the json encoding?

> My thoughts on this are that the hashes are not very big so the base64 doesn't add much overhead, and we don't accidentally fall into that silly serde trap where it encodes as an array of bytes instead of a byte array...

But do we add a new trap of accidentally using the wrong base64 alphabet when decoding?

neonphog:
> Why not just send an array of numbers for the json encoding?

This is slow in rust/serde because it has to transcode the bytes one at a time as opposed to internally being able to do a memcopy. Of course this is less of a problem with these smaller hashes.

But I still think debug readability is important. When you see a dump of something that is an array of decimal numbers, it's much harder to immediately know there is a difference as opposed to when you see to base64 strings.

neonphog:
> But do we add a new trap of accidentally using the wrong base64 alphabet when decoding?

Haha, I don't, because I always use url safe base64... but that is certainly something to consider.

crates/api/src/id.rs
```diff
@@ -0,0 +1,288 @@
...
+    // xor to shrink down to a u32. This extra step is not needed
+    // and does not provide any benefit. One extra hash step does
+    // not prevent location farming, and if the original hash was
+    // distributed well enough, re-hashing it again doesn't improve
+    // distribution.
+    pub fn loc(&self) -> u32 {
+        let mut out = [0_u8; 4];
+        let mut i = 0;
+        for c in &self.0 {
+            out[i] ^= c;
```
ThetaSinner, line 80: I would find this clearer as a ranged loop with indexing on the modulo of the index. 

```
for i in 0..self.0.len() {
  out[i % 4] ^= self.0[i]
}
```

Or something along those lines, I haven't checked that code runs

crates/api/src/id.rs
```diff
@@ -0,0 +1,288 @@
...
+    ///
+    /// The remaining 4 bytes are then interpreted as a little-endian u32.
+    //
+    // Holochain previously would re-hash the hash, and then
+    // xor to shrink down to a u32. This extra step is not needed
+    // and does not provide any benefit. One extra hash step does
+    // not prevent location farming, and if the original hash was
+    // distributed well enough, re-hashing it again doesn't improve
+    // distribution.
+    pub fn loc(&self) -> u32 {
```
ThetaSinner, line 76: This is a change from the location being a host decision to being part of Kitsune. 

That would prevent a host implementation from doing things like colocating ops for a chunked up document in the same location. I know that we generally want to keep the DHT balanced but this does seem like something the host should decide.

crates/api/src/id.rs
```diff
@@ -0,0 +1,288 @@
+//! Types dealing with data identity or hashing.
+
+macro_rules! imp_deref {
```
ThetaSinner, line 3: `derive_more` can handle this and the `1.0` library has been split up so that you can just import what you need

crates/api/src/id.rs
```diff
@@ -0,0 +1,288 @@
...
+            assert_eq!(
+                u,
+                Id(bytes::Bytes::copy_from_slice(&u.to_le_bytes())).loc()
+            );
+        }
+    }
+
+    #[test]
+    fn loc_fixtures() {
+        const F: &[(&[u8], u32)] = &[
```
ThetaSinner, line 257: Looks like a [test-case](https://crates.io/crates/test-case)

crates/api/src/id.rs
neonphog: That is true. This case was supported by the dyn in the spike... I could solve this in the same way as the display function. Do you think that is reasonable @ThetaSinner ? If so, the question then becomes, do we want to support different loc derivation functions for each Id type (Agent, Space, etc)? Or, should we just have a single one on the base Id type?

crates/api/src/id.rs
neonphog: We agreed to go without this dep for now, but we can always choose differently in the future.

crates/api/src/id.rs
neonphog: We agreed to go without this dep for now, but we can always choose differently in the future.

crates/api/src/id.rs
neonphog: Gave it a first shot of configurable once at the whole process level here: https://github.com/holochain/kitsune2/pull/3/commits/16a111dcd6361f773043731635e938ee7e830ac6

ThetaSinner: Let's start with one callback. I'm assuming that in order to apply logic beyond some simple algorithm, you'd need to look up the hashes and group them in some meaningful way. In which case you don't need to know their types unless you expect clashes... which would be surprising. So yes, this looks good!

ThetaSinner (approved):

jost-s (approved):
I'm good with this

PR #4, opened 2024-11-14
serde + timestamp type

neonphog:
DEPENDS ON #3 

- Adds serde, because we need to be able to serialize / deserialize our types
- Adds a basic Timestamp type without all the complexity that somehow ended up in our legacy timestamp type


crates/api/src/timestamp.rs
```diff
@@ -0,0 +1,69 @@
...
+#[serde(transparent)]
+pub struct Timestamp(i64);
+
+impl Timestamp {
+    /// Construct a new timestamp of "now".
+    pub fn now() -> Self {
+        Self(
+            std::time::SystemTime::now()
+                .duration_since(std::time::SystemTime::UNIX_EPOCH)
+                .expect("invalid system clock")
```
jost-s, line 30: This doesn't happen on functional machines, right?!

jost-s (approved):
LGTM

crates/api/src/timestamp.rs
neonphog: It might be possible to hack your OS to be at a time before the epoch... but I imagine so much other software would crash that the system would essentially be unusable.

crates/api/src/timestamp.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    fn add(self, rhs: std::time::Duration) -> Self::Output {
+        Timestamp(self.0 + rhs.as_micros() as i64)
+    }
+}
+
+impl From<std::time::SystemTime> for Timestamp {
+    fn from(t: std::time::SystemTime) -> Self {
+        Self(
+            t.duration_since(std::time::SystemTime::UNIX_EPOCH)
```
ThetaSinner, line 57: Can the from be used in the constructor? Seems a little weird but it's the same code :)

ThetaSinner (approved):
That old Timestamp type is set up to play nicely with JavaScript on Holochain's public interface. it never belonged in Kitsune and I think the Kitsune type just got updated to support all kinds of things that it didn't need to for Kitsune's use. This looks good!

crates/api/src/timestamp.rs
neonphog: ha! good point. done: https://github.com/holochain/kitsune2/pull/4/commits/ebf79e717a61f4193f5297bfd6a734c6072dd386

ThetaSinner (approved):

PR #6, opened 2024-11-15
create the bootstrap spec

neonphog:
Create a spec for the new bootstrap server API. All feedback welcome. Pay attention to the transition from `//!` to `///` comments ; )

### DIFF
--- a/crates/boot_srv/Cargo.toml
+++ b/crates/boot_srv/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "kitsune2_boot_srv"
+version = "0.0.1-alpha"
+description = "p2p / dht communication framework api"
+license = "Apache-2.0"
+homepage = "https://github.com/holochain/kitsune2"
+documentation = "https://docs.rs/kitsune2_boot_srv"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "holo", "p2p", "dht", "networking"]
+categories = ["network-programming"]
+edition = "2021"
+
+[dependencies]
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -129,7 +129,8 @@
 /// and when any entries are removed, the indexes of any items after them
 /// will be decrimented.
 ///
-/// - The server SHOULD provide a configurable per-space max info count.
+/// - The server SHOULD allow setting the maximum number of infos that will be
+///   allowed in a single space.
 ///   For the duration of this document, that value will be called MAX_INFOS.
 ///   It is recommended to default that value to `32` to keep the maximum
 ///   response size from a request to around 32KiB.
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -129,7 +129,7 @@
 /// - If the store count is >= MAX_INFOS, the server MUST delete the entry
 ///   at MAX_INFOS / 2 and then push the info onto the stack.
 ///
-/// #### 3. Rate Limiting
+/// #### 4. Rate Limiting
 ///
 /// Info count within a space is limited by the storage strategy above. But
 /// space count within a server is unbounded. While servers will likely want
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -159,5 +159,18 @@
 ///   which result in space creation, and error with 429 if that frequency
 ///   is beyond a limit.
 /// - A server MAY make this limit configurable.
+///
+/// #### 5. Client Recommendations
+///
+/// - In the case of server unreachable or 500 error, a client should use
+///   an exponential backoff strategy to avoid stressing the server when
+///   it does come back online.
+/// - A client should poll the bootstrap server on an interval >= 5 minutes
+///   to balance getting new infos to avoid partitioning without worry
+///   of hitting the rate limit.
+/// - A client, in general, should set expiration to 20 minutes beyond
+///   the info creation time.
+/// - A client, in general, should re-publish a new info on a 15 minute
+///   interval.
 #[cfg(doc)]
 pub mod spec {}
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -22,7 +22,7 @@
 /// - `Base64Space` - base64UrlNoPad safe encoded string space id.
 /// - `Base64Sig` - base64UrlNoPad safe encoded string crypto signature.
 /// - `Json` - string containing json that can be decoded.
-/// - `I64` - string contaning an i64 number.
+/// - `I64` - string containing an i64 number.
 ///
 /// ```text
 /// AgentInfoSigned = { "agentInfo": Json, "signature": Base64Sig }
@@ -93,7 +93,7 @@
 ///
 /// ##### 3.1. The Future
 ///
-/// It is the intention to someday in the future to add a "trusted" strategy,
+/// It is the intention someday in the future to add a "trusted" strategy,
 /// that will be triggerd via a new api, perhaps `/registerTrust/Base64Space`.
 ///
 /// Even when that API is implemented, however, the default strategy defined
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -131,7 +131,8 @@
 ///
 /// - The server SHOULD provide a configurable per-space max info count.
 ///   For the duration of this document, that value will be called MAX_INFOS.
-///   It is recommended to default that value to `32`.
+///   It is recommended to default that value to `32` to keep the maximum
+///   response size from a request to around 32KiB.
 /// - The server SHOULD delete expired infos periodically.
 /// - If the server is already storing an agent, and a `PUT` with a newer
 ///   `createdAt` arrives, the existing entry MUST be replaced.
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -18,11 +18,18 @@
 ///
 /// #### 1. Types
 ///
-/// - `Base64Agent` - base64UrlNoPad safe encoded string agent id.
-/// - `Base64Space` - base64UrlNoPad safe encoded string space id.
-/// - `Base64Sig` - base64UrlNoPad safe encoded string crypto signature.
+/// All base64 in this spec uses: <https://datatracker.ietf.org/doc/html/rfc4648#section-5>
+///
+/// The url safe alphabet (with `-` and `_`) and no padding characters (`=`)
+/// added when encoding or required when decoding.
+///
+/// - `Base64Agent` - base64UrlSafeNoPad encoded string agent id.
+/// - `Base64Space` - base64UrlSafeNoPad encoded string space id.
+/// - `Base64Sig` - base64UrlSafeNoPad encoded string crypto signature.
 /// - `Json` - string containing json that can be decoded.
-/// - `I64` - string containing an i64 number.
+/// - `I64` - string containing an i64 number indicating the number of
+///           microseconds since the unix epoch. Since the unix epoch
+///           is canonically defined in UTC, the timestamp is also in UTC.
 ///
 /// ```text
 /// AgentInfoSigned = { "agentInfo": Json, "signature": Base64Sig }
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -1,0 +1,150 @@
+#![deny(missing_docs)]
+//! Kitsune2 boot server is an HTTP REST server for handling bootstrapping
+//! discovery of peer network reachability in p2p applications.
+//!
+//! Despite being in the kitsune2 repo, `boot_srv` and `boot_cli` do not depend
+//! on any Kitsune2 crates. This is to ensure the bootstrapping functionality
+//! is well-defined, self-contained, easily testable in isolation, and
+//! usable for projects that don't choose to make use of Kitsune2 itself.
+//!
+//! That being said, the boot server and client are designed to transfer the
+//! `AgentInfoSigned` data type defined in the `kitsune2_api` crate. Since
+//! the canonical encoding of that type is JSON, we just redefine a subset
+//! of the schema here, and only validate the parts required for bootstrapping.
+//!
+//! For additional details, please see the [spec].
+
+/// This is a documentation module containing the kitsune2_boot spec.
+///
+/// #### 1. Types
+///
+/// - `Base64Agent` - base64UrlNoPad safe encoded string agent id.
+/// - `Base64Space` - base64UrlNoPad safe encoded string space id.
+/// - `Base64Sig` - base64UrlNoPad safe encoded string crypto signature.
+/// - `Json` - string containing json that can be decoded.
+/// - `I64` - string contaning an i64 number.
+///
+/// ```text
+/// AgentInfoSigned = { "agentInfo": Json, "signature": Base64Sig }
+/// AgentInfo = {
+///   "agent": Base64Agent,
+///   "space": Base64Space,
+///   "createdAt": I64,
+///   "expiresAt": I64,
+///   "isTombstone": boolean
+/// }
+/// ```
+///
+/// Any other properties on these objects will be ignored and pass through.
+///
+/// #### 2. REST API
+///
+/// ##### 2.1. In Brief
+///
+/// ```text
+/// ErrResponse = { "error": string }
+/// OkResponse = {}
+/// ListResponse = [ AgentInfoSigned, .. ]
+/// ```
+///
+/// - `PUT /boot/Base64Space/Base64Agent`
+///   - Request Body: `AgentInfoSigned`
+///   - Response Body: `OkResponse | ErrResponse`
+/// - `GET /boot/Base64Space`
+///   - Response Body: `ListResponse | ErrResponse`
+/// - `GET /`
+///   - Response Body: `OkResponse | ErrResponse`
+///
+/// ##### 2.2. Publishing info to the boot server.
+///
+/// A `PUT` on `/boot/Base64Space/Base64Agent` with an `AgentInfoSigned`
+/// json object as the request body.
+///
+/// - The server MUST reject the request if the body is > 1024 bytes.
+/// - The server MUST reject the request if `createdAt` is not within
+///   3 minutes in either direction of the server time.
+/// - The server MUST reject the request if `expiresAt` is in the past.
+/// - The server MUST reject the request if `signature` is invalid vs
+///   the `agentInfo`.
+/// - If `isTombstone` is `true`, the server MUST delete any existing
+///   info being held.
+/// - If `isTombstone` is `false`, the server MAY begin storing the info.
+///   See section 3. on storage strategies below.
+///
+/// ##### 2.3. Listing data stored on the boot server.
+///
+/// A `GET` on `/boot/Base64Space`.
+///
+/// - The server MUST respond with a complete list of stored infos.
+/// - If there are no infos stored at this space, the server MUST return
+///   an empty list (`[]`).
+/// - The only reason a server MAY return an `ErrResponse` for this request
+///   is in the case of internal server error.
+///
+/// ##### 2.4. Health check.
+///
+/// A `GET` on `/`.
+///
+/// - The server, in general, SHOULD return `OkResponse` to this request.
+/// - The server MAY return `ErrResponse` for internal errors or some other
+///   inability to continue serving correctly.
+///
+/// #### 3. Storage Strategies
+///
+/// ##### 3.1. The Future
+///
+/// It is the intention to someday in the future to add a "trusted" strategy,
+/// that will be triggerd via a new api, perhaps `/registerTrust/Base64Space`.
+///
+/// Even when that API is implemented, however, the default strategy defined
+/// next will be used on all spaces that have not been registered with a
+/// different strategy. Therefore, we can proceed for now without consideration
+/// for any other future strategies.
+///
+/// ##### 3.2. The "Default" Storage Strategy
+///
+/// Assumptions:
+///
+/// - We don't want to store unbounded count infos in each space.
+/// - We like storing long-running reliable nodes.
+/// - We don't want to ONLY store long-running nodes to avoid eclipse scenarios.
+///
+/// The solution put forth here is to allocate half our storage space to
+/// long-running nodes, and the other half to whoever has most recently
+/// published an info. The strategy for accomplishing this is defined here:
+///
+/// Consider "the store" acting as a stack. It can be implemented in any manner,
+/// but the following strategy assumes new entries are added to the end,
+/// and when any entries are removed, the indexes of any items after them
+/// will be decrimented.
+///
+/// - The server SHOULD provide a configurable per-space max info count.
+///   For the duration of this document, that value will be called MAX_INFOS.
+///   It is recommended to default that value to `32`.
+/// - The server SHOULD delete expired infos periodically.
+/// - If the server is already storing an agent, and a `PUT` with a newer
+///   `createdAt` arrives, the existing entry MUST be replaced.
+/// - If the store count is < MAX_INFOS, the server MUST push the info
+///   onto the stack.
+/// - If the store count is >= MAX_INFOS, the server MUST delete the entry
+///   at MAX_INFOS / 2 and then push the info onto the stack.
+///
+/// #### 3. Rate Limiting
+///
+/// Info count within a space is limited by the storage strategy above. But
+/// space count within a server is unbounded. While servers will likely want
+/// to limit request counts in general, we are also going to have to take some
+/// special care with PUT requests that result in creation of a new space that
+/// was not previously tracked in order to mitigate attacks.
+///
+/// - A server SHOULD delete any spaces that no longer contain agents.
+/// - A server SHOULD decide a max count of spaces it intends to support
+///   (based on available memory or disk space for storing agents within those
+///   spaces), and then return error 429 beyond that.
+/// - A server MAY make the above max space count configurable.
+/// - A server SHOULD track the IP addresses of clients that make PUT requests
+///   which result in space creation, and error with 429 if that frequency
+///   is beyond a limit.
+/// - A server MAY make this limit configurable.
+#[cfg(doc)]
+pub mod spec {}
--- a/crates/boot_srv/src/lib.rs
+++ b/crates/boot_srv/src/lib.rs
@@ -54,17 +54,17 @@
 /// ListResponse = [ AgentInfoSigned, .. ]
 /// ```
 ///
-/// - `PUT /boot/Base64Space/Base64Agent`
+/// - `PUT /boot/<Base64Space>/<Base64Agent>`
 ///   - Request Body: `AgentInfoSigned`
 ///   - Response Body: `OkResponse | ErrResponse`
-/// - `GET /boot/Base64Space`
+/// - `GET /boot/<Base64Space>`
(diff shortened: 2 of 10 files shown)


crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///   It is recommended to default that value to `32`.
+/// - The server SHOULD delete expired infos periodically.
+/// - If the server is already storing an agent, and a `PUT` with a newer
+///   `createdAt` arrives, the existing entry MUST be replaced.
+/// - If the store count is < MAX_INFOS, the server MUST push the info
+///   onto the stack.
+/// - If the store count is >= MAX_INFOS, the server MUST delete the entry
+///   at MAX_INFOS / 2 and then push the info onto the stack.
+///
+/// #### 3. Rate Limiting
```
neonphog, line 132:
```suggestion
/// #### 4. Rate Limiting
```

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+//! the canonical encoding of that type is JSON, we just redefine a subset
+//! of the schema here, and only validate the parts required for bootstrapping.
+//!
+//! For additional details, please see the [spec].
+
+/// This is a documentation module containing the kitsune2_boot spec.
+///
+/// #### 1. Types
+///
+/// - `Base64Agent` - base64UrlNoPad safe encoded string agent id.
```
mattyg, line 21: Might be good to include the base64 spec from the rust package:
- The URL-safe alphabet (with - and _) specified in [RFC 4648](https://datatracker.ietf.org/doc/html/rfc4648#section-5).
- Don’t add padding when encoding, and require no padding when decoding.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// ##### 2.1. In Brief
+///
+/// ```text
+/// ErrResponse = { "error": string }
+/// OkResponse = {}
+/// ListResponse = [ AgentInfoSigned, .. ]
+/// ```
+///
+/// - `PUT /boot/Base64Space/Base64Agent`
```
mattyg, line 50: Should we version the url?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// - The server MUST respond with a complete list of stored infos.
+/// - If there are no infos stored at this space, the server MUST return
+///   an empty list (`[]`).
+/// - The only reason a server MAY return an `ErrResponse` for this request
+///   is in the case of internal server error.
+///
+/// ##### 2.4. Health check.
+///
+/// A `GET` on `/`.
```
mattyg, line 86: Maybe we should use a prefix `/health` for the api and leave `/` available for an fancy health monitor html page?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// ##### 2.2. Publishing info to the boot server.
+///
+/// A `PUT` on `/boot/Base64Space/Base64Agent` with an `AgentInfoSigned`
+/// json object as the request body.
+///
+/// - The server MUST reject the request if the body is > 1024 bytes.
+/// - The server MUST reject the request if `createdAt` is not within
+///   3 minutes in either direction of the server time.
+/// - The server MUST reject the request if `expiresAt` is in the past.
```
mattyg, line 66: Must reject if expiresAt >= createdAt

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - The server MUST reject the request if `signature` is invalid vs
+///   the `agentInfo`.
+/// - If `isTombstone` is `true`, the server MUST delete any existing
+///   info being held.
```
mattyg, line 70: must delete any existing AgentInfo with the keys "space" and "agent" matching the request URL components Base64Space and Base64Agent, and the key "createdAt" is being less than or equal to the request data. 

///   "space": Base64Space,
///   "createdAt": I64,
///   "expiresAt": I64,
///   "isTombstone": boolean
/// }

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - If `isTombstone` is `false`, the server MAY begin storing the info.
```
mattyg, line 71: I'd find it clearer to split create and delete into different endpoints. Then we could also drop the expiresAt field for delete requests.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///   See section 3. on storage strategies below.
+///
+/// ##### 2.3. Listing data stored on the boot server.
+///
+/// A `GET` on `/boot/Base64Space`.
+///
+/// - The server MUST respond with a complete list of stored infos.
```
mattyg, line 78: Does that include expired info that hasn't been deleted?

crates/boot_srv/src/lib.rs
mattyg, line 66: Whats the rationale is for the client requesting an expiration time? Do we want to put some configurable constraints on the expiration timespan?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// published an info. The strategy for accomplishing this is defined here:
+///
+/// Consider "the store" acting as a stack. It can be implemented in any manner,
+/// but the following strategy assumes new entries are added to the end,
+/// and when any entries are removed, the indexes of any items after them
+/// will be decrimented.
+///
+/// - The server SHOULD provide a configurable per-space max info count.
+///   For the duration of this document, that value will be called MAX_INFOS.
+///   It is recommended to default that value to `32`.
```
mattyg, line 123: why?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// The solution put forth here is to allocate half our storage space to
+/// long-running nodes, and the other half to whoever has most recently
+/// published an info. The strategy for accomplishing this is defined here:
+///
+/// Consider "the store" acting as a stack. It can be implemented in any manner,
+/// but the following strategy assumes new entries are added to the end,
+/// and when any entries are removed, the indexes of any items after them
+/// will be decrimented.
+///
+/// - The server SHOULD provide a configurable per-space max info count.
```
mattyg, line 121: Should it also provide a max expiration timespan?

mattyg, line 121: Is this saying the max info count can be configured to be different for each space?

crates/boot_srv/src/lib.rs
mattyg, line 78: We may want to include GET parameters to support pagination such as "count" and "offset"

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+//! on any Kitsune2 crates. This is to ensure the bootstrapping functionality
+//! is well-defined, self-contained, easily testable in isolation, and
+//! usable for projects that don't choose to make use of Kitsune2 itself.
+//!
+//! That being said, the boot server and client are designed to transfer the
+//! `AgentInfoSigned` data type defined in the `kitsune2_api` crate. Since
+//! the canonical encoding of that type is JSON, we just redefine a subset
+//! of the schema here, and only validate the parts required for bootstrapping.
+//!
+//! For additional details, please see the [spec].
```
mattyg, line 15: Might also include a section about recommended client implementation. I.e. expiration times, retry, scheduling of updating infos

crates/boot_srv/src/lib.rs
neonphog: The tombstone field is require for the functionality on the kitsune side. Since in the active network they cannot just delete an agent info--they might get the old one gossiped back to them again, so they need tombstone records. And having them be gossiped just the same as other infos saves complexity. Since we need that anyways, I don't want to implement a different method for the bootstrap server.

crates/boot_srv/src/lib.rs
neonphog: That's up to the server implementation. In my imaginings there will be a periodic task cleaning up expired entries, and it would also adjust the space whenever a PUT happened. So there is a chance some expired entries will be gossiped, but only just, and kitsune will be tolerant to that.

neonphog: Avoiding complexity like this, as well as needing to manage our server memory usage is the reason for limiting the infos per space. A max of 32k (plus a little overhead for commas and square brackets) will be easy to send in a single response.

crates/boot_srv/src/lib.rs
neonphog: See: https://github.com/holochain/kitsune2/pull/6#discussion_r1844999439

We need enough that you can reasonably bootstrap an app, but not so much that servers can't be maintained. We can, of course, adjust this recommendation post-testing.

crates/boot_srv/src/lib.rs
neonphog: I'm inclined towards no... we can always add a `/boot2/` if we really needed, but the structure of the agent info itself is extensible, so any changes we need to make we should be constrainable to additional fields within that.

crates/boot_srv/src/lib.rs
neonphog: Hehe, no, one single config, but the effect is at a per-space basis. : ) I'll try to figure out how to reword this.

crates/boot_srv/src/lib.rs
mattyg: I think the request or response types change the url version should update and requests in the old type explicitly transformed into the new type if both are still supported.

crates/boot_srv/src/lib.rs
mattyg: But the infos per space is configurable, what if another server has a very high limit and sends much more in a response?

crates/boot_srv/src/lib.rs
mattyg: Maybe worth saying explicitly "to keep the maximum response size from a request around 32K"

crates/boot_srv/src/lib.rs
neonphog: I assert that allowing so many bootstrap agents within a space that they cannot be sent in a single response is YAGNI. We could put documentation around the config explaining that the max response size will be `MAX_INFOS * MAX_INFO_BYTES` and to consider that before setting. Or error starting the server if that response size is beyond some threshold.

crates/boot_srv/src/lib.rs
neonphog: The main upgrade scenarios I'm considering are things like using new crypto signing algorithms, which can be accomplished with additional fields, usable through the same REST API. If you send a new info to an old server it will error, but a new server will take and respond with both new and old infos no problem.

In my experience, maintaining multiple API versions requires more overhead in terms of maintenance. Making existing APIs work with new functionality is more up-front work, but less overhead overall.

mattyg: I defer to you but I feel like explicit versioning makes clearer code paths, we get the freedom to deprecate fields if needed, and it makes it explicit that the server you're calling is running on a certain version of the spec.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// - The server, in general, SHOULD return `OkResponse` to this request.
+/// - The server MAY return `ErrResponse` for internal errors or some other
+///   inability to continue serving correctly.
+///
+/// #### 3. Storage Strategies
+///
+/// ##### 3.1. The Future
+///
+/// It is the intention to someday in the future to add a "trusted" strategy,
```
jost-s, line 96:
```suggestion
/// It is the intention someday in the future to add a "trusted" strategy,
```

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// - We don't want to store unbounded count infos in each space.
+/// - We like storing long-running reliable nodes.
+/// - We don't want to ONLY store long-running nodes to avoid eclipse scenarios.
+///
+/// The solution put forth here is to allocate half our storage space to
+/// long-running nodes, and the other half to whoever has most recently
+/// published an info. The strategy for accomplishing this is defined here:
+///
+/// Consider "the store" acting as a stack. It can be implemented in any manner,
```
jost-s, line 116: That sounds more like a queue. To a stack elements are added and removed only from the top.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - The server SHOULD provide a configurable per-space max info count.
+///   For the duration of this document, that value will be called MAX_INFOS.
+///   It is recommended to default that value to `32`.
+/// - The server SHOULD delete expired infos periodically.
+/// - If the server is already storing an agent, and a `PUT` with a newer
+///   `createdAt` arrives, the existing entry MUST be replaced.
+/// - If the store count is < MAX_INFOS, the server MUST push the info
+///   onto the stack.
+/// - If the store count is >= MAX_INFOS, the server MUST delete the entry
+///   at MAX_INFOS / 2 and then push the info onto the stack.
```
jost-s, line 130: MAX_INFOS / 2 is the first of the recently put agent infos?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Consider "the store" acting as a stack. It can be implemented in any manner,
+/// but the following strategy assumes new entries are added to the end,
+/// and when any entries are removed, the indexes of any items after them
+/// will be decrimented.
+///
+/// - The server SHOULD provide a configurable per-space max info count.
+///   For the duration of this document, that value will be called MAX_INFOS.
+///   It is recommended to default that value to `32`.
+/// - The server SHOULD delete expired infos periodically.
+/// - If the server is already storing an agent, and a `PUT` with a newer
```
jost-s, line 125: If `createdAt` is not newer, the incoming agent is ignored? And I suppose that if `isTumbstone` is `true`, `expiresAt` and `createdAt` are ignored as well?

jost-s (commented):
Some clarifications

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// ##### 3.1. The Future
+///
+/// It is the intention to someday in the future to add a "trusted" strategy,
+/// that will be triggerd via a new api, perhaps `/registerTrust/Base64Space`.
+///
+/// Even when that API is implemented, however, the default strategy defined
+/// next will be used on all spaces that have not been registered with a
+/// different strategy. Therefore, we can proceed for now without consideration
+/// for any other future strategies.
```
artbrock, line 102: Should the list of returned agent infos be segmented into two lists by trusted/authoritative and self-published?

And for now the trusted list could be empty, and the self-published list works as being described in this doc?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - A server SHOULD decide a max count of spaces it intends to support
+///   (based on available memory or disk space for storing agents within those
+///   spaces), and then return error 429 beyond that.
+/// - A server MAY make the above max space count configurable.
+/// - A server SHOULD track the IP addresses of clients that make PUT requests
+///   which result in space creation, and error with 429 if that frequency
+///   is beyond a limit.
+/// - A server MAY make this limit configurable.
+#[cfg(doc)]
+pub mod spec {}
```
artbrock, line 150: Can we also have a /metrics path which behaves similar to the current metrics system (showing number of space and active nodes in each space (or maybe number of trusted and self-published nodes in each space)) for a block of time, except this time have the results be REVERSE chronological showing the most recent time blocks first?

crates/boot_srv/src/lib.rs
mattyg: I personally don't want any placeholders in the codebase, as IMO they just add noise and friction for development. When we're ready to implement this functionality, we can bump the spec version, update the spec, bump the api version, implement the updated spec, and add tests for the updated implementation.

crates/boot_srv/src/lib.rs
mattyg: Actually I want to make a stronger case for this.

We know the spec will change, and the functionality will change. Our intention is that all changes will be forwards compatible. By using api versioning we:
- make the forwards compatibility explicit in the codebase, rather than relying on folk knowledge or reviewing old versions of the codebase to confirm this.
- make dropping support for old versions explicit in the codebase
- make explicit to the client which version of the spec is supported by the server it is calling
- allow ourselves the freedom to make breaking changes gracefully. Because we can't be certain we will never want any breaking changes

If we *want* to maintain multiple api versions we can, but having versioning doesn't require that. It just means we're explicit in the codebase and to clients that an api version has been deprecated.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///   - Response Body: `ListResponse | ErrResponse`
+/// - `GET /`
+///   - Response Body: `OkResponse | ErrResponse`
+///
+/// ##### 2.2. Publishing info to the boot server.
+///
+/// A `PUT` on `/boot/Base64Space/Base64Agent` with an `AgentInfoSigned`
+/// json object as the request body.
+///
+/// - The server MUST reject the request if the body is > 1024 bytes.
```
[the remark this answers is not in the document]
neonphog: Encoded agent infos are around 300 bytes, 1K is a nice round binary number with some wiggle room for long urls and lots of extended metadata in the future, while still being reasonable in terms of not letting people store video files in our bootstrap server.

crates/boot_srv/src/lib.rs
neonphog: No, it's the `MAX_INFOS / 2`th most recently pushed info.

crates/boot_srv/src/lib.rs
neonphog: No, expiresAt and createdAt are both still required to be valid for a tombstoning to happen. That's more relevant for the p2p gossip of agent infos, but we might as well respect the rules for the bootstrap server as well.

crates/boot_srv/src/lib.rs
neonphog: My intention is that spaces will *either* use the "default" storage strategy defined here, *or* the "trusted" strategy to be defined in the future. Once we have the ability, if you register a space as trusted, it will then only accept trusted infos and only return those trusted infos.

crates/boot_srv/src/lib.rs
neonphog: Usually metrics are a server-specific implementation detail. I'm happy to include a metrics api with the server.

@artbrock - do you think a metrics api needs to be included in the spec? Indicating that any k2 bootstrap server implementation anybody made would respond with the exact same metrics?

crates/boot_srv/src/lib.rs
neonphog: [Just FYI, it isn't just me that disagrees with versioning apis](https://www.hmeid.com/blog/just-say-no-to-versioning)

Aside from that I don't think there's any need to *start* with the versioning. If we ever do decide there's no way forward except to version the api. We could introduce a `/v2/` at the front of the path OR an `x-api-ver=2` http header, then say if you don't include those, you get the v1 api while it's deprecated, and then say those are required when we delete the deprecation.

I feel pretty strongly about this as well, but it's not a hill I need to die on. If others also prefer versioned APIs, I'm fine doing it.

crates/boot_srv/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "kitsune2_boot_srv"
```
ThetaSinner, line 2: `boot` sounds like it's for startup, I'd prefer to spell this one out :)

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+///
+/// ##### 2.1. In Brief
+///
+/// ```text
+/// ErrResponse = { "error": string }
+/// OkResponse = {}
+/// ListResponse = [ AgentInfoSigned, .. ]
+/// ```
+///
+/// - `PUT /boot/Base64Space/Base64Agent`
```
ThetaSinner, line 50: Can we make it clearer that these are path parameters? `{Base64Space}` or `<Base64Space>`?

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - `Base64Sig` - base64UrlNoPad safe encoded string crypto signature.
+/// - `Json` - string containing json that can be decoded.
+/// - `I64` - string containing an i64 number.
+///
+/// ```text
+/// AgentInfoSigned = { "agentInfo": Json, "signature": Base64Sig }
+/// AgentInfo = {
+///   "agent": Base64Agent,
+///   "space": Base64Space,
+///   "createdAt": I64,
```
ThetaSinner, line 32: Can we be more specific with this type? local time vs UTC time is so easy to get tripped up on

crates/boot_srv/src/lib.rs
ThetaSinner: This doesn't seem right to me. A tombstone to me is a marker that something has been deleted. So I'd expect people querying the bootstrap server to see this agent as having removed themselves. If we just want ourselves gone, that should be  DELETE verb?

crates/boot_srv/src/lib.rs
mattyg: Yeah that makes sense, we can wait and add if needed. Not the hill I want to die on either, and I'm feeling like I got a bit bikesheddy. I see there are valid arguments both ways.

crates/boot_srv/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+/// - If `isTombstone` is `true`, the server MUST delete any existing
+///   info being held.
+/// - If `isTombstone` is `false`, the server MAY begin storing the info.
+///   See section 3. on storage strategies below.
+///
+/// ##### 2.3. Listing data stored on the boot server.
+///
+/// A `GET` on `/boot/Base64Space`.
+///
+/// - The server MUST respond with a complete list of stored infos.
```
ThetaSinner, line 78: I like this, it will simplify things like hcterm and general debugging where you have to keep calling "give me some more please" and hope to build a full set for display :)

Was there some kind of security reason not to do this in the first place? Or was it just that Kitsune wasn't supposed to call it like this?

crates/boot_srv/src/lib.rs
jost-s: Agreed, it's a chimera data type. =)

crates/boot_srv/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/dd48b48f27a04f078fe82d3a77af60d6df60a43c

crates/boot_srv/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/dd48b48f27a04f078fe82d3a77af60d6df60a43c

crates/boot_srv/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/5e4fc8768a7dd43757a8b4c6d2ffc895208a50a0

crates/boot_srv/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/a2338e688dcb6793f7a37065d85d69fa9f454d6a

crates/boot_srv/src/lib.rs
neonphog: Consider the verifiability aspect. We cannot let just anybody delete any record, so we'd have to come up with some alternate payload for the clients to sign anyways... Can we just use the existing functionality?

crates/boot_srv/src/lib.rs
neonphog: See https://github.com/holochain/kitsune2/pull/6#discussion_r1849077855

neonphog: I know it's not working as a tombstone on the bootstrap server... but it *is* working as a tombstone within the kitsune peer gossip, so let's just reuse that already written functionality.

crates/boot_srv/src/lib.rs
neonphog: I only added the reject if more than 30 minutes: https://github.com/holochain/kitsune2/pull/6/commits/84799bc759c5b2200b34e0d7036a6eaa062c1a98

The clients can already spam the server by sending a new createdAt every millisecond or whatever, so that side will have to be covered by the rate limiting.

crates/boot_srv/src/lib.rs
neonphog: To avoid the complexity of the server needing to tell the clients what the "configured" expiration time is so they know how often to refresh / what they can set it to... Let's start with just a hard-coded maximum of 30 minutes, and see if there is need for anything else.

neonphog: Let me know if this is any better: https://github.com/holochain/kitsune2/pull/6/commits/984dc99ec51274bfa7bd60245f44daf089c972a4

crates/boot_srv/src/lib.rs
neonphog: That's described in the first paragraphs of section 3.2, basically the first half of the stack is allocated to known reliable nodes (because they've been online without expiring for the longest). And the second half of the stack is allocated to an ever changing list of the most recent nodes to publish their agent infos (to make it less possible for the older peers to collude on an eclipse attack).

The "delete the middle one" algorithm was my attempt at the simplest design of this.

crates/boot_srv/Cargo.toml
neonphog: Shall we just go with `bs`? /jk

crates/boot_srv/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/d49c23eb24e6960bb5847ed030e062a9d1d0f579

crates/boot_srv/src/lib.rs
neonphog: Good call, done: https://github.com/holochain/kitsune2/pull/6/commits/4b12a8b8d4b6c042bb3f0e2d072bd5039396c379

crates/boot_srv/src/lib.rs
neonphog: The legacy bootstrap server holds all agents within the space. We didn't ever want to send down the potentially millions of entries we thought holochain might scale to, so we had the server randomize the list and send you only a handful.

crates/boot_srv/Cargo.toml
neonphog: Done: https://github.com/holochain/kitsune2/pull/6/commits/edc254629867487c5bb99ac47f6ebf531dc1d96d

crates/boot_srv/src/lib.rs
ThetaSinner: Ahh, I get why it's been done this way now. We'd have to sign something to send to the delete anyway... Yeah okay, fair enough :)

crates/boot_srv/src/lib.rs
ThetaSinner: That makes sense, so we probably shouldn't actually expose that in `hcterm`. But still, for debugging if we know a space is relatively new and small then it would be useful :)

ThetaSinner (approved):

ThetaSinner (approved):

PR #8, opened 2024-11-21
K2Error Type

neonphog:
Please review this as the error type to be implemented in the kitsune2 external api. (and potentially for internal module traits if we so choose).

Once this PR is merged, I will go through the existing code implementing its use.

### DIFF
--- a/crates/api/src/error.rs
+++ b/crates/api/src/error.rs
@@ -50,7 +50,7 @@ pub enum K2Error {
     #[error("{ctx} (src: {src})")]
     Other {
         /// Any context associated with this error.
-        ctx: String,
+        ctx: Arc<str>,
 
         /// The inner error (if any).
         #[source]
@@ -68,7 +68,7 @@ impl K2Error {
         src: S,
     ) -> Self {
         Self::Other {
-            ctx: ctx.to_string(),
+            ctx: ctx.to_string().into_boxed_str().into(),
             src: DynInnerError::new(src),
         }
     }
@@ -76,7 +76,7 @@ impl K2Error {
     /// Construct an "other" error.
     pub fn other<C: std::fmt::Display>(ctx: C) -> Self {
         Self::Other {
-            ctx: ctx.to_string(),
+            ctx: ctx.to_string().into_boxed_str().into(),
             src: DynInnerError::default(),
         }
     }
@@ -117,5 +117,11 @@ mod test {
             "Other { ctx: \"foo\", src: Some(Custom { kind: Other, error: \"bar\" }) }",
             format!("{:?}", K2Error::other_src("foo", std::io::Error::other("bar"))).as_str(),
         );
+    }
+
+    #[test]
+    fn ensure_k2error_type_is_send_and_sync() {
+        fn ensure<T: std::fmt::Display + Send + Sync>(_t: T) {}
+        ensure(K2Error::other("bla"));
     }
 }
--- a/crates/api/src/error.rs
+++ b/crates/api/src/error.rs
@@ -1,0 +1,121 @@
+//! Kitsune2 error types.
+
+use std::sync::Arc;
+
+/// A clonable trait-object inner error.
+#[derive(Clone, Default)]
+pub struct DynInnerError(
+    pub Option<Arc<dyn std::error::Error + 'static + Send + Sync>>,
+);
+
+impl std::fmt::Debug for DynInnerError {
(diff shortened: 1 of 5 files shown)


crates/api/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// This type is required to implement `Clone` to ease the use of
+/// shared futures, which require the entire `Result` to be `Clone`.
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum K2Error {
+    /// Generic Kitsune2 internal error.
+    #[error("{ctx} (src: {src})")]
+    Other {
+        /// Any context associated with this error.
+        ctx: Arc<str>,
```
cdunster, line 53: Why not a `String`?

crates/api/src/error.rs
```diff
@@ -0,0 +1,127 @@
...
+
+        /// The inner error (if any).
+        #[source]
+        src: DynInnerError,
```
cdunster, line 57: Will the inner type ever be anything other than a `std::io::Error`? If not then can we just make this `Option<std::io::Error>` or if there's a small chance that we might add another error type later then maybe we could call this variant `OtherIo` and still make the change?

crates/api/src/error.rs
ThetaSinner: Agree, we'll want to be able to format values in to error messages sometimes

crates/api/src/error.rs
```diff
@@ -0,0 +1,127 @@
+//! Kitsune2 error types.
+
+use std::sync::Arc;
+
+/// A clonable trait-object inner error.
+#[derive(Clone, Default)]
+pub struct DynInnerError(
```
ThetaSinner, line 7: I'd like the inner errors to be reasonably expressive. Is that what this is allowing? 

Like I could define a `thiserror::Error` in the gossip module and use that internally to define expressive errors for testing but at the API boundary I should use `DynInnerError` so that other modules can use the API without ending up nesting the module errors?

crates/api/src/error.rs
ThetaSinner: I think there's a good chance we want other error types. I'd put some error handling into my gossip code just so I could keep moving, not necessarily planning to keep it. I was using a similar trait definition to this one and it's been useful so far :)

crates/api/src/error.rs
neonphog: The ref pointer makes cloning cheaper, and Arc or Box str is a type indication that the string won't be change-able.

Take a look at the `other()` and `other_src()` helper functions which I imagine should be used most of the time for constructing. If you need to do it manually, you totally can.

All that being said, I'm fine switching back to Strings if we don't care about the clone overhead.

crates/api/src/error.rs
neonphog: Yeah, this other type is the general fallback case. Everything from io errors to every kind of dependency library error, to our own internal custom errors can go in here until we identify the need to break one/any of those out into their own top-level variants.

crates/api/src/error.rs
neonphog: Yes, this is exactly the use-case I was hoping for with this.

ThetaSinner: Super, I like it :)

crates/api/src/error.rs
ThetaSinner: Got it, we can do what we need to do then, sounds good!

ThetaSinner (approved):

crates/api/src/error.rs
neonphog: @ThetaSinner - Since this comment is specifically on the DynInnerError type, I feel the need to point out: My comment above is actually about the "Other" variant. This DynInnerError type is specifically because the thiserror Derive logic is not expressive enough, I had to have a newtype here or require a custom Display implementation on the K2Error type which would have meant later variants couldn't make use of the handy macros for configuring the variant display impl.

crates/api/src/error.rs
mattyg: Huh interesting point. Yeah that does make sense

crates/api/src/error.rs
ThetaSinner: Hmm okay, I think I follow but I also think I need to get hands on with this and see how it applies. I'm happy with this as a way to keep errors simple initially and if we end up wanting to make adjustments to the inner details then that is fine :)

ThetaSinner (approved):

PR #9, opened 2024-11-21
Kitsune2 Bootstrap Server--Preparation

neonphog:
The complete boostrap server branch is rather large. Breaking out the dependencies needed and some unrelated CI tooling changes so the actual server pull requests can be reviewed with more focus.

- See the follow on [Core #10](https://github.com/holochain/kitsune2/pull/10) for the core implementation code of the server.
- See the follow on [Testing #11](https://github.com/holochain/kitsune2/pull/11) PR for api endpoint tests and general notes about this effort in the PR description.

PR #10, opened 2024-11-21
Kitsune2 Bootstrap Server--Core

neonphog:
This is the core functionality of the kitsune2 bootstrap server code.

To make the PRs more manageable, I've broken them out into some preparatory work before this, and all the endpoint test after this.

- See the previous [Prep #9](https://github.com/holochain/kitsune2/pull/9) PR for that prep work.
- See the follow on [Testing #11](https://github.com/holochain/kitsune2/pull/11) PR for api endpoint tests and general notes about this effort in the PR description.

### DIFF
--- a/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
+++ b/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
@@ -1,0 +1,49 @@
+//! The binary kitsune2-bootstrap-srv.
+
+use kitsune2_bootstrap_srv::*;
+
+#[derive(clap::Parser, Debug)]
+#[command(version)]
+pub struct Args {
+    /// By default kitsune2-boot-srv runs in "testing" configuration
+    /// with much lighter resource usage settings. This testing mode
+    /// should be more than enough for most developer application testing
+    /// and continuous integration or automated tests.
+    ///
+    /// To setup the server to be ready to use most of the resources available
+    /// on a single given machine, you can set this "production" mode.
+    #[arg(long)]
+    pub production: bool,
+    // TODO - Implement the ability to specify TLS certificates
+    // TODO - Implement the ability to specify the listening address
+    // TODO - Implement the ability to override any other relevant
+    //        config params that we wish to expose
+}
+
+fn main() {
+    let args = <Args as clap::Parser>::parse();
+
+    let config = if args.production {
+        Config::production()
+    } else {
+        Config::testing()
+    };
+
+    println!("{args:?}--{config:?}");
+
+    let (send, recv) = std::sync::mpsc::channel();
+
+    ctrlc::set_handler(move || {
+        send.send(()).unwrap();
+    })
+    .unwrap();
+
+    let srv = BootstrapSrv::new(config);
+
+    let _ = recv.recv();
+
+    println!("Terminating...");
+    drop(srv);
+    println!("Done.");
+    std::process::exit(0);
+}
--- a/crates/bootstrap_srv/src/config.rs
+++ b/crates/bootstrap_srv/src/config.rs
@@ -1,0 +1,84 @@
+//! config types.
+
+/// Configuration for running a BootstrapSrv.
+#[derive(Debug)]
+pub struct Config {
+    /// Worker thread count.
+    ///
+    /// This server is currently built using blocking io and filesystem
+    /// storage. It is therefore beneficial to have more worker threads
+    /// than system cpus, since the workers will be bound on io, not
+    /// on cpu. On the other hand, increasing this will also increase
+    /// memory overhead and tempfile handle count, so we don't want to
+    /// set it too high.
+    ///
+    /// Defaults:
+    /// - `testing = 2`
+    /// - `production = 4 * cpu_count`
+    pub worker_thread_count: usize,
+
+    /// The maximum agent info entry count per space.
+    ///
+    /// All entries will be returned in a get space request, so
+    /// this count should be low enough to reasonably send this response
+    /// over http without needing pagination.
+    ///
+    /// Defaults:
+    /// - `testing = 32`
+    /// - `production = 32`
+    pub max_entries_per_space: usize,
+
+    /// The duration worker threads will block waiting for incoming connections
+    /// before checking to see if the server is shutting down.
+    ///
+    /// Setting this very high will cause ctrl-c / server shutdown to be slow.
+    /// Setting this very low will increase cpu overhead (and in extreme
+    /// conditions, could cause a lack of responsiveness in the server).
+    ///
+    /// Defaults:
+    /// - `testing = 10ms`
+    /// - `production = 2s`
+    pub request_listen_duration: std::time::Duration,
+
+    /// The address(es) at which to listen.
+    ///
+    /// Defaults:
+    /// - `testing = "127.0.0.1:0"`
+    /// - `production = "0.0.0.0:443"`
+    pub listen_address: std::net::SocketAddr,
+
+    /// The interval at which expired agents are purged from the cache.
+    /// This is a fairly expensive operation that requires iterating
+    /// through every registered space and loading all the infos off the disk,
+    /// so it should not be undertaken too frequently.
+    ///
+    /// Defaults:
+    ///
+    /// - `testing = 10s`
+    /// - `production = 60s`
+    pub prune_interval: std::time::Duration,
+}
+
+impl Config {
+    /// Get a boot_srv config suitable for testing.
+    pub fn testing() -> Self {
+        Self {
+            worker_thread_count: 2,
+            max_entries_per_space: 32,
+            request_listen_duration: std::time::Duration::from_millis(10),
+            listen_address: ([127, 0, 0, 1], 0).into(),
+            prune_interval: std::time::Duration::from_secs(10),
+        }
+    }
+
+    /// Get a boot_srv config suitable for production.
+    pub fn production() -> Self {
+        Self {
+            worker_thread_count: num_cpus::get() * 4,
+            max_entries_per_space: 32,
+            request_listen_duration: std::time::Duration::from_secs(2),
+            listen_address: ([0, 0, 0, 0], 443).into(),
+            prune_interval: std::time::Duration::from_secs(60),
+        }
+    }
+}
--- a/crates/bootstrap_srv/src/parse.rs
+++ b/crates/bootstrap_srv/src/parse.rs
@@ -1,4 +1,4 @@
-/// An entry with known content.
+/// An entry with known content: [crate::spec#1-types].
 #[derive(Debug)]
 pub struct ParsedEntry {
     /// agent
--- a/crates/bootstrap_srv/src/parse.rs
+++ b/crates/bootstrap_srv/src/parse.rs
@@ -25,7 +25,7 @@ pub struct ParsedEntry {
 
 impl ParsedEntry {
     /// Parse entry from a slice.
-    pub fn from_slice(slice: &[u8]) -> std::io::Result<Self> {
+    pub fn try_from_slice(slice: &[u8]) -> std::io::Result<Self> {
         use base64::prelude::*;
 
         #[derive(serde::Deserialize)]
--- a/crates/bootstrap_srv/src/parse.rs
+++ b/crates/bootstrap_srv/src/parse.rs
@@ -1,0 +1,83 @@
+/// An entry with known content.
+#[derive(Debug)]
+pub struct ParsedEntry {
+    /// agent
+    pub agent: ed25519_dalek::VerifyingKey,
+
+    /// space
+    pub space: bytes::Bytes,
+
+    /// created_at
+    pub created_at: i64,
+
+    /// expires_at
+    pub expires_at: i64,
+
+    /// is_tombstone
+    pub is_tombstone: bool,
+
+    /// encoded
+    pub encoded: String,
+
+    /// signature
+    pub signature: ed25519_dalek::Signature,
+}
+
+impl ParsedEntry {
+    /// Parse entry from a slice.
+    pub fn from_slice(slice: &[u8]) -> std::io::Result<Self> {
+        use base64::prelude::*;
+
+        #[derive(serde::Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct Outer {
+            agent_info: String,
+            signature: String,
+        }
+
+        #[derive(serde::Deserialize)]
+        #[serde(rename_all = "camelCase")]
+        struct Inner {
+            agent: String,
+            space: String,
+            created_at: String,
+            expires_at: String,
+            is_tombstone: bool,
+        }
+
+        let out: Outer = serde_json::from_slice(slice)?;
+        let inn: Inner = serde_json::from_str(&out.agent_info)?;
+
+        let agent: [u8; 32] = BASE64_URL_SAFE_NO_PAD
+            .decode(inn.agent)
+            .map_err(std::io::Error::other)?
+            .try_into()
+            .map_err(|_| std::io::Error::other("InvalidAgentPubKey"))?;
+
+        let signature: [u8; 64] = BASE64_URL_SAFE_NO_PAD
+            .decode(out.signature)
+            .map_err(std::io::Error::other)?
+            .try_into()
+            .map_err(|_| std::io::Error::other("InvalidSignature"))?;
+
+        Ok(ParsedEntry {
+            agent: ed25519_dalek::VerifyingKey::from_bytes(&agent)
+                .map_err(std::io::Error::other)?,
+            space: BASE64_URL_SAFE_NO_PAD
+                .decode(inn.space)
+                .map_err(std::io::Error::other)?
+                .into(),
+            created_at: inn
+                .created_at
+                .parse()
+                .map_err(std::io::Error::other)?,
+            expires_at: inn
+                .expires_at
+                .parse()
+                .map_err(std::io::Error::other)?,
+            is_tombstone: inn.is_tombstone,
+            encoded: out.agent_info,
+            signature: ed25519_dalek::Signature::from_bytes(&signature),
+        })
+    }
+}
--- a/crates/bootstrap_srv/src/server.rs
+++ b/crates/bootstrap_srv/src/server.rs
@@ -91,7 +91,7 @@ impl BootstrapSrv {
         })
     }
 
-    /// Get the bound listinging address of this server.
+    /// Get the bound listening address of this server.
     pub fn listen_addr(&self) -> std::net::SocketAddr {
         self.addr
     }
--- a/crates/bootstrap_srv/src/server.rs
+++ b/crates/bootstrap_srv/src/server.rs
@@ -5,6 +5,16 @@ use std::sync::Arc;
 use crate::*;
 use tiny_http::*;
 
+/// Don't allow created_at to be greater or less than this far away from now.
+/// 3 minutes.
+const CREATED_AT_CLOCK_SKEW_ALLOWED_MICROS: i64 =
+    std::time::Duration::from_secs(60 * 3).as_micros() as i64;
+
+/// Don't allow expires_at - created_at to be greater than this duration.
+/// 30 minutes.
+const EXPIRES_AT_DURATION_MAX_ALLOWED_MICROS: i64 =
+    std::time::Duration::from_secs(60 * 30).as_micros() as i64;
+
 /// An actual kitsune2_bootstrap_srv server instance.
 ///
 /// This server is built to be direct, light-weight, and responsive.
@@ -68,10 +78,10 @@ impl BootstrapSrv {
         }
 
         // also set up a worker for pruning expired infos
-        let maint_cont = cont.clone();
-        let maint_space_map = space_map.clone();
+        let prune_cont = cont.clone();
+        let prune_space_map = space_map.clone();
         workers.push(std::thread::spawn(move || {
-            maint_worker(config, maint_cont, maint_space_map)
+            prune_worker(config, prune_cont, prune_space_map)
         }));
 
         Ok(Self {
@@ -87,7 +97,7 @@ impl BootstrapSrv {
     }
 }
 
-fn maint_worker(
+fn prune_worker(
     config: Arc<Config>,
     cont: Arc<std::sync::atomic::AtomicBool>,
     space_map: crate::SpaceMap,
@@ -222,16 +232,12 @@ impl<'lt> Handler<'lt> {
         }
 
         // validate created at is not older than 3 min ago
-        if info.created_at
-            < now - (std::time::Duration::from_secs(60 * 3).as_micros() as i64)
-        {
+        if info.created_at < now - CREATED_AT_CLOCK_SKEW_ALLOWED_MICROS {
             return Err(std::io::Error::other("InvalidCreatedAt"));
         }
 
         // validate created at is less than 3 min in the future
-        if info.created_at
-            > now + (std::time::Duration::from_secs(60 * 3).as_micros() as i64)
-        {
+        if info.created_at > now + CREATED_AT_CLOCK_SKEW_ALLOWED_MICROS {
             return Err(std::io::Error::other("InvalidCreatedAt"));
         }
 
@@ -247,7 +253,7 @@ impl<'lt> Handler<'lt> {
 
         // validate expires_at is not more than 30 min after created_at
         if info.expires_at - info.created_at
-            > (std::time::Duration::from_secs(60 * 30).as_micros() as i64)
+            > EXPIRES_AT_DURATION_MAX_ALLOWED_MICROS
         {
             return Err(std::io::Error::other("InvalidExpiresAt"));
         }
--- a/crates/bootstrap_srv/src/server.rs
+++ b/crates/bootstrap_srv/src/server.rs
@@ -15,6 +15,15 @@ const CREATED_AT_CLOCK_SKEW_ALLOWED_MICROS: i64 =
 const EXPIRES_AT_DURATION_MAX_ALLOWED_MICROS: i64 =
     std::time::Duration::from_secs(60 * 30).as_micros() as i64;
 
+/// Print out a message if this thread dies.
+struct ThreadGuard(&'static str);
+
+impl Drop for ThreadGuard {
+    fn drop(&mut self) {
+        eprintln!("{}", self.0);
+    }
+}
+
 /// An actual kitsune2_bootstrap_srv server instance.
 ///
 /// This server is built to be direct, light-weight, and responsive.
@@ -116,6 +125,8 @@ fn prune_worker(
     cont: Arc<std::sync::atomic::AtomicBool>,
     space_map: crate::SpaceMap,
 ) -> std::io::Result<()> {
+    let _g = ThreadGuard("WARN: prune_worker thread has ended");
+
     let mut last_check = std::time::Instant::now();
 
     while cont.load(std::sync::atomic::Ordering::SeqCst) {
@@ -138,6 +149,8 @@ fn worker(
     server: Arc<Server>,
     space_map: crate::SpaceMap,
 ) -> std::io::Result<()> {
+    let _g = ThreadGuard("WARN: worker thread has ended");
+
     while cont.load(std::sync::atomic::Ordering::SeqCst) {
         let req = match server.recv_timeout(config.request_listen_duration)? {
             Some(req) => req,
--- a/crates/bootstrap_srv/src/server.rs
+++ b/crates/bootstrap_srv/src/server.rs
@@ -1,0 +1,333 @@
+//! bootstrap http server types.
+
+use std::sync::Arc;
+
+use crate::*;
+use tiny_http::*;
+
+/// An actual kitsune2_bootstrap_srv server instance.
+///
+/// This server is built to be direct, light-weight, and responsive.
+/// On the server-side, as one aspect toward accomplishing this,
+/// we are eschewing async code in favor of os thread workers.
+pub struct BootstrapSrv {
+    cont: Arc<std::sync::atomic::AtomicBool>,
+    workers: Vec<std::thread::JoinHandle<std::io::Result<()>>>,
+    addr: std::net::SocketAddr,
+}
+
+impl Drop for BootstrapSrv {
+    fn drop(&mut self) {
+        self.cont.store(false, std::sync::atomic::Ordering::SeqCst);
+        for worker in self.workers.drain(..) {
+            let _ = worker.join().expect("Failure shutting down worker thread");
+        }
+    }
+}
+
+impl BootstrapSrv {
+    /// Construct a new BootstrapSrv instance.
+    pub fn new(config: Config) -> std::io::Result<Self> {
+        let config = Arc::new(config);
+
+        // atomic flag for telling worker threads to shutdown
+        let cont = Arc::new(std::sync::atomic::AtomicBool::new(true));
+
+        // synchronization type for managing infos in spaces
+        let space_map = crate::SpaceMap::default();
+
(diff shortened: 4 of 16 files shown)


crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// should be more than enough for most developer application testing
+    /// and continuous integration or automated tests.
+    ///
+    /// To setup the server to be ready to use most of the resources available
+    /// on a single given machine, you can set this "production" mode.
+    #[arg(long)]
+    pub production: bool,
+    // TODO - Implement the ability to specify TLS certificates
+    // TODO - Implement the ability to specify the listening address
+    // TODO - Implement the ability to override any other relevant
```
mattyg, line 19: Do we want to move these into issues in the kitsune2 repo?

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    /// Read the body while respecting our max message size.
+    fn read_body(&mut self) -> std::io::Result<Vec<u8>> {
+        // these are the same right now, but *could* be different
+        const MAX_INFO_SIZE: usize = 1024;
+        const READ_BUF_SIZE: usize = 1024;
+
+        let mut buf = [0; READ_BUF_SIZE];
+        let mut out = Vec::new();
+        loop {
```
mattyg, line 301: So the idea is that if reading gets interrupted for some reason, it retries? What conditions would cause it to get interrupted? Should we limit the number of retries to avoid something like a disk failure causing infinite looping?

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+        // validate created at is not older than 3 min ago
+        if info.created_at
+            < now - (std::time::Duration::from_secs(60 * 3).as_micros() as i64)
+        {
+            return Err(std::io::Error::other("InvalidCreatedAt"));
+        }
+
+        // validate created at is less than 3 min in the future
+        if info.created_at
+            > now + (std::time::Duration::from_secs(60 * 3).as_micros() as i64)
```
mattyg, line 233: Maybe we should move this value to a const in the config file.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        // validate space matches url path
+        if space != info.space {
+            return Err(std::io::Error::other("InvalidSpace"));
+        }
+
+        // validate created at is not older than 3 min ago
+        if info.created_at
+            < now - (std::time::Duration::from_secs(60 * 3).as_micros() as i64)
```
mattyg, line 226: Maybe we should move this value to a const in the config file.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+        })
+    }
+
+    /// Get the bound listinging address of this server.
+    pub fn listen_addr(&self) -> std::net::SocketAddr {
+        self.addr
+    }
+}
+
+fn maint_worker(
```
mattyg, line 90: I read this and thought it was a typo on "main_worker" -- maybe spelling it fully would be cleaer.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+
+impl<'lt> Handler<'lt> {
+    /// Wrap the handle call so we can respond to the client with errors.
+    pub fn handle(mut self) -> std::io::Result<()> {
+        match self.handle_inner() {
+            Ok((status, body)) => self.respond(status, body),
+            Err(err) => self.respond(
+                500,
+                serde_json::to_string(&serde_json::json!({
+                    "error": format!("{err:?}"),
```
mattyg, line 167: It looks like there's only one error where the error string is more than just a "type" label (line 259 `                std::io::Error::other(format!("InvalidSignature: {err:?}"))`.

If we want to include any message along with a "type" label, then I'd want them split them out into 2 fields for clients to easily match on.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+    req: tiny_http::Request,
+}
+
+impl<'lt> Handler<'lt> {
+    /// Wrap the handle call so we can respond to the client with errors.
+    pub fn handle(mut self) -> std::io::Result<()> {
+        match self.handle_inner() {
+            Ok((status, body)) => self.respond(status, body),
+            Err(err) => self.respond(
+                500,
```
mattyg, line 165: Nitpick but some of these errors would better fit the 400 response type.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+            > (std::time::Duration::from_secs(60 * 30).as_micros() as i64)
+        {
+            return Err(std::io::Error::other("InvalidExpiresAt"));
+        }
+
+        // validate signature (do this at the end because it's more expensive
+        info.agent
+            .verify(info.encoded.as_bytes(), &info.signature)
+            .map_err(|err| {
+                std::io::Error::other(format!("InvalidSignature: {err:?}"))
```
mattyg, line 259: The error mentioned above.

crates/bootstrap_srv/src/space.rs
```diff
@@ -0,0 +1,228 @@
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+use crate::store::*;
+
+/// A map of spaces.
+#[derive(Clone)]
+pub struct SpaceMap(Arc<Mutex<HashMap<bytes::Bytes, Space>>>);
```
mattyg, line 8: Is there a reason to use `bytes::Bytes` instead of the wrapper type you made `SpaceId`?

crates/bootstrap_srv/src/space.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl Default for SpaceMap {
+    fn default() -> Self {
+        Self(Arc::new(Mutex::new(HashMap::new())))
+    }
+}
+
+impl SpaceMap {
+    /// Read the content of a space.
+    pub fn read(&self, space: &bytes::Bytes) -> std::io::Result<Vec<u8>> {
```
mattyg, line 18: same here - should this be a `SpaceId`?

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
neonphog: This PR is against the kitsune2 repo. Sorry, not sure what you mean here.

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+                Err(e) if e.kind() == std::io::ErrorKind::Interrupted => {
+                    continue;
+                }
+                Err(e) => return Err(e),
+            };
+            if read == 0 {
+                return Ok(out);
+            }
+            out.extend_from_slice(&buf[..read]);
+            if out.len() > MAX_INFO_SIZE {
```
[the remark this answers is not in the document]
neonphog: Clients only optionally send this header, and then (while technically required by the spec) may not always send the same byte count... so I thought best to guard on the actual data sent.

crates/bootstrap_srv/src/server.rs
neonphog: This is the fundamental behavior of the core libc read function, carried forward from back in the days when interrupt requests were how processes were parallelized. When you call `write_all`, under the hood it is doing this exact loop check. There is no need to diverge from cannon here.

crates/bootstrap_srv/src/space.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+/// A concurrent single space.
+#[derive(Clone)]
+struct Space {
+    // Using a separate write lock instead of a RwLock, because
+    // we still want to allow reads while the write_lock is locked.
+    // The write function will lock the readable when it is done processing
+    // and quickly update the reader, then drop both locks (read first).
```
mattyg, line 89: I don't understand the difference between this behavior and RwLock, but as long as others do that's fine with me.

crates/bootstrap_srv/src/config.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// storage. It is therefore beneficial to have more worker threads
+    /// than system cpus, since the workers will be bound on io, not
+    /// on cpu. On the other hand, increasing this will also increase
+    /// memory overhead and tempfile handle count, so we don't want to
+    /// set it too high.
+    ///
+    /// Defaults:
+    /// - `testing = 2`
+    /// - `production = 4 * cpu_count`
+    pub worker_thread_count: usize,
```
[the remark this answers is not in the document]
neonphog: mmm... I'm doing that as a DRY tool, since the same cleanup logic was required. Many other threads will be created by the tiny_http library itself, and who knows about any others for tracing libraries, etc. I'm not sure we need to enumerate them all.

crates/bootstrap_srv/src/space.rs
neonphog: Yes. I see this server / protocol being useful for WAN discovery outside of projects using kitsune itself. Perhaps it would actually be better to exist outside this monorepo, but that is more overhead for us. At the very least, I didn't want this to have any dependencies on other kitsune crates.

crates/bootstrap_srv/src/server.rs
neonphog: @mattyg - For this and https://github.com/holochain/kitsune2/pull/10#discussion_r1854225991 (which I don't consider a nitpick, I agree these are errors in user input, not in what the server is doing), I wonder if the error reporting types, since they are defined in json, and statuses actually belong in the spec itself. Would you be okay with updating the spec and fixing these responses being a follow-on effort after these are merged?

crates/bootstrap_srv/src/server.rs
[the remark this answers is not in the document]
neonphog: See https://github.com/holochain/kitsune2/pull/10#discussion_r1854281943

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,191 @@
+//! This is a virtual-memory inspired tempfile store.
+//!
+//! ### Rationale
+//!
+//! - We don't need to persist anything in the store beyond a single
+//!   process invocation, because the infos are going to expire after
+//!   a matter of minutes anyways, and peers will continue to re-publish them.
+//! - We would like a server to be able to store more spaces and infos
+//!   than would reasonably fit in RAM.
```
mattyg, line 9:
```suggestion
//!   than would reasonably fit in RAM.
//! - We want to avoid requiring on a database server as an additional system dependency for ease of deployment and reduced overhead.
```

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,191 @@
...
+//!
+//! ### Implementation
+//!
+//! - Have a pool of tempfiles that can grow to match the worker thread
```
mattyg, line 13: I'm confused by this line -- it sounds like the max number of tempfiles is the worker_thread_count?

neonphog: Yes exactly. Since we're not using async code, the items from the pool can only be checked out by an active thread, and then they are returned before that thread can check out another. So the worker thread count is the *implicit* max count of active temp files that can be checked out from the pool.

There is a nuance here, however: when the tempfiles roll over from size, they are not re-added to the pool. So the pool will never grow beyond that, but while there are still read refs held for these other items, they will remain in memory.

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        for _ in 0..COUNT {
+            let (wrote, r) = recv.recv().unwrap();
+            let mut read = Vec::new();
+            r.read(&mut read).unwrap();
+            assert_eq!(wrote.as_bytes(), read);
+        }
+    }
+}
```
mattyg, line 193: I assume we want a lot testing more here, but i assume you know that :)

neonphog: @mattyg - I'm considering the testing in [Testing #11](https://github.com/holochain/kitsune2/pull/11) as sufficient for covering this. If you don't agree, please let me know what else you'd like to see here before approving : )

crates/bootstrap_srv/src/space.rs
```diff
@@ -0,0 +1,228 @@
...
+            .collect::<Vec<_>>();
+
+        // if we still have a new info to add, it must not have matched
+        // any existing ones
+        if let Some((parsed, store)) = new_info {
+            if !parsed.is_tombstone {
+                // if we are full, follow the "default" strategy rule of
+                // deleting the half-way info
+                if list.len() >= max_entries {
+                    list.remove(max_entries / 2);
```
mattyg, line 212: would it be clearer to have an explicit "strategy" defined somewhere and this function to call it?

neonphog: For sure eventually this will need to change. But I'm not sure the scope of the change that will be required until we go to actually design the new strategy... so I'm inclined to keep this simple until needed.

crates/bootstrap_srv/src/space.rs
mattyg: In that case do you want to define a SpaceId or similar wrapper type in this crate just for readability?

crates/bootstrap_srv/src/space.rs
mattyg: I see so we need the write lock for a long time because the re-write code is processing and mutating the data in one go, rather than constructing a new list with only a read lock, then at the end getting the write lock and writing it all at once.

And there is no downside to allowing readers at the same time because a single entry will be mutated atomically, and so there's no risk of a reader getting some partially-written entry change? (forgive my ignorance)

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
neonphog: [Done](https://github.com/holochain/kitsune2/issues)

crates/bootstrap_srv/src/store.rs
mattyg: whoops my bad

crates/bootstrap_srv/src/space.rs
neonphog: Since the infos are only stored on disk, the logic to update them means reading all the infos individually then doing any reorganization of the list of read references needed and then updating the specific reference list in memory that the read requests are cloning.

Since the read references themselves are not changing at all, it's that specific list in memory that readers are cloning that is okay for them to keep cloning. They will just get the older list (that may contain some recently expired infos, but that's okay).

Then that is locked and swapped out atomically, so thereafter read requests will get the updated list.

(no worries, understanding is what code review is all about! : )

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+    cont: Arc<std::sync::atomic::AtomicBool>,
+    workers: Vec<std::thread::JoinHandle<std::io::Result<()>>>,
+    addr: std::net::SocketAddr,
+}
+
+impl Drop for BootstrapSrv {
+    fn drop(&mut self) {
+        self.cont.store(false, std::sync::atomic::Ordering::SeqCst);
+        for worker in self.workers.drain(..) {
+            let _ = worker.join().expect("Failure shutting down worker thread");
```
zippy, line 23: Is there any value in passing the error through here?

neonphog: What would you like to pass this through? This is a drop impl, we don't have many options. Technically you're not supposed to panic in drop impls, you'll get a stack dump that is not very useful... I'm open to other suggestions... but rust's handling of panics in other threads is pretty poor, I haven't really heard of a better pattern.

crates/bootstrap_srv/src/server.rs
neonphog: [Created Issue](https://github.com/holochain/kitsune2/issues/17)

crates/bootstrap_srv/src/space.rs
neonphog: Sure, I can start with just a typedef for code clarity. I don't think there's any need for a newtype yet.

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+//!   max expiration time on the infos they contain (30 minutes).
+
+use std::sync::{Arc, Mutex};
+
+/// How large we should allow individual tempfiles to grow.
+/// This is a balance between using up too much disk space
+/// and having too many file handles open.
+///
+/// Start with 10MiB?
+const MAX_PER_TEMPFILE: u64 = 1024 * 1024 * 10;
```
zippy, line 38: Should this also be in the config?

neonphog: The pressure leading to wanting this larger is the inefficiency of having multiple tempfiles open. This lets us store more than 10_000 entries in one file (because infos are often much less than 1024 bytes long). It doesn't feel like we should need this to be much larger.

The pressure leading to wanting this smaller is the constraints of disk space. On a 16 core system (which cloud servers will likely not have anywhere near that) the default production config will have 64 active write tempfiles, giving 64MiB. Even if we have 10 older generations still around, that's only 640 MiB. Do we think servers will ever have that little disk space? Or to look at it from the opposite perspective, let's say the cloud server provides 10GiB of disk space. We can store 10,485,760 agent records at once, even if many of those are defunct, waiting to be cleaned up.

I don't *feel* like dev-ops folks are going to understand the consequences of making changes to this number. Providing the option to may just be confusing, and not practically useful. What do you think?

P.S. Re-reading this post, I'm not sure I even described it without being confusing...

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,333 @@
...
+        }
+
+        // validate expires_at is not before (or equal to) created_at
+        if info.expires_at <= info.created_at {
+            return Err(std::io::Error::other("InvalidExpiresAt"));
+        }
+
+        // validate expires_at is not more than 30 min after created_at
+        if info.expires_at - info.created_at
+            > (std::time::Duration::from_secs(60 * 30).as_micros() as i64)
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

crates/bootstrap_srv/src/server.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

crates/bootstrap_srv/src/server.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

crates/bootstrap_srv/src/server.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

crates/bootstrap_srv/src/space.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

crates/bootstrap_srv/src/space.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/263d5ab3f9e742b19c46ce7a11d370205ad904ae

jost-s (commented):
Seems great! Pruning is tested in the other PR - so just a couple of questions.

crates/bootstrap_srv/src/server.rs
neonphog: It needn't be the main thread if the server were passed to a different thread. But yes, whichever thread the server instance variable is dropped on will panic.

crates/bootstrap_srv/src/parse.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// encoded
+    pub encoded: String,
+
+    /// signature
+    pub signature: ed25519_dalek::Signature,
+}
+
+impl ParsedEntry {
+    /// Parse entry from a slice.
+    pub fn from_slice(slice: &[u8]) -> std::io::Result<Self> {
```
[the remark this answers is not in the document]
neonphog: Good call, done: https://github.com/holochain/kitsune2/pull/10/commits/a4ac1e1156bc1026f9df6e19743edb0b22ad1e15

crates/bootstrap_srv/src/server.rs
jost-s: Won't that stop the whole server when it's only about dropping workers?

ThetaSinner: I'd suggest that this doesn't have to be a drop implementation here. The `main` program is doing the drop explicitly but you only need to create one instance of this per program, so it could equally be a `shutdown` method that can return an error?

ThetaSinner: A drop implementation that calls `shutdown().unwrap()` would be good, to make sure that the user gets some attempt at shutdown if they forget to call it

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        // validate expires_at is not more than 30 min after created_at
+        if info.expires_at - info.created_at
+            > EXPIRES_AT_DURATION_MAX_ALLOWED_MICROS
+        {
+            return Err(std::io::Error::other("InvalidExpiresAt"));
+        }
+
+        // validate signature (do this at the end because it's more expensive
```
ThetaSinner, line 261:
```suggestion
        // validate signature (do this at the end because it's more expensive)
```

crates/bootstrap_srv/src/store.rs
ThetaSinner, line 1: This module is a little scary to me :) The documentation makes sense, and I can follow the code, I'm not sure how much I can asses the behavior

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -0,0 +1,49 @@
...
+    .unwrap();
+
+    let srv = BootstrapSrv::new(config);
+
+    let _ = recv.recv();
+
+    println!("Terminating...");
+    drop(srv);
+    println!("Done.");
+    std::process::exit(0);
```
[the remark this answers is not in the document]
neonphog: No, the ctrl-c hook gives control of ending the process to the process itself.

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+        };
+        let read_clone = writer.read_clone.clone();
+        let offset = writer.writer.stream_position()?;
+        let length = content.len();
+
+        writer.writer.write_all(content)?;
+        writer.writer.sync_data()?;
+
+        if offset + (length as u64) < MAX_PER_TEMPFILE {
+            self.writer_pool.lock().unwrap().push_back(writer);
```
ThetaSinner, line 134: This function reads as though if it's called fast enough across a number of threads, it will keep having to allocate new temp files. Because each call will lock/remove and then write the new content and then lock/insert, any time the queue is empty, a new file is created. Should the number of files be bounded here? That is how I understood the documentation above

ThetaSinner (commented):
Still working through the store, will pick up after meetings

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+//! - Designate one file handle only per pool entry as writable.
+//! - All other handles will be readonly.
+//! - After writing an entry, return a reference with a readonly file handle
+//!   to the tempfile that was written to, and an offset/length to
+//!   allow readback of exactly the bytes written for that entry.
+//! - Once a tempfile reaches a certain size, drop the write handle and
+//!   open a new tempfile for writing.
+//! - The older read handles will persist the existence of the older tempfiles
+//!   until the last read reference is dropped, at which point the tempfile
+//!   will be cleaned up by the os.
```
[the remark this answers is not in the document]
neonphog: The tempfile drop implementation indeed cleans up the tempfile. If that fails for any reason (panic on the thread) then it is up to the os to clean it up. That's completely [system specific](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#operating-system-specific-concerns).

crates/bootstrap_srv/src/config.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// The maximum agent info entry count per space.
+    ///
+    /// All entries will be returned in a get space request, so
+    /// this count should be low enough to reasonably send this response
+    /// over http without needing pagination.
+    ///
+    /// Defaults:
+    /// - `testing = 32`
+    /// - `production = 32`
```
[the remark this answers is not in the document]
neonphog: No, this is just for bootstrap. I'd prefer to handle more spaces and less agents per space than the other way around. Once they are in the space, they can get additional peers by gossip.

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+//! - We can trust these files will cycle through at a rate similar to the
+//!   max expiration time on the infos they contain (30 minutes).
```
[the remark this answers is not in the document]
neonphog: Almost. We're writing the individual agent infos to the store, and keeping the space list with only the references in memory. So when we update, only new infos are written to the disk. This avoids us re-writing the same infos over and over every time the space updates.

Yes indeed, the max expires_at is 30 min past the created_at and the max in the future created_at can be is 3 min, so the max we'll hold an individual info is 33 min plus the interval the prune task runs on.

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// buffer with the bytes read from the store.
+    pub fn read(&self, buf: &mut Vec<u8>) -> std::io::Result<()> {
+        use std::io::{Read, Seek};
+        let mut reader = self.read_clone.reopen()?;
+        reader.seek(std::io::SeekFrom::Start(self.offset))?;
+
+        buf.reserve(self.length);
+
+        unsafe {
+            let offset = buf.len();
```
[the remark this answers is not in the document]
neonphog: I thought I did : ) "This will extend the provided buffer with the bytes read from the store."

crates/bootstrap_srv/src/store.rs
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /// Read the content of this entry. This will extend the provided
+    /// buffer with the bytes read from the store.
+    pub fn read(&self, buf: &mut Vec<u8>) -> std::io::Result<()> {
+        use std::io::{Read, Seek};
+        let mut reader = self.read_clone.reopen()?;
+        reader.seek(std::io::SeekFrom::Start(self.offset))?;
+
+        buf.reserve(self.length);
```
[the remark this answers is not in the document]
neonphog: Yep, exactly.

crates/bootstrap_srv/src/store.rs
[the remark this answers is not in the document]
ThetaSinner: I'd optionally include that in the documentation here but maybe it's too much detail

crates/bootstrap_srv/src/store.rs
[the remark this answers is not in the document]
ThetaSinner: I see, so I'd slightly misunderstood what's being written. So that's more efficient than I was thinking :)

I see the logic of not holding the info for more than 33 minutes ish. However, that doesn't prevent rapid agent updates right? So I could send a new agent info every 5ms and it would force the space required by temporary files to grow rapidly. It would expire after 33 minutes so the files could be dropped but still, you could add quite a load before the first ones start expiring

crates/bootstrap_srv/src/store.rs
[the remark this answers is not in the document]
ThetaSinner: I see! Oh I get how this works now with the code in a file above with `Space::read`, for building an array from the content. Okay cool, I get it now :)

crates/bootstrap_srv/src/store.rs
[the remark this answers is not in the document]
neonphog: Your logic is correct. I'd say this will be mitigated by rate limiting.

We could potentially use a different strategy such as having a tempfile per space, but that inevitably increases system resource usage (like max file handles).

crates/bootstrap_srv/src/store.rs
ThetaSinner: Right, that makes sense with the docs further up. Thank you

crates/bootstrap_srv/src/store.rs
neonphog: I agree. It would be nice to find a way to test it, at least a little bit. Perhaps if I made the max file size configurable at least internally, and then wrote some tests that explicitly dropped read handles and checked if the files still exist on-disk? I don't know if that would be flaky or not...

ThetaSinner: I've been through this in detail now and I'm much happier. Thank you for letting me try and fail to pick it apart so I could make sense of it :)

crates/bootstrap_srv/src/store.rs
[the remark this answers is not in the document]
ThetaSinner: Cool, yes that sounds like a rate limiting issue - happy for this to be solved elsewhere :)

crates/bootstrap_srv/src/server.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/ebac6ff5dcaff4f2b78c66fab4e696e75c504847

crates/bootstrap_srv/src/parse.rs
```diff
@@ -0,0 +1,83 @@
+/// An entry with known content.
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/85dffc79aec3d3d5493b4cab9c9ddabce0bc6bfb

crates/bootstrap_srv/src/server.rs
```diff
@@ -0,0 +1,339 @@
+//! bootstrap http server types.
+
+use std::sync::Arc;
+
+use crate::*;
+use tiny_http::*;
+
+/// Don't allow created_at to be greater or less than this far away from now.
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/10/commits/b407784ebfeca5aa999c249cce3f2871ed1350ff

mattyg (approved):

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
neonphog: I'm going to decline this because if I change anything it'll dismiss the approval. But also because I don't really care about todos in the code. Feel free to make a PR to remove them! : )

jost-s: > But also because I don't really care about todos in the code.

Exactly, no one does! :grinning:

PR #11, opened 2024-11-21
Kitsune2 Bootstrap Server--Testing

neonphog:
See the [Prep #9](https://github.com/holochain/kitsune2/pull/9) and [Core #10](https://github.com/holochain/kitsune2/pull/10) PRs leading up to this.

This PR adds API endpoint tests to the kitsune2 bootstrap server.

### General notes about this kitsune2 bootstrap server effort:

- The server is designed to be robust for production use, but some of the final touches, such as rate limiting and TLS have yet to be implemented. This server is ready for use now in local development and testing of kitsune2.
- The server is designed to be light-weight both in dependency usage, and in terms of maintenance understanding. As such, it does NOT use async code, but instead multi-threaded blocking code, as dictated by the chosen tiny_http server library.


crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,654 @@
...
+fn multi_thread_stress() {
+    let s = BootstrapSrv::new(Config::testing()).unwrap();
+    let addr = s.listen_addr();
+
+    let start = std::time::Instant::now();
+
+    // the testing config has 2 worker threads.
+    // Read and write with more than that.
+    const T_W_COUNT: u32 = 16;
+    const T_R_COUNT: u32 = 32;
```
zippy, line 512: I'm curious about what we would expect in production environments.

On my machine when I bump the T_R_COUNT above 300 & start getting this error:

```
thread '<unnamed>' panicked at crates/bootstrap_srv/src/test.rs:561:18:
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: Transport { kind: Io, message: None, url: Some(Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(37587), path: "/bootstrap/2o79pTXHaK1FTPZeBiJo2lCgXW_P0ULjX_5Div_2qxU/ovovSjVboukHpTAJ6eN8rd96x-ZqCLoHYx9VMHKz8kw", query: None, fragment: None }), source: Some(Custom { kind: ConnectionReset, error: Transport(Transport { kind: Io, message: Some("Error encountered in the status line"), url: None, source: Some(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }) }) }) } }
```

neonphog: This will also be mitigated somewhat when we implement rate limiting, and if we front the server with cloudflare.

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,654 @@
...
+        is_tombstone: true,
+        ..Default::default()
+    }
+    .call()
+    .unwrap();
+
+    let addr = format!("http://{:?}/bootstrap/{}", s.listen_addr(), S1);
+    let res = ureq::get(&addr).call().unwrap().into_string().unwrap();
+    let res: Vec<DecodeAgent> = serde_json::from_str(&res).unwrap();
+    assert_eq!(1, res.len());
```
mattyg, line 244: This should be 0, right? Or am I misunderstanding?

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,654 @@
...
+        ..Default::default()
+    }
+    .call()
+    .unwrap_err();
+
+    assert!(err.to_string().contains("InvalidExpiresAt"));
+}
+
+#[test]
+fn reject_expired_at_before_created_at() {
```
mattyg, line 388: Maybe another test for rejecting created_at == expired_at

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,654 @@
...
+     * pruner will have already run on it.
+     */
+
+    std::thread::sleep(std::time::Duration::from_secs(1));
+
+    let res = ureq::get(&addr).call().unwrap().into_string().unwrap();
+    let res: Vec<DecodeAgent> = serde_json::from_str(&res).unwrap();
+
+    assert_eq!(1, res.len());
+}
```
mattyg, line 654: do we want a test for InvalidPathSegment error?

mattyg, line 654: also a test for InvalidAgentPubKey

crates/bootstrap_srv/src/test.rs
neonphog: This is a test to ensure that a PUT of an info with a created_at earlier than the one the server is tracking does not cause the tracked one to be deleted. So it should be a 1.

mattyg: ah derp

crates/bootstrap_srv/src/test.rs
neonphog: I'm not sure there *is* an invalid SpaceId. One app can even claim the empty one : ) You can get a base64 decode error. Do you think we need a test for that?

The agent key will similarly generate base64 decode errors, or signature validation errors if it is not a valid pub key.

neonphog: Oh, I see you mean the error generated if they don't even specify the right count of path segments. Yes I can test that.

neonphog: I think this will just be a signature error, but I can test that too :+1:

crates/bootstrap_srv/src/test.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/11/commits/8cfe688c6779c30e6b214b4bbcfb08eaa3da4652

crates/bootstrap_srv/src/test.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/11/commits/8cfe688c6779c30e6b214b4bbcfb08eaa3da4652

neonphog: You were right, it gives InvalidAgentPubKey! - https://github.com/holochain/kitsune2/pull/11/commits/8cfe688c6779c30e6b214b4bbcfb08eaa3da4652

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,748 @@
...
+#[test]
+fn tombstone_will_not_put() {
+    let s = BootstrapSrv::new(Config::testing()).unwrap();
+
+    let _ = PutInfo {
+        addr: s.listen_addr(),
+        is_tombstone: true,
+        ..Default::default()
+    }
+    .call()
```
jost-s, line 220: Return an error so that the caller knows it was unsuccessful? This would go into the server implementation obviously.
Checking there, `call` should return an error here, because `created_at` is too far in the past. Yet the result to `call` is unwrapped. What's happening?

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,748 @@
...
+    .unwrap();
+
+    let _ = PutInfo {
+        addr: s.listen_addr(),
+        created_at: now()
+            - std::time::Duration::from_secs(60).as_micros() as i64,
+        is_tombstone: true,
+        ..Default::default()
+    }
+    .call()
```
jost-s, line 246: Same as above, failed attempts return an error?

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,748 @@
...
+
+    assert!(err.to_string().contains("InvalidCreatedAt"));
+}
+
+#[test]
+fn reject_expired() {
+    let s = BootstrapSrv::new(Config::testing()).unwrap();
+
+    let expires_at = crate::now() - 500;
+    let created_at = crate::now() - 1500;
```
jost-s, line 432: created_at can be valid here, right?

```suggestion
    let created_at = crate::now();
```

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,748 @@
...
+        ]
+    );
+
+    for _ in 0..32 {
+        put_info();
+    }
+
+    let res = get();
+
+    assert_eq!(
```
jost-s, line 586:
```suggestion
   // Agents put beyond MAX_STORAGE will be insert at MAX_STORAGE  / 2, replacing the existing agent at that place. That is here 32 / 2 = 16, so from 17 onward agents > 32 are replaced. With 32 agents this happens twice, first from 33-47 and then from 48-63.
    assert_eq!(
```

jost-s (commented):
Nice test suite! I'll double-check tomorrow with fresh eyes, but this looks quite good to me.

crates/bootstrap_srv/src/test.rs
neonphog: Did you write this comment in the right test? The created_at here is valid. A correct tombstone just ensures there are no entries in the bootstrap server for this agent. There are none in this case already, but that's still a success here.

crates/bootstrap_srv/src/test.rs
neonphog: I don't want to trigger the other failure case of rejecting puts where the expired_at <= created_at, so I need the created_at here to be before the expired_at here.

crates/bootstrap_srv/src/test.rs
neonphog:
```suggestion
    // Agents put beyond MAX_STORAGE will be insert at MAX_STORAGE  / 2,
    // replacing the existing agent at that place. That is here 32 / 2 = 16,
    // so from 17 onward agents > 32 are replaced. With 32 agents this happens
    // twice, first from 33-47 and then from 48-63.
    assert_eq!(
```

:+1: - but reformat to 80 character width

jost-s (commented):
Going by the spec, the only test that's missing is a test with a payload exceeding 1024 bytes.

You could edit the spec that new agent infos when storage is satured are pushed to the position MAX_INFOS / 2.

neonphog:
> Going by the spec, the only test that's missing is a test with a payload exceeding 1024 bytes.

Indeed, that's a good test to have. Will add :+1:

> You could edit the spec that new agent infos when storage is satured are pushed to the position MAX_INFOS / 2.

@jost-s - This is already in the spec, here: https://github.com/holochain/kitsune2/blob/accbf23e2fd3a57767f65c45fc974f181791416a/crates/bootstrap_srv/src/lib.rs#L143-L144

neonphog:
@jost-s 

> > Going by the spec, the only test that's missing is a test with a payload exceeding 1024 bytes.
>
> Indeed, that's a good test to have. Will add :+1:

Done: https://github.com/holochain/kitsune2/pull/11/commits/e241587761e415614cc68e36fb8cb6a84a6c0fc9

crates/bootstrap_srv/src/test.rs
jost-s: Ah, I had misunderstood that. All good.

crates/bootstrap_srv/src/test.rs
jost-s: Yup, understood.

jost-s:
> This is already in the spec, here:

Just to clarify that "push onto the stack" there means that it's inserted at that position. Push sounds like added to the beginning or end. 

"If the store count is >= MAX_INFOS, the server MUST delete the entry at MAX_INFOS / 2 and then push the info onto the stack." 

That's nitpicky... :grimacing:

jost-s (dismissed):
:partying_face:

neonphog:
@jost-s

> Just to clarify that "push onto the stack" there means that it's inserted at that position. Push sounds like added to the beginning or end.

No, it's not added at the midpoint. It is pushed onto the end of the stack. If it was inserted into the midpoint, we'd just keep replacing that midpoint item... but we actually want the second half of the buffer to rotate with the most recently pushed agent infos, so the midpoint is deleted, but the new one is pushed onto the end.

jost-s:
Ah yes, otherwise the order of the test numbers would be inverted.

crates/bootstrap_srv/src/test.rs
```diff
@@ -0,0 +1,773 @@
...
+#[test]
+fn multi_thread_stress() {
+    let s = BootstrapSrv::new(Config::testing()).unwrap();
+    let addr = s.listen_addr();
+
+    let start = std::time::Instant::now();
+
+    // the testing config has 2 worker threads.
+    // Read and write with more than that.
+    const T_W_COUNT: u32 = 16;
```
jost-s, line 630: We probably won't change these (ever?), but if we did, would it be helpful to make these multiples of the configured number of workers in the testing config? I.e. `let t_w_count = config.worker_thread_count * 8`.

neonphog: I'm not sure there's any particular reason to have an even multiple of the worker threads. It should work totally fine to have it be any value.

jost-s: No, I misphrased that. Instead of a hardcoded `const` use a calculated number with `config.worker_thread_count`, e.g. `let t_w_count = config.worker_thread_count + x`.

neonphog: Ah, I understand : ) Done: https://github.com/holochain/kitsune2/pull/11/commits/826d7de0f4cdaaab5125ce0000f166198ef2aa41

jost-s: One more suggestion:
```rust
let config = Config::testing();
let s = BootstrapSrv::new(config.clone()).unwrap();
let addr = s.listen_addr();

let start = std::time::Instant::now();

// the testing config has a small number worker threads (currently 2).
// Read and write with more than that.
let worker_count = config.worker_thread_count as u32;
```

neonphog: Done: https://github.com/holochain/kitsune2/pull/11/commits/d803ec7e3599d8d7900f2bccce0171303a6ce6ea

jost-s: Thanks for your patience here =)

jost-s (approved):
Sweet!

PR #12, opened 2024-11-21
simplified mvp module factory config pattern

neonphog:
This is a greatly simplified version of the config pattern I had put in the kitsune spike, with the added benefit that modules can make public their config structs so they are available in rust docs for understanding how to use them.

@ThetaSinner - I was considering your request for runtime changeable configs and came to the conclusion that would be better handled via some other mechanism. Namely we could specify a `default_` prefixed version of that setting here in the immutable config, and then expose a function on the module api to update runtime settings.

I was just looking through the [current tuning params](https://docs.rs/kitsune_p2p_types/latest/kitsune_p2p_types/config/tuning_params_struct/struct.KitsuneP2pTuningParams.html) and started worrying about how poorly legacy kitsune would probably handle many of these things changing at runtime : )

But, I'm totally still up for discussing it if you feel differently!

### DIFF
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -60,7 +60,7 @@ impl Config {
     /// to extract a module config. Note that this config is loaded from
     /// disk and can be edited by humans, so the serialization on the module
     /// config should be tolerant to missing properties, setting sane defaults.
-    /// A module may choose to warn about missing or extraneous properties.
+    /// A module may choose to warn about missing properties and should warn about extraneous properties.
     pub fn get_module_config<M: ModConfig>(
         &self,
         module_name: &str,
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -1,0 +1,176 @@
+//! Types for use when configuring kitsune2 modules.
+
+/// helper transcode function
+fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
+    s: &S,
+) -> std::io::Result<D> {
+    Ok(serde_json::from_str(&serde_json::to_string(s)?)?)
+}
+
+/// Denotes a type used to configure a specific kitsune2 module.
+///
+/// Note, the types defined in this struct are specifically for configuration
+/// that cannot be changed at runtime, the likes of which might be found
+/// in a configuration file.
+///
+/// If a specific module has a config that can be changed at runtime, the
+/// component found in this type might be a `default_` prefixed version
+/// of it, then the runtime value can be altered through different means.
+///
+/// It is highly recommended that you expose this struct in your module
+/// docs to help devs using your module understand how to configure it.
+pub trait ModConfig:
+    'static
+    + Sized
+    + Default
+    + std::fmt::Debug
+    + serde::Serialize
+    + serde::de::DeserializeOwned
+    + Send
+    + Sync
+{
+}
+
+/// Kitsune configuration.
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+pub struct Config(serde_json::Map<String, serde_json::Value>);
+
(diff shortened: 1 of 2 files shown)


mattyg (approved):
happy to see per-module config for clarity

crates/api/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+impl Config {
+    /// When kitsune2 is generating a default or example configuration
+    /// file, it will pass a mutable reference of this config struct to
+    /// the module factories that are configured to be used. Those factories
+    /// should call this function any number of times to add any default
+    /// configuration parameters to that file.
+    pub fn add_default_module_config<M: ModConfig>(
+        &mut self,
+        module_name: String,
+    ) -> std::io::Result<()> {
```
ThetaSinner, line 47: Use `K2Result` here? There is an io::Error from the `tc` function but the `module_name` check could be a `K2Error`

crates/api/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+        Ok(())
+    }
+
+    /// When kitsune2 is initializing, it will call the factory function
+    /// for all of its modules with an immutable reference to this config
+    /// struct. Each of those modules may choose to call this function
+    /// to extract a module config. Note that this config is loaded from
+    /// disk and can be edited by humans, so the serialization on the module
+    /// config should be tolerant to missing properties, setting sane defaults.
+    /// A module may choose to warn about missing or extraneous properties.
```
ThetaSinner, line 63:
```suggestion
    /// A module may choose to warn about missing properties and should warn about extraneous properties.
```

ThetaSinner: I really dislike it when I can't spell and the code doesn't tell me I gave it a property it doesn't know about :)

crates/api/src/config.rs
```diff
@@ -0,0 +1,176 @@
...
+            p_b: String,
+        }
+
+        impl ModConfig for Mod1 {}
+
+        #[derive(Debug, serde::Serialize, serde::Deserialize, PartialEq)]
+        #[serde(default)]
+        struct Mod2 {
+            #[serde(rename = "#p_c", skip_deserializing)]
+            doc_p_a: &'static str,
```
ThetaSinner, line 97: I'd prefer a JSON schema with a field description. Yes it's a little further from the code but it's usable when editing the document, without having to commit the help text to source control

ThetaSinner (commented):
I like the pattern!

I would prefer to have YAML as the data format though. For being able to do things like

```yaml
gossip:
    # Currently only full and empty are supported by Holochain stable, revert to "dynamic" when supported again
    arc_clamping: full
```

neonphog:
@ThetaSinner 

> I would prefer to have YAML as the data format though.

Despite the code, it's actually format agnostic. Internally it's using json as the transcode format, but the exported api `Config` struct is just serialize / deserialize, so when we're writing the conductor config it can use yaml.

The serde_yaml crate is deprecated and its `Value` memory struct was never as good as the one in serde_json. I wish there were a more full featured serde memory struct that *all* the serialization crates could depend on so they didn't each have to re-implement it...

crates/api/src/config.rs
neonphog: I'm thinking about this from the writing the example/default config file perspective. It's nice if the comments *actually* get written there so humans can have them right next to where they are manually editing the file.

jost-s (approved):
LGTM

crates/api/src/config.rs
neonphog: Well, this is just an example of how we *could* have it look. I'll not add doc fields like this to any of my config structs until we determine it is desirable.

PR #18, opened 2024-11-22
use K2Error in existing k2 code

neonphog:
Heh, I should have searched for uses of std::io::Error before not including this [in the previous pr](https://github.com/holochain/kitsune2/pull/8), but in any case: here is implementing the usage of K2Error in the existing code in this repo.


jost-s (approved):
K2... makes me think of the skates brand.

PR #19, opened 2024-11-22
PeerStore API and MemPeerStore Implementation

neonphog:
RESOLVES https://github.com/holochain/holochain/issues/4434

### DIFF
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -1,12 +1,17 @@
 //! Types for use when configuring kitsune2 modules.
 
+use crate::*;
 use std::collections::BTreeMap;
 
 /// helper transcode function
 fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
     s: &S,
-) -> std::io::Result<D> {
-    Ok(serde_json::from_str(&serde_json::to_string(s)?)?)
+) -> K2Result<D> {
+    serde_json::from_str(
+        &serde_json::to_string(s)
+            .map_err(|e| K2Error::other_src("encode", e))?,
+    )
+    .map_err(|e| K2Error::other_src("decode", e))
 }
 
 /// Denotes a type used to configure a specific kitsune2 module.
@@ -46,9 +51,9 @@ impl Config {
     pub fn add_default_module_config<M: ModConfig>(
         &mut self,
         module_name: String,
-    ) -> std::io::Result<()> {
+    ) -> K2Result<()> {
         if self.0.contains_key(&module_name) {
-            return Err(std::io::Error::other(format!(
+            return Err(K2Error::other(format!(
                 "Refusing to overwrite conflicting module name: {module_name}"
             )));
         }
@@ -62,11 +67,11 @@ impl Config {
     /// to extract a module config. Note that this config is loaded from
     /// disk and can be edited by humans, so the serialization on the module
     /// config should be tolerant to missing properties, setting sane defaults.
-    /// A module may choose to warn about missing or extraneous properties.
+    /// A module may choose to warn about missing properties and should warn about extraneous properties.
     pub fn get_module_config<M: ModConfig>(
         &self,
         module_name: &str,
-    ) -> std::io::Result<M> {
+    ) -> K2Result<M> {
         self.0
             .get(module_name)
             .map(tc)
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -1,5 +1,7 @@
 //! Types for use when configuring kitsune2 modules.
 
+use std::collections::BTreeMap;
+
 /// helper transcode function
 fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
     s: &S,
@@ -33,7 +35,7 @@ pub trait ModConfig:
 
 /// Kitsune configuration.
 #[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
-pub struct Config(serde_json::Map<String, serde_json::Value>);
+pub struct Config(BTreeMap<String, serde_json::Value>);
 
 impl Config {
     /// When kitsune2 is generating a default or example configuration
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -1,0 +1,176 @@
+//! Types for use when configuring kitsune2 modules.
+
+/// helper transcode function
+fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
+    s: &S,
+) -> std::io::Result<D> {
+    Ok(serde_json::from_str(&serde_json::to_string(s)?)?)
+}
+
+/// Denotes a type used to configure a specific kitsune2 module.
+///
+/// Note, the types defined in this struct are specifically for configuration
+/// that cannot be changed at runtime, the likes of which might be found
+/// in a configuration file.
+///
+/// If a specific module has a config that can be changed at runtime, the
+/// component found in this type might be a `default_` prefixed version
+/// of it, then the runtime value can be altered through different means.
+///
+/// It is highly recommended that you expose this struct in your module
+/// docs to help devs using your module understand how to configure it.
+pub trait ModConfig:
+    'static
+    + Sized
+    + Default
+    + std::fmt::Debug
+    + serde::Serialize
+    + serde::de::DeserializeOwned
+    + Send
+    + Sync
+{
+}
+
+/// Kitsune configuration.
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+pub struct Config(serde_json::Map<String, serde_json::Value>);
+
+impl Config {
+    /// When kitsune2 is generating a default or example configuration
+    /// file, it will pass a mutable reference of this config struct to
+    /// the module factories that are configured to be used. Those factories
+    /// should call this function any number of times to add any default
+    /// configuration parameters to that file.
+    pub fn add_default_module_config<M: ModConfig>(
+        &mut self,
+        module_name: String,
+    ) -> std::io::Result<()> {
+        if self.0.contains_key(&module_name) {
+            return Err(std::io::Error::other(format!(
+                "Refusing to overwrite conflicting module name: {module_name}"
+            )));
+        }
+        self.0.insert(module_name, tc(&M::default())?);
+        Ok(())
+    }
+
+    /// When kitsune2 is initializing, it will call the factory function
+    /// for all of its modules with an immutable reference to this config
+    /// struct. Each of those modules may choose to call this function
+    /// to extract a module config. Note that this config is loaded from
+    /// disk and can be edited by humans, so the serialization on the module
+    /// config should be tolerant to missing properties, setting sane defaults.
+    /// A module may choose to warn about missing or extraneous properties.
+    pub fn get_module_config<M: ModConfig>(
+        &self,
+        module_name: &str,
+    ) -> std::io::Result<M> {
+        self.0
+            .get(module_name)
+            .map(tc)
+            .unwrap_or_else(|| Ok(M::default()))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn config_usage_example() {
+        #[derive(
+            Debug, Default, serde::Serialize, serde::Deserialize, PartialEq,
+        )]
+        struct Mod1 {
+            #[serde(default)]
+            p_a: u32,
+            #[serde(default)]
+            p_b: String,
+        }
+
+        impl ModConfig for Mod1 {}
+
+        #[derive(Debug, serde::Serialize, serde::Deserialize, PartialEq)]
+        #[serde(default)]
+        struct Mod2 {
+            #[serde(rename = "#p_c", skip_deserializing)]
+            doc_p_a: &'static str,
+            #[serde(default)]
+            p_c: u32,
+            #[serde(default)]
+            p_d: String,
+        }
+
+        impl Default for Mod2 {
+            fn default() -> Self {
+                Self {
+                    doc_p_a: "This is a potential pattern for how to document config props.",
+                    p_c: 0,
+                    p_d: "".into(),
+                }
+            }
+        }
+
+        impl ModConfig for Mod2 {}
+
+        let mut config = Config::default();
+        config
+            .add_default_module_config::<Mod1>("mod1".into())
+            .unwrap();
+        config
+            .add_default_module_config::<Mod2>("mod2".into())
+            .unwrap();
+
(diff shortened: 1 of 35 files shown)


crates/api/src/peer_store.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// to gossip (and also the local agent). If there are multiple
+    /// local agents in this space, you'll need to call this function
+    /// multiple times and union the results.
+    fn get_overlapping_storage_arc(
+        &self,
+        arc: agent::BasicArc,
+    ) -> BoxFut<'_, K2Result<Vec<Arc<agent::AgentInfoSigned>>>>;
+
+    /// Get a list of agents sorted by nearness to a target basis location.
+    /// If the
```
mattyg, line 37: Missing something in this comment?

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,261 @@
...
+    match (a, b) {
+        (None, _) | (_, None) => false,
+        (Some((a_beg, a_end)), Some((b_beg, b_end))) => {
+            // The only way for there to be overlap is if
+            // either of a's start or end points are within b
+            // or either of b's start or end points are within a
+            calc_dist(a_beg, Some((b_beg, b_end))) == 0
+                || calc_dist(a_end, Some((b_beg, b_end))) == 0
+                || calc_dist(b_beg, Some((a_beg, a_end))) == 0
+                || calc_dist(b_end, Some((a_beg, a_end))) == 0
```
mattyg, line 255: Don't we only need the first 2 comparisons and the next 2 are redundant? Or am I missing a case?

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,261 @@
...
+struct Inner {
+    store: HashMap<AgentId, Arc<AgentInfoSigned>>,
+    no_prune_until: std::time::Instant,
+}
+
+impl Inner {
+    pub fn new(now_inst: std::time::Instant) -> Self {
+        Self {
+            store: HashMap::new(),
+            no_prune_until: now_inst + std::time::Duration::from_secs(10),
```
mattyg, line 103: Should this be configurable?

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,261 @@
...
+        }
+    }
+
+    fn do_prune(&mut self, now_inst: std::time::Instant, now_ts: Timestamp) {
+        self.store.retain(|_, v| v.expires_at > now_ts);
+
+        // we only care about not looping on the order of tight cpu cycles
+        // even a couple seconds gets us away from this.
+        self.no_prune_until = now_inst + std::time::Duration::from_secs(10)
```
mattyg, line 112: Should this be configurable?

crates/kitsune2/src/factories/mem_peer_store/test.rs
```diff
@@ -0,0 +1,283 @@
...
+            .collect::<Vec<_>>();
+
+        got.sort();
+
+        assert_eq!(exp, &got.as_slice());
+    }
+}
+
+#[test]
+fn fixture_get_near_location() {
```
mattyg, line 254: might be good to include zome 0 arc and expired agents and confirm they aren't included in results

mattyg (approved):

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,261 @@
+//! A production-ready memory-based peer store.
```
mattyg, line 1: Might be good to have some more docs about the architecture and rationale.

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,261 @@
...
+}
+
+impl MemPeerStore {
+    pub fn new() -> Self {
+        Self(Mutex::new(Inner::new(std::time::Instant::now())))
+    }
+}
+
+impl peer_store::PeerStore for MemPeerStore {
+    fn insert(
```
mattyg, line 55: So it looks like we keep all peers unless they are expired right?

Does it make sense to follow a similar approach to the bootstrap server of splitting the store between long-running and recently added nodes? And to limit the number of peers we store? Seems like the same rationale applies unless I'm missing something?

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: We need to check both sides, because they can fully contain each other, in which case one direction might not catch it:

```
|-A-B--B-A-|
vs
|-B-A--A-B-|
```

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Mmm, the selecting the peers is a higher level concern, but making them available is a this-level concern.

Basically, anyone we overlap storage arcs with we need to keep here, because we must gossip with all of them. We *could* drop agents outside that set, but I'd consider that a premature optimization, because it's actually helpful to have them in our store if we need to publish to or get from that neighborhood. We can get more fancy later about cleaning up the peer store if we find it's using too much memory.

crates/api/src/peer_store.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/a368c14881972682fbaf135d2bc702a8a41758b2

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/cc03e5be6c53848dec1c5c8e3fc5d500cb774e84

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/cc03e5be6c53848dec1c5c8e3fc5d500cb774e84

crates/kitsune2/src/factories/mem_peer_store/test.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/24bef62698948159e9e3ca5f0f11593b38c51eb1

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/44e4bf23e4d2e0b35f5a2bdcac51a5563775b1cb

I put it on the MemPeerStoreFactory because that is the only bit actually exported.

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,303 @@
...
+                } else {
+                    (loc - arc_end, arc_start - loc)
+                }
+            } else {
+                // this arc does not wrap
+
+                if loc >= arc_start && loc <= arc_end {
+                    return 0;
+                } else if loc < arc_start {
+                    (arc_start - loc, u32::MAX - arc_end + loc + 1)
```
ThetaSinner, line 276: Just to make sure I've got this understood correctly. The `u32::MAX - arc_end` is the distance from the end of the arc to the top of a u32, then `+ 1` wraps around to 0 and `+ loc` takes us back to our location?

I wonder if it might make this slightly clearer to organize the terms this way?

```suggestion
                    (arc_start - loc, (u32::MAX - arc_end + 1) + loc)
```

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,303 @@
...
+        out.into_iter()
+            .take(limit)
+            .map(|(_, v)| v.clone())
+            .collect()
+    }
+}
+
+/// Get the min distance from a location to an arc in a wrapping u32 space.
+/// This function will only return 0 if the location is covered by the arc.
+/// This function will return u32::MAX if the arc is not set.
```
jost-s, line 257:
```suggestion
/// This function will return u32::MAX if the arc is not set.
///
/// All possible cases:
///
/// s = arc_start
/// e = arc_end
/// l = location
///
/// Arc wraps around, loc >= arc_start
/// |----e-----------s--l--|
/// 0                          u32::MAX
///
/// Arc wraps around, loc <= arc_end
/// |-l--e-----------s-----|
/// 0                          u32::MAX
///
/// Arc wraps around, loc outside of arc
/// |----e----l------s-----|
/// 0                          u32::MAX
///
/// Arc does not wrap around, loc inside of arc
/// |---------s--l---e-----|
/// 0                          u32::MAX
///
/// Arc does not wrap around, loc < arc_start
/// |-----l---s------e-----|
/// 0                          u32::MAX
///
/// Arc does not wrap around, loc > arc_end
/// |---------s------e--l--|
/// 0                          u32::MAX
```

crates/kitsune2/src/factories/mem_peer_store.rs
jost-s: You could add this to the function doc.

jost-s (commented):
Looks fantastic! I appreciate the simple logic and all the tests that cover all specified cases.

One question about the general working of the mem peer store - the pruning is set up so that it's executed whenever the peer store is accessed. How will this play out in its usage? Will there be a periodic querying/storing of/to the store which will take care of regular pruning?

crates/kitsune2/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+edition = "2021"
+
+[dependencies]
+bytes = { workspace = true }
+kitsune2_api = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["full"] }
```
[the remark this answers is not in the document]
ThetaSinner: If possible yes, just what we need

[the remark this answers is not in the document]
neonphog: For dev-dependencies, I don't like to dither. Between "rt", the test "macros", and "timeouts" we always end up needing almost everything anyways. I 100% agree on being specific in the true `dependencies` section, but I'm hoping we can get away with not pulling in tokio as a true dependency at all. It'd be very nice if kitsune was runtime agnostic.

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        self.do_prune(now_inst, Timestamp::now());
+    }
+
+    pub fn insert(&mut self, agent_list: Vec<Arc<AgentInfoSigned>>) {
+        self.check_prune();
+
+        for agent in agent_list {
+            // Don't insert expired infos.
+            if agent.expires_at < Timestamp::now() {
```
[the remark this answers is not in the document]
neonphog: Yes! I'm usually so diligent about that... Thanks for catching this : )

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Sooo... the actual reason is that `u32::MAX` is the final index, not the count. We actually want the count in this arithmetic, but since the count doesn't fit inside a u32, we have to be careful where we add it to make sure we don't overflow.

The parens, for me, are a lateral step, and don't help or hinder my understanding of this, so I'm happy to change it, but I'll wait for your response before implementing it.

crates/api/src/config.rs
```diff
@@ -39,7 +40,7 @@ pub trait ModConfig:
 
 /// Kitsune configuration.
 #[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
-pub struct Config(serde_json::Map<String, serde_json::Value>);
+pub struct Config(BTreeMap<String, serde_json::Value>);
```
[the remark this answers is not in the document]
neonphog: The serde_json Map is an abstraction around either a standard collection on indexed map depending on features being enabled. It occurred to me that we might prefer the entries to be sorted alphabetically for the default config file generation.

crates/kitsune2/src/factories/mem_peer_store.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/051985b666b17dfb07d172005ed233b021111cfe

crates/kitsune2/src/factories/mem_peer_store.rs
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/19/commits/f383026d0cd2f6d6e948ee3aeab482dea3eeda48

jost-s:
@neonphog It looks like you overlooked this question https://github.com/holochain/kitsune2/pull/19#pullrequestreview-2468772402

crates/kitsune2/src/factories/mem_peer_store.rs
```diff
@@ -0,0 +1,363 @@
...
+        loc: u32,
+        limit: usize,
+    ) -> BoxFut<'_, K2Result<Vec<Arc<AgentInfoSigned>>>> {
+        let r = self.0.lock().unwrap().get_near_location(loc, limit);
+        Box::pin(async move { Ok(r) })
+    }
+}
+
+struct Inner {
+    config: MemPeerStoreConfig,
```
jost-s, line 131: Does the config need to be mutable/part of the inner state?

crates/kitsune2/src/factories/mem_peer_store.rs
ThetaSinner: Understood :) This isn't important enough to me to change. Now I've had a chance to puzzle out the logic, I'm happy I could figure it out again written either way

ThetaSinner (approved):

neonphog:
@jost-s 

> One question about the general working of the mem peer store - the pruning is set up so that it's executed whenever the peer store is accessed. How will this play out in its usage? Will there be a periodic querying/storing of/to the store which will take care of regular pruning?
>
> It looks like you overlooked this question

Whoops! There is not any explicit periodic task here. However, if no-one is accessing it, it doesn't harm anything to have the old records in memory. As soon as they do access it, they will be pruned. And in reality, the peer store was accessed so frequently, we had real performance data that showed we needed to create a memory cache for the legacy version. I don't imagine K2 will be any different.

jost-s (approved):
Cool, that makes perfect sense.

PR #20, opened 2024-11-25
Time slice

ThetaSinner:
Incomplete implementation of time slicing. 

The pieces that are missing:
- telling the time slice when the host has received new ops, so the slice hashes can be updated
- providing a time slice summary to gossip and comparing with one from another peer
- any logic for hash range partitioning. I see the time slicing being done per hash range, so this would need to be filtering by hash range when querying the host.


crates/memory/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
```
ThetaSinner, line 1: Memory module for testing and whatever else it's useful for

crates/api/src/timestamp.rs
```diff
@@ -46,6 +46,36 @@ impl std::ops::Add<std::time::Duration> for Timestamp {
     }
 }
 
+impl std::ops::AddAssign<std::time::Duration> for Timestamp {
```
ThetaSinner, line 49: Already have add, useful to be able to write "+="

crates/api/src/op_store.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+impl PartialOrd for MetaOp {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// The API that a kitsune2 host must implement to provide data persistence for kitsune2.
+pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
```
ThetaSinner, line 44: This is definitely not finished. We need to be able to partition by hash range too. Not sure what granularity we want the "space" at. Whether we create one op store per space or it needs to be passed through the API. I'm thinking the former but I'm not planning on dealing with those things as part of this change.

crates/api/src/op_store.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// The id of the op.
+    pub op_id: OpId,
+
+    /// The creation timestamp of the op.
+    ///
+    /// This must be the same for everyone who sees this op.
+    ///
+    /// The host must reject the op if the timestamp does not agree with any timestamps inside the
+    /// op data.
+    pub timestamp: Timestamp,
```
ThetaSinner, line 22: Want to expose this outside the op data because we want a concept of "when an op was created" but it's duplicated inside the op data. Suggestions? 

One option would be to make this field test only and we don't know when and op is received what it's timestamp is. Then have another op type for gossip to use where the timestamp is provided.

crates/api/src/lib.rs
```diff
@@ -35,6 +35,15 @@ pub(crate) mod serde_bytes_base64 {
...
 }
 
+/// A single struct to hold all the API types.
+///
+/// This is for convenience when passing around the API.
+#[derive(Clone)]
+pub struct HostApi {
+    /// The op store.
+    pub op_store: DynOpStore,
+}
```
neonphog, line 45: I don't think we have one unified host api in the new kitsune..

I've started [a builder pattern](https://github.com/holochain/kitsune2/pull/19/files#diff-31277162337faccf6affc0143f7bf04374c80b2dae71efef1a00f231b000317bR16) in the peer store PR... but that's for the factories.

The `tx` module instance will be instantiated at the top-level and passed into the individual spaces and most other modules as they are constructed. The spaces themselves will create the op store and peer store instances. Maybe we should just pass the needed instances into the factory functions and let the module implementations figure out how to store them? Or we can provide a `SpaceModuleInstances` struct or something?

crates/api/src/op_store.rs
neonphog: Not sure I understand the "partition by hash range" comment. The pagination for a catch-up sync would just be by timestamp cursor, and the only other query mode we need is for an arc-range, right?

I agree we should have a different op store instance per space (though holochain may or may not choose to use the same database for all the instances).

crates/api/src/op_store.rs
neonphog: Since kitsune can't parse the op content anyways... we would need to either:
- always hand raw ops received directly to the host, and get back out the MetaOp as part of querying the host with the timestamp included
- or, have the host provide some kind of parsing function when initializing kitsune that could extract the timestamp

I feel like we want the timestamp and any other data we need to be explicitly available to kitsune because that will simplify our implementation, but you should tell me what you think here.

crates/memory/Cargo.toml
neonphog: In my [peer store wip](https://github.com/holochain/kitsune2/pull/19/files#diff-72a2ccbabb7487b60b153d720970a3bf4ce28548158ec52e4ac35a4a93b00e8dR1) I put the memory store in the `kitsune2` crate itself. I feel like these will be small enough an useful enough for testing, that maybe we don't need separate crates for them. Also the rust compiler is good at dead code elimination : )

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+use kitsune2_api::{DynOpStore, K2Error, K2Result, OpId, Timestamp};
+use std::time::Duration;
+
+#[derive(Debug)]
+#[cfg_attr(test, derive(Clone, PartialEq))]
+pub struct PartitionedTime {
+    /// The timestamp at which the time partitioning starts.
+    ///
+    /// This must be a constant among peers who need to compute the same time slices.
+    origin_timestamp: Timestamp,
```
neonphog, line 59: My understanding is that the origin time largely only affects the alignment, and that the sizes of the partitions are fully defined by when "now" is. In which case it shouldn't matter how far back in time the origin is.

If that is true, could always use the unix epoch, and eliminate this aspect of complexity?

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+        // The end timestamp of the last full slice
+        let full_slice_end_timestamp = pt.full_slice_end_timestamp();
+
+        // Given the time reserved by full slices, how much time is left to partition into smaller slices
+        let recent_time = (Timestamp::now() - full_slice_end_timestamp).map_err(|_| {
+            K2Error::other("Failed to calculate recent time, either the clock is is wrong or this is a bug")
+        })?;
+
+        if pt.full_slices > 0 && recent_time < pt.min_recent_time {
+            return Err(K2Error::other("Not enough recent time reserved, other the clock is is wrong or this is a bug"));
```
neonphog, line 145:
```suggestion
            return Err(K2Error::other("Not enough recent time reserved, either the clock is is wrong or this is a bug"));
```

neonphog:
Gave a cursory look over this and responded to some of your own comments. But my brain is too tired to dive in deeply tonight. I'd like to say I'll dig back into this tomorrow morning, but we've got meetings.... so after lunch?

crates/dht/src/time.rs
jost-s: The origin serves as a starting point for the first time slice. Without it, how do you know when the first slice begins? You'd have several empty time slices until the first timestamp of an op. It'd be consistent across peers, but with these empty slices. That's my understanding.

neonphog: I'm not convinced we need to know when the first slice begins. When a gossip round starts we might need to gossip the earliest entry we have for the overlapping arcs, though.

jost-s: I understand that the protocol would be
- host initiates gossip
- remote returns list of time slices with super hashes
- host compares to its own list of time slices and super hashes (first number of time slices, if it differs, add list of ops to fetch queue, if it is equal compare hashes)
- if host or remote does not have list of time slices and super hashes in memory yet, compute them
To compute them, you need to start somewhere, and this somewhere must match on host and remote to arrive at the same slicing. If you start at unix epoch, you'd have a load of time slices at the beginning that contain no ops, because they are before the beginning of the app.

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+    ///
+    /// The size of a time slice is 2^factor * [UNIT_TIME].
+    /// Partial slices are created for "recent time" and are created in decreasing sizes from
+    /// 2^factor * [UNIT_TIME] down to 2^0 * [UNIT_TIME].
+    factor: u8,
+    /// The number of full time slices that have been stored.
+    ///
+    /// These full slices are always of size 2^factor * [UNIT_TIME].
+    full_slices: u64,
+    /// The partial slices, wish hashes stored in memory.
```
neonphog, line 70:
```suggestion
    /// The partial slices, with hashes stored in memory.
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+#[cfg_attr(test, derive(Clone, PartialEq))]
+pub struct PartialSlice {
+    /// The start timestamp of the time slice.
+    ///
+    /// This timestamp is included in the time slice.
+    start: Timestamp,
+    /// Size is used in the formula 2^size * [UNIT_TIME].
+    ///
+    /// That means a 0 size translates to [UNIT_TIME], and a 1 size translates to 2*[UNIT_TIME].
+    size: u8,
```
neonphog, line 99: I think I'm actually fine with this as-is, but consider if it would be clearer to test and reason about if you split the concerns of the time slicing and the hash maintenance:

```rust
pub struct TimeSliceLayout {
    pub full_slice_count: u64,
    pub partial_slice_list: Vec<(Timestamp, u8)>,
}

pub fn calc_current_time_slices(origin_timestamp: Timestamp, now: Timestamp) -> TimeSliceLayout {
    // ...
}

impl PartitionedTime {
    pub async fn update(
        &mut self,
        store: DynOpStore,
        layout: TimeSliceLayout,
    ) -> K2Result<()> {
        // ...
    }
}
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+
+/// Combine a series of op hashes into a single hash.
+///
+/// Requires that the op hashes are already ordered.
+/// If the input is empty, then the output is an empty byte array.
+fn combine_op_hashes(hashes: Vec<OpId>) -> bytes::Bytes {
+    hashes
+        .into_iter()
+        .map(|x| x.0 .0)
+        .reduce(|a, b| a.iter().zip(b.iter()).map(|(a, b)| a ^ b).collect())
```
neonphog, line 363: nitpick, but I'm not sure functional/reduce is right here... it's going to keep re-allocating memory. It would be more efficient to have a mut [BytesMut](https://docs.rs/bytes/1.8.0/bytes/struct.BytesMut.html) and freeze it once at the end into a Bytes.

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+    /// The store is needed to check how many time slices were created last time this
+    /// [PartitionedTime] was updated, if any.
+    ///
+    /// The method will then update the state of the [PartitionedTime] to the current time.
+    /// It does this by checking that the construction of this [PartitionedTime] is consistent
+    /// and then calling [PartitionedTime::update].
+    ///
+    /// The resulting [PartitionedTime] will be consistent with the store and the current time.
+    /// It should be updated again after [PartitionedTime::next_update_at].
+    pub async fn from_store(
```
neonphog, line 127:
```suggestion
    /// It should be updated again after [PartitionedTime::next_update_at].
    ///
    /// # Panics
    ///
    /// This function will panic if factor >= 54.
    pub async fn from_store(
```

crates/memory/Cargo.toml
ThetaSinner: The in memory store you've created really is production ready and matches how the store should work in production. This one really isn't a production implementation because without persistence, why go to all this work to sync ops between peers. 

My thinking with this is that I need the in-memory implementations is crates that shouldn't depend on `kitsune2` because `kitsune2` needs to depend on them. So I think this does need to stay outside `kitsune2`

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+    /// state with full and partial slices for the current time.
+    fn new(origin_timestamp: Timestamp, factor: u8) -> Self {
+        Self {
+            origin_timestamp,
+            factor,
+            full_slices: 0,
+            partial_slices: Vec::new(),
+            // This only changes based on the factor, which can't change at runtime,
+            // so compute it on construction
+            min_recent_time: residual_duration_for_factor(factor - 1),
```
neonphog, line 279: why -1 here?

crates/memory/Cargo.toml
neonphog: That makes sense. And it's not too hard to re-arrange if we want to make a different choice in the future :+1:

crates/api/src/op_store.rs
ThetaSinner: I think it should be by timestamps cursor, per hash range. So initially as we're doing full or empty arc, it's only full arc where we fetch ops and so yes it's just a time cursor. When we have the hash space partitioned, I'm thinking at the moment that hash partitioning lives at the top level and then we slice time within that hash partition.

So it looks like concentric circles, cut up like a pizza. 

For that to work, looking up data by time slice is always "what is the next block of ops from time 203942380492834, limited to 5Mb, in the hash range 0x...0x"

crates/api/src/op_store.rs
ThetaSinner: I like the first version. We shouldn't be doing anything with ops that haven't been persisted on the host. So whether we wait for persistence or integration, I think we need to hand ops to the host and have the host tell us when it's "processed" them. Whatever that means to a given host.

crates/api/src/lib.rs
ThetaSinner: Gotcha, yeah this was something I started for gossip that might not even be used on this PR. I'll remove it

crates/api/src/op_store.rs
neonphog: Okay, by hash range, you actually mean arc range, or hash location range. (just to distinguish from holochain's hash-bounded-queries).

crates/dht/src/time.rs
ThetaSinner: The time slices are of fixed size for full slices. So if you have more time, like back to the Unix epoch, then you have to store `[0, 0, ...., 0]` hashes for a bunch of time slices that didn't have a chance of data being in them. My thinking is that when the factor is appropriately chosen, storing a hash for a time range is an acceptable amount of metadata. 

I am very much relying on this per-app constant to reduce the amount of work that needs to be done.

I have a slightly different way I want to go about negotiating changed time slices but that is the right idea yes. You want to both be able to compute the same time slices without prior knowledge of each other's state. Because your earliest op could change, especially early on in a network's life, you'd have to agree that before you could send your summary hashes. That adds an extra round trip to gossip.

crates/memory/Cargo.toml
ThetaSinner: Agreed, we should take a good look at the crate layout before we publish to crates.io for the first time :)

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+                assert!(
+                    pt.partial_slices[i - 2].size
+                        != pt.partial_slices[i - 1].size
+                        || pt.partial_slices[i - 1].size != slice.size,
+                    "no more than two slices of the same size are permitted"
+                );
+            }
+
+            // Check that the slices are
+            assert_eq!(start_at, slice.start);
```
neonphog, line 1027:
```suggestion
            // Check that the slices start at the correct time.
            assert_eq!(start_at, slice.start);
```

crates/dht/src/time.rs
neonphog: I feel like it's possible to do this differently, but I don't want to spend any more time on it. Current holochain already uses an origin time, so it will be easier to integrate if it works this way anyways.

crates/api/src/op_store.rs
```diff
@@ -0,0 +1,86 @@
...
+        start: Timestamp,
+        end: Timestamp,
+    ) -> BoxFuture<'_, K2Result<Vec<OpId>>>;
+
+    /// Store the combined hash of a time slice.
+    fn store_slice_hash(
+        &self,
+        slice_id: u64,
+        slice_hash: Vec<u8>,
+    ) -> BoxFuture<'_, K2Result<()>>;
```
neonphog, line 69: How do you imagine this changing to integrate sharding? Will this take something like a BasicArc used just as another index key to identify the pie slice? (And similarly be required in the `retrieve_slice_hash()` fn?

neonphog:
Looks awesome! Some questions/comments, but I'm pretty happy with this : )

crates/dht/src/time.rs
ThetaSinner: I can't change this actually, because it's also used literally in the calculation code, so I'd end up adding an equally strange `+ 1`. So I'll add some information into the code comment about why that's being done

crates/dht/src/time.rs
ThetaSinner: Yes. There is at least one way. We could just compress runs of 0 hashes.

I was thinking that this might be useful anyway. If your time slices are something like 48 hours and everyone is offline for a week, then you could get some value from compressing those `[0, 0, ..., 0]`.

Then this could just compress everything from the unix timestamp up to whenever we first see a non-empty time slice. That would be simple enough to do I think. There is the possibility that we insert into a compressed section and have to expand it but I can live with that.

It's a touch of extra complexity, but I like that it always applies so that it eliminates the `origin_time` field and solves a general problem

ThetaSinner: This would affect the storage on the host and it would also mean we'd never send lots of 0 hashes over the network, we'd just send one "big" one at the front of the list. Which sounds like it's a constant overhead, but it isn't really if you install an app with a network seed, 6 months after it was packaged. That's just different bits in the "length" of the first block

crates/dht/src/time.rs
ThetaSinner: Trying to picture what this would look like. I think the update would need to return a diff or something like that. That's simple enough for the full slices but it's tricky for the partial slices where the hashes are kept in memory and we'd have to keep both old and new around to be able to figure out which changed. 

Definitely possible, and although I was thinking about these being one concern, this could be split. Do you think it's worthwhile?

crates/api/src/op_store.rs
ThetaSinner: Exactly that yes

crates/dht/src/time.rs
jost-s: You don't need to send the empty slices' hashes at all if the algorithm is deterministic on both sides, do you? You'd need to store the timestamp of the first actual time slice somehow, if you don't want to calculate it every time anew. So you'd end up with an internal origin time, just not set by app devs. Is that right?

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+    /// The method will then update the state of the [PartitionedTime] to the current time.
+    /// It does this by checking that the construction of this [PartitionedTime] is consistent
+    /// and then calling [PartitionedTime::update].
+    ///
+    /// The resulting [PartitionedTime] will be consistent with the store and the current time.
+    /// It should be updated again after [PartitionedTime::next_update_at].
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if factor >= 54.
```
jost-s, line 130: Why does it panic?

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+        let mut pt = Self::new(origin_timestamp, factor);
+
+        pt.full_slices = store.slice_hash_count().await?;
+
+        // The end timestamp of the last full slice
+        let full_slice_end_timestamp = pt.full_slice_end_timestamp();
+
+        // Given the time reserved by full slices, how much time is left to partition into smaller slices
+        let recent_time = (Timestamp::now() - full_slice_end_timestamp).map_err(|_| {
+            K2Error::other("Failed to calculate recent time, either the clock is is wrong or this is a bug")
```
jost-s, line 145:
```suggestion
            K2Error::other("Failed to calculate recent time, either the clock is wrong or this is a bug")
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+        })?;
+
+        if pt.full_slices > 0 && recent_time < pt.min_recent_time {
+            return Err(K2Error::other("Not enough recent time reserved, either the clock is is wrong or this is a bug"));
```
jost-s, line 149:
```suggestion
            return Err(K2Error::other("Not enough recent time reserved, either the clock is wrong or this is a bug"));
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+    /// It does this by checking that the construction of this [PartitionedTime] is consistent
+    /// and then calling [PartitionedTime::update].
+    ///
+    /// The resulting [PartitionedTime] will be consistent with the store and the current time.
+    /// It should be updated again after [PartitionedTime::next_update_at].
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if factor >= 54.
+    pub async fn from_store(
```
jost-s, line 131: As it's fallible
```suggestion
    pub async fn try_from_store(
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+                / ((1u64 << self.factor) * UNIT_TIME.as_secs())
+        } else {
+            0
+        };
+
+        for i in 0..new_full_slices_count {
+            // Store the hash of the full slice
+            let op_hashes = store
+                .retrieve_op_hashes_in_time_slice(
+                    full_slices_end_timestamp,
```
jost-s, line 204: It seems that this need to be updated in the loop. Otherwise it's always retrieving ops for the same time slice.

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+    /// Layout the partial slices for the given time range.
+    ///
+    /// Tries to allocate as many large slices as possible to fill the space.
+    fn layout_partials(
+        &self,
+        current_time: Timestamp,
+        mut start_at: Timestamp,
+    ) -> K2Result<Vec<(Timestamp, u8)>> {
+        let mut recent_time = (current_time - start_at).map_err(|_| {
+            K2Error::other("Failed to calculate recent time for partials, either the clock is is wrong or this is a bug")
```
jost-s, line 310:
```suggestion
            K2Error::other("Failed to calculate recent time for partials, either the clock is wrong or this is a bug")
```

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+/// Computes what duration is required for a series of time slices.
+///
+/// The duration is computed as the sum of the powers of two from 0 to the factor, multiplied by
+/// the [UNIT_TIME].
+///
+/// Panics if the factor is greater than 53.
+/// Note that the maximum factor changes if the [UNIT_TIME] is changed.
+fn residual_duration_for_factor(factor: u8) -> Duration {
+    if factor >= 54 {
+        panic!("Factor must be less than 54");
```
jost-s, line 350: Ah, that's why. Maybe an explanation in the docs.

jost-s (commented):
Gonna review the tests in a bit. The code of the functions is super readable, awesome work!

crates/dht/src/time.rs
neonphog: Hehe, cool, that works for me.

crates/dht/src/time.rs
neonphog: That's true, I might just be shifting the complexity around with this.

Although to clarify my suggestion, the layout function wouldn't have the hash in it, just like my example code, so I feel like you could iterate the new layout list and pluck the old one out by start time if it exists, doing the existing comparison if it exists and either re-hashing or not.

Part of the time it took me here to achieve enough understanding to review was in figuring out the interplay between the figuring out where the time slices are bounded, and then how that is used to generate the hashes. I *feel* like I might be able to get up to speed faster if I could set aside the time slice logic and focus only on the hash generation.

But I don't want to make things worse, so I'll rely on your judgment from here! : )

crates/dht/src/time.rs
neonphog: Yep, perfect!

crates/dht/src/time.rs
neonphog: I like the combining zeros idea. It's pretty easy to understand, especially since we can give ipv6 string encoding as an example. If you think it's worth considering, would it make the most sense to do that as a separate effort after this PR?

jost-s (commented):
I'll continue tomorrow.

crates/api/src/op_store.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// This must be the same for everyone who sees this op.
+    ///
+    /// The host must reject the op if the timestamp does not agree with any timestamps inside the
+    /// op data.
+    pub timestamp: Timestamp,
+
+    /// The actual op data.
+    pub op_data: Vec<u8>,
+
+    /// Crdt-style add-only opaque implementor-use flags.
```
mattyg, line 27: Should we give some more context about this? I don't understand it from just this comment

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
+//! Partition of time into slices.
+//!
+//! Provides an encapsulation for calculating time slices that is consistent when computed
+//! by different nodes. This is used to group Kitsune2 op data into time slices for combined hashes.
+//! Giving multiple nodes the same time slice boundaries allows them to compute the same combined
+//! hashes, and therefore effectively communicate which of their time slices match and which do not.
+//!
+//! A time slice is defined to be an interval of time that is closed (inclusive) at the start and
+//! exclusive at the end. That is, the time slice includes the start timestamp but not the end.
```
mattyg, line 9: Had to read this a few times but I'm bad with words. I found this clearer:

/// A time slice represents an interval [start, end] where:
/// - `start`: Included in the slice (inclusive bound)
/// - `end`: Not included in the slice (exclusive bound)

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
+//! Partition of time into slices.
```
mattyg, line 1: I like this approach a lot

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+                }
+            };
+
+            self.partial_slices.push(PartialSlice { start, size, hash })
+        }
+
+        // There will be a small amount of time left over, which we can't partition into smaller
+        // slices. Some amount less than [UNIT_TIME] will be left over.
+        // Once that has elapsed up to the next [UNIT_TIME] boundary, we can update again.
+        if let Some(last_partial) = self.partial_slices.last() {
```
mattyg, line 261: Does this mean that the ops in this range don't get gossiped until they're included in a time slice? Or does it mean that we have to send the full list of op hashes in this range to compare with our gossip peer?

jost-s (commented):
Done with the first review. I appreciate the fantastic test suite!

crates/api/src/op_store.rs
ThetaSinner: Exactly that, though based on this we'd need to make some changes to the types used in that

crates/dht/src/time.rs
ThetaSinner: Sounds like good support for this, I vote for a follow up change because it'll be easier to review as a change. Yes that means introducing origin_time just to remove it but I think that's okay

Cargo.toml
```diff
@@ -26,6 +32,8 @@ serde = { version = "1.0.215", features = ["derive"] }
 # kitsune2 agent info is serialized as json to improve debugability of
 # bootstrapping. So, we need a json library.
 serde_json = "1.0.132"
+# Simplify writing async code
+futures = "0.3"
```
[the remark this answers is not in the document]
ThetaSinner: It's the same either way. If you put "0.3.1" or "0.3" you'll get "0.3.2" and a `cargo update` will give you "0.3.4" when that's been released. It's really about whether you want to record what your absolute minimum version is. My opinion is that Cargo doesn't provide good tools for checking whether you need 0.3.0 or 0.3.1 or 0.3.2, so I'd rather just specify the minor version... but either way is fine

Should we be more consistent about this?

crates/api/src/op_store.rs
```diff
@@ -0,0 +1,86 @@
...
+        Some(self.cmp(other))
+    }
+}
+
+/// The API that a kitsune2 host must implement to provide data persistence for kitsune2.
+pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
+    /// Process incoming ops.
+    ///
+    /// Pass the incoming ops to the host for processing. The host is expected to store the ops
+    /// if it is able to process them.
```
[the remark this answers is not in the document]
ThetaSinner: This would be coming from the network, so we can't trust the ops being ordered by timestamp is correct. Plus it's not so important because the ops have to be merged into the host's list which will involve inserting them into an ordered index of some kind whether they are sorted or not. It would be slightly more efficient if they were sorted.. but not very. Oh and also, we can't have the timestamps here, that needs to be part of the op data and not the metaop, so at this point we wouldn't even know how to sort them in Kitsune.

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1034 @@
...
+//!      how to partition the remaining time into partial slices.
+//!    - Store the combined hash of any new full slices in the Kitsune2 op store.
+//!    - Store the combined hash of any new partial slices in memory.
+//!    - Update the `next_update_at` field to the next time an update is required.
+//!
+//! As an example, consider a factor of 9. This means that full slices are 2^9 = 512 times the [UNIT_TIME].
+//! With a unit time of 15 minutes, that means full slices are 512 * 15 minutes = 7680 minutes = 128 hours.
+//! So every 5 days, a new full slice is created.
+//! The partial slices reserve at least 2^8 + 2^7 ... 2^0 = 519 times the [UNIT_TIME], so 519 * 15 minutes
+//! = 7785 minutes = 129.75 hours. That gives roughly 5 days of recent time.
```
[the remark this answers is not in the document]
ThetaSinner: Oh dear, that's some bad maths right there :D Thank you for catching that

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+    /// The factor used to determine the size of the time slices.
+    ///
+    /// The size of a time slice is 2^factor * [UNIT_TIME].
+    /// Partial slices are created for "recent time" and are created in decreasing sizes from
+    /// 2^factor * [UNIT_TIME] down to 2^0 * [UNIT_TIME].
+    factor: u8,
+    /// The number of full time slices that have been stored.
+    ///
+    /// These full slices are always of size 2^factor * [UNIT_TIME].
+    full_slices: u64,
```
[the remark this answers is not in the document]
ThetaSinner: It is, the full slices have their hashes stored on the host. They will change much less frequently and ideally we wouldn't permit them to change at all after some amount of time has passed

crates/dht/src/time.rs
ThetaSinner: Integer overflow, the numbers just get too large. 

This is a good point though, because if we let this be settable by apps then maybe we need to have this return an error and reject the app install rather than crashing the whole conductor... Yeah I should change this I think

crates/dht/src/time.rs
ThetaSinner: Oh dear, that seems like something one could write a test for! That's certainly a bug. Let me get a test and a fix in for that :)

ThetaSinner: Added a new test `update_allocate_multiple_new_complete` that was failing and now passes. Also extracted a pre-calculated full slice size because that was being done in several places. That reminded me that I'd meant to do that and fix this issue... Glad it got caught at review :)

[fffee075](https://github.com/holochain/kitsune2/pull/20/commits/fffee0753fb996628fa6e11c237dbd645392b69a)

crates/dht/src/time.rs
ThetaSinner: Right, I have a better idea of what I should change to answer this question now. Yes I think this is valuable to do now because `update` is a rather large function and it is doing two things. I would like to keep all the logic contained inside this type but I think it can be split into two functions that do different things. I'll have a shot at this now :)

ThetaSinner: I think this is better and easier to digest, what do you think? Does it go far enough?

[e9729db7](https://github.com/holochain/kitsune2/pull/20/commits/e9729db7c1063a390357a38e190483e08746ff47)

crates/dht/src/time.rs
ThetaSinner: Agreed, added a note about selecting a factor

crates/memory/src/op_store.rs
```diff
@@ -0,0 +1,85 @@
...
+        &self,
+        start: Timestamp,
+        end: Timestamp,
+    ) -> BoxFuture<'_, K2Result<Vec<OpId>>> {
+        async move {
+            let self_lock = self.read().await;
+            Ok(self_lock
+                .op_list
+                .iter()
+                .filter(|op| op.timestamp >= start && op.timestamp <= end)
```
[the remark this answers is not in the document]
ThetaSinner: Yes, correct. Found and fixed this because of the issue you spotted above with multiple full slices being allocated in a single update

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+        // of each smaller slice size
+        assert_eq!((8 + 4 + 2 + 1) * UNIT_TIME, pt.min_recent_time);
+    }
+
+    #[tokio::test]
+    async fn from_store() {
+        let origin_timestamp = Timestamp::now();
+        let factor = 4;
+        let store = Arc::new(Kitsune2MemoryOpStore::default());
+        let pt = PartitionedTime::from_store(origin_timestamp, factor, store)
```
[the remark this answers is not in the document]
ThetaSinner: I was thinking that this is an app level setting but that it affects the space hash. I.e. changing the factor is not something that can be done within a network and attempting to do so would require a new app install. That also means we couldn't change the default without it being a breaking change.

What do you think? Is it okay to prevent this at a higher level than here?
I'm not really sure how to fit it into the store and it feels redundant to store it with every slice hash because it should be a constant.

crates/api/src/op_store.rs
ThetaSinner: Good question, this was copied from the original kitsune2-api spike. I think we actually agreed to make this a u32 and expose that as an enum to that humans don't have to look at it.

I think I should just remove this from the current PR and we'll add it in when we need op flags.

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+    }
+
+    #[tokio::test]
+    async fn all_single_partial_slices() {
+        let factor = 7;
+        let origin_timestamp = (Timestamp::now()
+            - Duration::from_secs(
+                // Enough time for all the partial slices
+                PartitionedTime::new(Timestamp::now(), factor)
+                    .min_recent_time
```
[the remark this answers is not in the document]
ThetaSinner: I'll create a helper function, I want this to stay in sync with the implementation as well as possible

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+        )
+        .await
+        .unwrap();
+
+        // Capture the initially created state
+        let pt_original = pt.clone();
+
+        // Update repeatedly, the state should not change
+        for i in 1..10 {
+            let call_at = now + Duration::from_secs(i);
```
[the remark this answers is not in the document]
ThetaSinner: Yep, sensible

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+        let full_slice_hash =
+            store.retrieve_slice_hash(0).await.unwrap().unwrap();
+        // The hashes should be combined using XOR
+        assert_eq!(vec![7 ^ 23; 32], full_slice_hash);
+
+        validate_partial_slices(&pt);
+    }
+
+    #[tokio::test]
+    async fn compute_hashes_for_full_slices_and_partials() {
```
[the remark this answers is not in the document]
ThetaSinner: Absolutely right, this got added as part of the comment above about updating the end time of the full slices in a loop

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+            .unwrap();
+
+        assert_eq!(2, pt.full_slices);
+        assert_eq!(factor as usize, pt.partial_slices.len());
+
+        validate_partial_slices(&pt);
+    }
+
+    #[tokio::test]
+    async fn missing_full_slices_with_all_partials() {
```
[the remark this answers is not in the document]
ThetaSinner: Yep, good point, removed

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1051 @@
...
+            2u32.pow(factor as u32 - 1) as u64 * UNIT_TIME.as_secs()
+                + 1,
+            ))
+        .unwrap();
+        let store = Arc::new(Kitsune2MemoryOpStore::default());
+        let pt = PartitionedTime::from_store(origin_timestamp, factor, store)
+            .await
+            .unwrap();
+
+        assert_eq!(factor as usize + 1, pt.partial_slices.len());
```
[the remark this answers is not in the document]
ThetaSinner: Yep, a good addition, done

crates/dht/src/time.rs
ThetaSinner: The notation is good, I'd like to keep the initial definition in terms of open/closed. Which I'd actually failed to do in the first place. Then I've added an alternative description that is just this.

Does that read better?

crates/dht/src/time.rs
ThetaSinner: I am undecided about this. That is what kitsune1 does with its final time slice. So it would certainly be a valid strategy to just say, at the point that we start a gossip round, compute the hash for the time slice between the end of the last partial and the current timestamp.

Alternatively, we don't really need to communicate that last hash. It's a constantly moving target in most cases and we're already just going to eagerly push "what's new since we last spoke". I'd be fine with saying that everything more recent than UNIT_TIME doesn't get hash checked. If we happen to miss something in incremental updates (ideally rare!) then we can find out when those ops make it into a slice.
If we crash and lose some ops that way, then gossip has to try to figure out when to start syncing from anyway, and UNIT_TIME is really a lower bound for that because it would only take one op being created while we were offline to upset that "most recent" hash. So we're going to need to resync at least that window of hashes anyway to get caught up.

What do you think?

crates/api/src/op_store.rs
ThetaSinner: https://github.com/holochain/kitsune2/issues/31

crates/dht/src/time.rs
ThetaSinner: https://github.com/holochain/kitsune2/issues/32

crates/dht/src/time.rs
neonphog: Yes, that is easier to follow, thanks!

neonphog (approved):
:tada:

crates/dht/src/time.rs
```diff
@@ -0,0 +1,1123 @@
+//! Partition of time into slices.
+//!
+//! Provides an encapsulation for calculating time slices that is consistent when computed
+//! by different nodes. This is used to group Kitsune2 op data into time slices for combined hashes.
+//! Giving multiple nodes the same time slice boundaries allows them to compute the same combined
+//! hashes, and therefore effectively communicate which of their time slices match and which do not.
+//!
+//! A time slice is defined to be an interval of time that is closed at the start and
+//! open at the end. That is, an interval [start, end)` where:
```
mattyg, line 9:
```suggestion
//! open at the end. That is, an interval [start, end] where:
```

not sure if its a typo `or` notation

ThetaSinner: It's intentional [notation](https://www.statisticshowto.com/calculus-definitions/closed-interval-open-interval/) but I did mean to match the backticks, fixed

Cargo.toml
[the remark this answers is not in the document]
jost-s: Might as well use minor versions, but I don't feel strongly about this.

crates/dht/src/time.rs
mattyg: Either of those approaches seem fine  -- I wasn't sure if the idea was to intentionally *not* gossip ops in the leftover time until they get included in a time slice.

crates/dht/src/time.rs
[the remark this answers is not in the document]
jost-s: I understand, but since we're comparing them every time the number of slices matches the remote's number of slices, it would be more resource-efficient to keep them in memory than restoring them from a database each time. They shouldn't take up much space in memory.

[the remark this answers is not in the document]
ThetaSinner: I'm thinking that in the majority of cases, we don't need to retrieve them. If we have a full slice time of something like 5 days then over time we end up with quite a few hashes to keep in memory. Not an unreasonable number but when it's required across a hash partition too, that's a fair bit. 

In the majority of gossip rounds, where both nodes have been online since the last gossip round, all we really need to check is "do our all our previous 5 day slices, combined, still agree". We can use a count of full slices to understand which side needs to recalculate if they are at a different point (i.e. one has just updated their ParititionedTime and the other hasn't).

So we probably end up storing the current and maybe 1 previous combined hash of all full slices here. Yes, we will end up adding something here, but I intentionally left out storing them all. It's only when we find a historical mismatch that we need to grab more historical slices. But even then, we only need to fetch from where the remote is telling us they match up to.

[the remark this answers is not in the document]
ThetaSinner: I hope this makes some sense, it's a little rambly! :)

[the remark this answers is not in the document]
jost-s: Gotcha, yes. I was assuming that we don't store a timestamp when two peers last spoke. If they do, even just in memory, then you only have to restore those full slice hashes once and combine them to a... combined hash. That combined hash may be added, understood. That's sound.

crates/dht/src/time.rs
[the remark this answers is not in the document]
jost-s: Sure thing, yes. I see your reasoning for this.

Issue #22, opened 2024-11-27
[k2] space module


neonphog:
The space module is mapped in the top-level (kitsune2) module and allows joining and leaving of local agents. It creates sub-modules such as the peer store and gossip modules.

PR #33, opened 2024-11-27
Space Module Stubs

neonphog:
This does *NOT* resolve https://github.com/holochain/kitsune2/issues/22 -- this is just a set of stubs that set up what a space might look like.

We've got a chicken/egg problem with getting some of this code in here, and need some concrete bits to be in place to make it easier to do the things around it, like the top-level module or the tx module.

### DIFF
--- a/crates/api/src/space.rs
+++ b/crates/api/src/space.rs
@@ -3,6 +3,27 @@
 use crate::*;
 use std::sync::Arc;
 
+/// Handler for events coming out of Kitsune2 such as messages from peers.
+pub trait SpaceHandler: 'static + Send + Sync + std::fmt::Debug {
+    /// We have received an incomming message from a remote peer.
+    ///
+    /// If this callback handler returns an `Err` response, the connection
+    /// will be closed immediately.
+    //
+    // Note: this is the minimal low-level messaging unit. We can decide
+    //       later if we want to handle request/response tracking in
+    //       kitsune itself as a convenience or if users of this lib
+    //       should have to implement that if they want it.
+    fn incoming_message(
+        &self,
+        peer: AgentId,
+        data: bytes::Bytes,
+    ) -> K2Result<()>;
+}
+
+/// Trait-object [SpaceHandler].
+pub type DynSpaceHandler = Arc<dyn SpaceHandler>;
+
 /// Represents a unique dht space within which to communicate with peers.
 pub trait Space: 'static + Send + Sync + std::fmt::Debug {
     /// Get a reference to the peer store being used by this space.
@@ -24,6 +45,19 @@ pub trait Space: 'static + Send + Sync + std::fmt::Debug {
     /// be generated and sent to the bootstrap server. A best effort will
     /// be made to publish this tomstone to peers in the space as well.
     fn local_agent_leave(&self, local_agent: id::AgentId) -> BoxFut<'_, ()>;
+
+    /// Send a message to a remote peer. The future returned from this
+    /// function will track the message all the way down to the low-level
+    /// network transport implementation. But once the data is handed off
+    /// there, this function will return Ok, and you will not know if
+    /// the remote has received it or not.
+    //
+    // Note: this is the minimal low-level messaging unit. We can decide
+    //       later if we want to handle request/response tracking in
+    //       kitsune itself as a convenience or if users of this lib
+    //       should have to implement that if they want it.
+    fn send_message(&self, peer: AgentId, data: bytes::Bytes)
+        -> BoxFut<'_, ()>;
 }
 
 /// Trait-object [Space].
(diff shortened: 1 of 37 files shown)


crates/api/src/space.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// will be closed immediately.
+    //
+    // Note: this is the minimal low-level messaging unit. We can decide
+    //       later if we want to handle request/response tracking in
+    //       kitsune itself as a convenience or if users of this lib
+    //       should have to implement that if they want it.
+    fn incoming_message(
+        &self,
+        peer: AgentId,
+        data: bytes::Bytes,
```
ThetaSinner, line 20: Should this not be typed? Like an enum for the messages Kitsune supports?

crates/kitsune2/src/factories/core_space.rs
ThetaSinner, line 1: This one I think we maybe want in the memory module. To write tests that need to interact with the host like gossip will need to, for registering ops and being told their timestamps, we'll want to get at this without depending on the `kitsune2` module... I think

ThetaSinner (commented):
This is useful. Fine with some pieces being placeholders because we do need more of these pieces in place to build around :)

mattyg (approved):

crates/api/src/space.rs
jost-s: I think that decoding the bytes and then matching against kitsune-own message types for further processing will be done in this function.

crates/api/src/space.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn local_agent_join(
+        &self,
+        local_agent: agent::DynLocalAgent,
+    ) -> BoxFut<'_, K2Result<()>>;
+
+    /// Indicate that an agent is no longer online, and should stop
+    /// receiving messages and exchanging dht information.
+    /// Before the agent is actually removed a tombstone agent info will
+    /// be generated and sent to the bootstrap server. A best effort will
+    /// be made to publish this tomstone to peers in the space as well.
```
jost-s, line 46:
```suggestion
    /// be made to publish this tombstone to peers in the space as well.
```

crates/api/src/space.rs
```diff
@@ -0,0 +1,82 @@
...
+pub trait Space: 'static + Send + Sync + std::fmt::Debug {
+    /// Get a reference to the peer store being used by this space.
+    /// This could allow you to inject peer info from some source other
+    /// than gossip or bootstrap, or to query the store directly if
+    /// you have a need to determine peers for direct messaging.
+    fn peer_store(&self) -> &peer_store::DynPeerStore;
+
+    /// Indicate that an agent is now online, and should begin receiving
+    /// messages and exchanging dht information.
+    fn local_agent_join(
```
jost-s, line 37: Why are these methods for local agents? Is it because you can online join a particular kitsune node if you're local?

jost-s (dismissed):

crates/api/src/space.rs
neonphog: The `data` here is the opaque payload of something like receive_remote_signal. It's for holochain_p2p to apply any deserialization beyond that including picking the encoding scheme (although we know that will be messagepack).

Separately we still need to decide if kitsune is going to internally do request/response matching as a convenience to holochain, or if we'd prefer to move that complexity up a level to keep kitsune simple. But, I figured we can defer that decision a bit until we know if kitsune's internal modules are going to need that support anyways.

crates/kitsune2/src/factories/core_space.rs
neonphog: @ThetaSinner - Okay, I think I'm understanding better now what you're wanting from this crate separation...

I was really hoping to have a common builder somewhere that would default to usable factories for testing that we could swap out the specific ones we were testing. But that indeed leads to circular dependencies if you are trying to test a module that is included in that list.

Given this understanding, I don't think "mem" is correct, both because some of the impl are real and because some of the stubs will be using other module apis and not really have a mem aspect. I propose:

- `kitsune2_core` crate, with following modules:
  - "Real" implementations that will be used in production
    - `top-level` (we need a better name for this)
    - `space`
    - `peer-store`
    - `bootstrap-client`
  - "Stub" implementations that will make testing possible. The "real" impls are in other crates and are pulled into the `kitsune2` crate for production
    - `op-store` - STUB MEM implementation
    - `gossip` - STUB impl has access to both peers so just immediately sends op data between them
    - `publish` - STUB impl directly sends to all other relevant peers
    - others (sharding, fetch, etc) TBD as we get to designing them may or may not be "real" per their complexity

crates/api/src/space.rs
neonphog: The distinguishing feature of a "Local" agent vs a generic "Agent" or "AgentInfo" is that a local agent is able to sign messages. So this function would be used in conductor when a cell is enabled, it provides an instance of a "LocalAgent" to kitsune that is able to sign agent infos and make that peer discoverable online.

crates/kitsune2/src/factories/core_space.rs
neonphog: If we're cool moving this way, here's the first step: https://github.com/holochain/kitsune2/pull/39

PR #35, opened 2024-11-28
Fetch queue

jost-s:
Draft with traits for fetch queue and fetch task. There's a basic implementation of the fetch queue and of the fetch task. Whether the fetch queue benefits now from another data structure depends on how the fetch task should operate. Should fetch attempts be triggered when ops are added to the queue? Or should there be an constant loop running? There may also be better strategies for the fetch tasks instead of choosing a random agent and start fetching all ops.

Other than that I was going down the road of including a mock tx to start sending fetch requests to agents, but it turned out to be quite an undertaking, so I decided to leave it at this as a basis for discussion. Perhaps I could create a stub for the `FetchResponseQueue` which then starts removing items from the fetch queue, and based on that advance the fetch task implementation/testing?


crates/fetch_queue/src/lib.rs
jost-s, line 1: Not a finalized description.

crates/fetch_queue/src/fetch_queue.rs
```diff
@@ -0,0 +1,131 @@
+use std::collections::HashSet;
+
+use kitsune2_api::{fetch::FetchQueueT, AgentId, OpId};
+use rand::seq::IteratorRandom;
+
+#[derive(Debug)]
+pub struct FetchQueue {
+    ops: Vec<OpId>,
```
ThetaSinner, line 8: These fields need to be grouped in some way. Not every agent is a valid source for every op. In a sharded network, that will certainly not be the case. Even unsharded, we can't assume that every agent have a given op unless they've told us they do because it might be a new agent who is catching up

ThetaSinner (changes requested):

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,28 @@
+//! Kitsune2 fetch queue types.
+
+use std::sync::Arc;
+
+use crate::{AgentId, OpId};
+
+/// Trait for implementing a fetch queue for managing ops
+/// to be fetched from other agents.
+pub trait FetchQueueT: 'static + Send + Sync {
```
neonphog, line 9:
```suggestion
pub trait FetchQueue: 'static + Send + Sync {
```

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// Trait-object fetch queue instance.
+pub type DynFetchQueue = Arc<dyn FetchQueueT>;
+
+/// Trait for implementing a fetch task that periodically attempts to
+/// send op fetch requests to another agent.
+pub trait FetchTaskT {
+    /// Spawn a new fetch task, backed by a fetch queue.
+    fn spawn(&self, fetch_queue: DynFetchQueue);
+}
```
neonphog, line 28: I don't think we need the complexity of a separate "task". Following the abstraction set out in the spike, we need a FetchQueueFactory from which the space module in *its* factory can construct a real `DynFetchQueue` instance. That instance internally may have tasks it may have data stores... as far as abstractions go, we don't care, that's up to the implementation.

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,28 @@
...
+/// to be fetched from other agents.
+pub trait FetchQueueT: 'static + Send + Sync {
+    /// Add op ids to be fetched to the queue.
+    fn add_ops(&mut self, op_list: Vec<OpId>, source: AgentId);
+
+    /// Get a list of op ids to be fetched.
+    fn get_ops_to_fetch(&self) -> Vec<OpId>;
+
+    /// Get a random agent that holds the ops to be fetched.
+    fn get_random_source(&self) -> Option<AgentId>;
```
neonphog, line 17: I think we should start with *only* `add_ops`. Gossip may eventually need some extra view into how overworked the fetch queue is, but we don't really know what that needs to look like yet, and also it may not need that. To start, I think it'd be fine if it redundantly added ops again, and that was just a no-op.

crates/fetch_queue/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "holo", "p2p", "dht", "networking"]
+categories = ["network-programming"]
+edition = "2021"
+
+[dependencies]
+kitsune2_api = { path = "../api", version = "0.0.1-alpha" }
+bytes = { workspace = true }
+rand = "0.8.5"
+tokio = { workspace = true, features = ["full"] }
```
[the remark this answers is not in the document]
neonphog: Yes, if this is starting to be used as a real runtime dependency, we need to figure out what this looks like.

Perhaps it is naive of me to say I'd like kitsune to be async runtime independent...

I *think*  from tokio or some other runtime, we only care about spawning tasks and async timeouts. Both of these we could abstract with a trait that the implementer could fill out if they want, and we could provide a "tokio" feature that would provide a default impl of said trait through tokio.

Everything else (channels, etc) we could get from the futures crate.

It is a given that this approach would be more overhead for us as developers. And currently tx5 internally uses tokio directly, so for this to be useful, we'd have to gut that as well.

Is the end result of this worth the effort to us as an org?

crates/fetch_queue/src/fetch_queue.rs
neonphog: Also, we need to accommodate the request for validation receipts flags thing. Maybe something like:

```rust
type Flags = u32;
struct Q(VecDeque<(OpId, HashSet<AgentId>, Flags)>);
```

The `VecDeque` gives it actual queue properties, but perhaps on fresh syncs will contain too many items to be efficient enough? We could bring in IndexMap to keep them in order:

```rust
type Flags = u32;
struct Q(IndexMap<OpId, (HashSet<AgentId>, Flags)>);
```

crates/fetch_queue/src/fetch_queue.rs
```diff
@@ -0,0 +1,131 @@
+use std::collections::HashSet;
+
+use kitsune2_api::{fetch::FetchQueueT, AgentId, OpId};
+use rand::seq::IteratorRandom;
+
+#[derive(Debug)]
+pub struct FetchQueue {
```
neonphog, line 7: Not pub, the only pub item should be the factory.

crates/fetch_queue/src/fetch_task.rs
```diff
@@ -0,0 +1,52 @@
+use std::time::Duration;
+
+use kitsune2_api::fetch::{DynFetchQueue, FetchTaskT};
+
+/// Configuration for [`FetchTask`].
+pub struct FetchTaskConfig {
```
neonphog, line 6: The [peer-store pr](https://github.com/holochain/kitsune2/pull/19/files#diff-72a2ccbabb7487b60b153d720970a3bf4ce28548158ec52e4ac35a4a93b00e8dR9-R16) has an example of a module config and how to work with it in the factory.

crates/fetch_queue/src/fetch_task.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// How long to pause after sending a fetch request, before the next attempt.
+    pub pause_between_runs: u64, // in ms
+    /// Maximum number of ops to request from a remote.
+    pub max_ops_to_request: usize,
```
neonphog, line 10: I think we need to come at this from a slightly different angle.

We don't know the size of any individual ops we will be requesting. We also don't want to send so many hashes that the hashes themselves are using a significant part of our bandwidth for the remote to only be able to respond with data to a handful of them and then we need to request them again anyways. We don't know if a particular remote will be responsive or not, and if the *are* responsive, we don't know if they might respond (no, I'm too busy right now).

I suggest, config wise, something like:

```rust
struct Config {
    /// How many parallel fetch requests for actual data we should have at once. Default: 2.
    parallel_request_count: u8,
    /// Max hash count to request from one peer at a time. Default: 16.
    max_hash_count: u8,
    /// Max byte count that we want to get in a single request. Default: 10MiB.
    /// This will be sent to the remote as part of the request. If the data of the requested ops
    /// is larger than this amount, they will send back as the ACK the count of hashes
    /// that they will be responding with that is under this byte count.
    max_byte_count: u32,
}
```

neonphog (changes requested):
@jost-s - I feel like my comments sum up to "do this entirely differently". I don't want to be a jerk, but I have some pretty strong opinions on how I'd like to see this next version of Kitsune developed. I totally appreciate the work you've done to get the PR up so we have a starting point for discussion on how to move forward.

Let me know if there is anything different we could have done (or still should do) to get everyone on the same page. Or maybe having concrete code to pick apart is the best way at the moment.

neonphog:
@jost-s @ThetaSinner - Per this discussion: https://github.com/holochain/kitsune2/pull/33#discussion_r1863811918

Do you think we should shift our goal a little on this to truly just getting in an MVP api in the api crate, then maybe or maybe not adding a true stub (with actual `todo!()`s)?

I still think the MVP fetch queue api may be *just* the `add_ops` function with the flags added. And the MVP fetch response queue may have NO api at all. It just will be responding to protocol messages from the fetch queue.

jost-s:
> @jost-s - I feel like my comments sum up to "do this entirely differently". I don't want to be a jerk, but I have some pretty strong opinions on how I'd like to see this next version of Kitsune developed. I totally appreciate the work you've done to get the PR up so we have a starting point for discussion on how to move forward.
> 
> Let me know if there is anything different we could have done (or still should do) to get everyone on the same page. Or maybe having concrete code to pick apart is the best way at the moment.

@neonphog That's perfectly fine with me, no offense taken whatsoever. I haven't gone down any specific road yet here and kept it basic, just for that reason, that there are lots of ways how to skin this cat, and knowing that the two of you probably have something specific in mind already. Also you and Gregory know kitsune way more than I do, so I'm happy to be guided by that.

And I'm proud of the quality of the code that has gone into this repo so far!

Yes, this PR is a starting point for discussion. It seems that you as well as @ThetaSinner have ideas about the working of the fetch queue. I think we are not necessarily all on the same page, so maybe a design session would be helpful?

crates/fetch_queue/src/fetch_task.rs
mattyg: I like this suggestion as opening up possibility of conductor-level bandwidth constraints

crates/fetch_queue/src/fetch_queue.rs
jost-s: @ThetaSinner I got hung up on your idea of not fetching ops immediately from the same source we got the op list from, and randomizing who we fetch from. In the process of writing the code, I also thought it's very likely that not all agents will have the same range of ops to respond with. That's actually what I meant in Mattermost when I asked what to do when agents don't have all ops, and if we wanted to treat that optimistically and assume, it'll be complete eventually.

Too much thinking to myself and too little context in my messages. I agree this needs to be per agent, and now we're on the same page regarding immediate fetches as well.

jost-s: @neonphog IndexMap is used in kitsune1 too, I like that.

crates/api/src/fetch.rs
jost-s: I understand now. There's just the fetch queue that manages state and fetching, and all consumers do is add op lists with agent to it. How will the fetch response queue or its owning struct remove ops that have been fetched?

crates/fetch_queue/src/fetch_task.rs
jost-s: When I wrote this, the peer store PR didn't exist yet and I was hesitant to introduce the factory. I see now how it's done.

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::{AgentId, OpId};
+
+/// Trait for implementing a fetch queue for managing ops
+/// to be fetched from other agents.
+pub trait FetchQueueT: 'static + Send + Sync {
+    /// Add op ids to be fetched to the queue.
+    fn add_ops(&mut self, op_list: Vec<OpId>, source: AgentId);
+
+    /// Get a list of op ids to be fetched.
+    fn get_ops_to_fetch(&self) -> Vec<OpId>;
```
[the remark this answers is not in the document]
jost-s: Makes sense with a different approach, yes.

crates/api/src/fetch.rs
jost-s: I realize that that is an optimization and depends on how the gossip protocol will work.

crates/fetch_queue/Cargo.toml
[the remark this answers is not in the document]
jost-s: Actually having tokio with full features in the dependencies instead of the dev-dependencies was a mistake I forgot to fix.

Regarding async runtime-agnosticism, I'm up for evaluating how much overhead this is to adhere to. In general I don't see a pull for it, as I don't even know how much potential adoption kitsune could have outside of Holochain.

crates/fetch_queue/src/fetch_queue.rs
jost-s: What do you think of indexing by agent id instead of op id?
```rust
struct Q(IndexMap<AgentId, (Vec<OpId>, Vec<Flags>)>);
```

A fetch request takes `max_hash_count` number of elements from the vector of the first of the map's values, until that vector is empty. Then the next agent id's op list is processed.

Otherwise with the op id as index, ops are added individually with the agent as source, and when making a fetch request, you'd pop the first op id and then collect `max_hash_count` of op ids that are also in the map for that source. Seems more complex.

@neonphog @ThetaSinner

crates/api/src/fetch.rs
neonphog: My intention with the tx module is that other modules (such as this FetchQueue module) can register sub-protocol handlers. So the fetch module would send op hashes to remotes, those remotes would send the op data back, and the fetch queue module would receive it and be able to remove the item from the queue.

Looking at it this way, perhaps the response queue actually needs to be managed by this module as well so we don't have to split the protocol between multiple modules.

crates/fetch_queue/src/fetch_queue.rs
neonphog: There are 2 concerns here:

- First, we want the ops roughly in the order in which they were requested. Because a fresh sync will be going by time older to newer. This feels like a pretty soft requirement, because, frankly, who cares. We eventually need to get all of them, do we even actually care about the order?
- Second, it makes sense to optimize a request to a particular peer... that is, send a request for all the ops we think that peer will be able to respond with. This seems like a harder requirement because there is real-world overhead to connection establishment and bandwidth usage.

The downside of doing it the way you suggest is the dis-correlation of OpIds. The op data is the real bandwidth bottleneck, so I suggest we do not want to be making parallel requests for the op data at the same time. And this suggested data structure will make it very difficult to prevent this.

jost-s: I thought we did care about the order, but this is a part that leaves room for a lot different approaches and speculations.

Got your second point, that is a clear no to this other data structure. Another downside is that redundant op fetches are far more likely or need to be tracked in some way.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,348 @@
...
+//! - Exposes public method [`CoreFetch::add_ops`] that takes a list of ops and an agent id.
+//! - Stores pairs of ([`OpId`][`AgentId`]) in the order that ops have been added.
+//!
+//! ### Fetch task
+//!
+//! - Toeks pairs of ([`OpId`], [`AgentId`]) from the beginning of the queue and requests the op from the agent.
+//! - Appends removed pair to the end of the queue.
+//! - Put unresponsive peers on a cool-down list.
+//!
+//! ### Incoming op task
```
jost-s, line 24: To be implemented on another branch.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,348 @@
...
+
+const MOD_NAME: &str = "Fetch";
+
+/// Configuration parameters for [CoreFetchFactory].
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CoreFetchConfig {
+    /// How many parallel op fetch requests can be made at once. Default: 2.  
+    parallel_request_count: u8,
+    /// Duration in ms to pause when parallel request count is reached. Default: 10.
```
jost-s, line 56: This just serves for testing really.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+            )
+            .await;
+
+            Ok(())
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+struct Inner {
```
jost-s, line 136: Happy to receive suggestions if this can be improved.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+                        ))
+                        .await;
+                    } else if inner.ops.lock().await.is_empty() {
+                        // Sleep if there are no more ops to fetch.
+                        // Sleep is canceled by newly added ops.
+                        select! {
+                            _ = tokio::time::sleep(Duration::from_millis(
+                                fetch_loop_sleep,
+                            )) => {},
+                            _ = fetch_request_rx.recv() => {}
```
jost-s, line 272: Is this wake-up mechanism a premature optimization from your POV?

jost-s:
This is the first part of the fetch module now. It contains the fetch data object and the fetch task with tests for it. The tests use a mock transport with a simplified interface, testing that requests are sent as expected. I plan to create two more PRs for the fetch module, one for the fetch response queue that handles incoming fetch requests, and the other to add a handler for the incoming responses to fetch requests.

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,34 @@
+//! Kitsune2 fetch types.
+
+use std::sync::Arc;
+
+use crate::{builder, config, AgentId, BoxFut, K2Result, OpId};
+
+/// Trait for implementing a fetch module to fetch ops from other agents.
+pub trait Fetch: 'static + Send + Sync {
```
neonphog, line 8: Please add the debug requirement. When working with trait objects it's important to be able to get *some* information out of it when testing / tracing.

crates/api/src/tx.rs
```diff
@@ -0,0 +1,14 @@
...
+
+/// Trait for implementing a transport module for exchanging messages
+/// between agents.
+pub trait Transport: 'static + Send + Sync + std::fmt::Debug {
+    /// Send a request for ops to a peer.
+    fn send_op_request(
+        &mut self,
+        op_id: OpId,
+        dest: AgentId,
+    ) -> BoxFut<'static, K2Result<()>>;
```
neonphog, line 13: This is what the call currently looks like in my wip pr.

```suggestion
    /// Notify a remote peer module within a space. This is a fire-and-forget
    /// type message. The future this call returns will indicate any errors
    /// that occur up to the point where the message is handed off to
    /// the transport backend. After that, the future will return `Ok(())`
    /// but the remote peer may or may not actually receive the message.
    pub async fn send_module(
        &self,
        peer: Url,
        space: SpaceId,
        module: String,
        data: bytes::Bytes,
    ) -> BoxFut<'static, K2Result<()>>;
```

crates/core/Cargo.toml
```diff
@@ -14,9 +14,18 @@ edition = "2021"
...
+indexmap = "2.7.0"
 kitsune2_api = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+tokio = { workspace = true, features = [
+    "macros",
+    "rt",
+    "rt-multi-thread",
+    "sync",
+    "time",
```
neonphog, line 26: We should not need `macros` or `rt-multi-thread` in library code, right?
- task comes from `rt`
- channels come from `sync`
- timeout comes from `time`

```suggestion
    "rt",
    "sync",
    "time",
```

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+    /// Get fetch loop pause interval.
+    pub fn fetch_loop_pause(&self) -> u64 {
+        self.fetch_loop_sleep
+    }
+    /// Get cool-down interval.
+    pub fn cool_down_interval(&self) -> u64 {
+        self.cool_down_interval
+    }
+}
```
neonphog, line 94: I'd delete these pass-through accessors and just make the fields pub.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+
+// TODO: Temporary trait object of a transport module to facilitate unit tests.
+type DynTransport = Arc<Mutex<dyn Transport>>;
```
neonphog, line 97:
```suggestion
type DynTransport = Arc<dyn Transport>;
```

No mutex here : )

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+}
+
+#[derive(Clone, Debug)]
+struct Inner {
+    config: CoreFetchConfig,
+    // An `IndexMap` is used to retain order of insertion and have the ability to look up elements
+    // by key efficiently. Ops may be added redundantly to the map with different sources to fetch from, so
+    // the map is keyed by op and agent id together.
+    // The value field could be used to track number of attempts for example.
+    ops: Arc<Mutex<IndexMap<(OpId, AgentId), ()>>>,
```
neonphog, line 142: there is an [IndexSet](https://docs.rs/indexmap/latest/indexmap/set/struct.IndexSet.html) if you aren't storing values.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+                } = inner.config;
+
+                // Loop for processing op/agent pairs in the queue. Runs continuously and is paused
+                // briefly when the maximum of parallel requests is reached, and sleeps for a
+                // longer period when the queue is empty. It is awoken from sleep through a channel
+                // when new op ids are added to the queue.
+                loop {
+                    // Send new request if parallel request count is not reached.
+                    if parallel_request_count
+                        > (*current_request_count.lock().await)
```
neonphog, line 186: I find that it's easier to understand parallelism when there are actually parallel tasks: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=606abba6cb763c472a2037e6bf24e1ce

This way when we get to awaiting the response, it can just do that inline with a timeout and you won't need your waker channel anymore. If you get a timeout, mark the agent unreliable and send the op into limbo. Otherwise pull the next item to fetch.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+#[derive(Clone, Debug)]
+struct Inner {
+    config: CoreFetchConfig,
+    // An `IndexMap` is used to retain order of insertion and have the ability to look up elements
+    // by key efficiently. Ops may be added redundantly to the map with different sources to fetch from, so
+    // the map is keyed by op and agent id together.
+    // The value field could be used to track number of attempts for example.
+    ops: Arc<Mutex<IndexMap<(OpId, AgentId), ()>>>,
+    current_request_count: Arc<Mutex<u8>>,
+    cool_down_list: Arc<Mutex<HashMap<AgentId, Instant>>>,
```
neonphog, line 144: These three mutexes are around basic data structures that are not async. So they should not be tokio mutexes, but just regular `std::sync::Mutex`s. Furthermore, dealing with multiple mutexes creates a lot of opportunity for race conditions or deadlocks. I suggest having a top level Mutex `struct CoreFetch(std::sync::Mutex<Inner>);` and then sending any async work you need to do through a channel to a tokio task.

crates/core/src/factories/core_fetch.rs
neonphog: I think the sleeping itself is re-inventing what channels are for. When you await on a channel, if there is no data, all the logic for pausing the task and re-waking it is built in.

crates/core/src/factories/core_fetch.rs
jost-s: Yes, that seems like the missing piece I had in my thinking. That looks much easier to understand. There are 2 tasks which are the 2 parallel requesters, I understand. In the actual code the `for 0..5` loop is replaced by just a `loop {}`, right?

crates/core/Cargo.toml
```diff
@@ -14,9 +14,18 @@ edition = "2021"
 bytes = { workspace = true }
 ed25519-dalek = { workspace = true }
 futures = { workspace = true }
+indexmap = "2.7.0"
```
ThetaSinner, line 17: Not a workspace dependency?

crates/core/Cargo.toml
```diff
@@ -14,9 +14,18 @@ edition = "2021"
...
+tokio = { workspace = true, features = [
+    "macros",
+    "rt",
+    "rt-multi-thread",
+    "sync",
+    "time",
+] }
 
 [dev-dependencies]
+rand = "0.8.5"
```
ThetaSinner, line 30: This is likely something we'll use a few places too. I'd probably just go with workspace deps here too

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+const MOD_NAME: &str = "Fetch";
+
+/// Configuration parameters for [CoreFetchFactory].
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CoreFetchConfig {
+    /// How many parallel op fetch requests can be made at once. Default: 2.  
+    parallel_request_count: u8,
+    /// Duration in ms to pause when parallel request count is reached. Default: 10.
+    parallel_request_pause: u64,
```
ThetaSinner, line 57: It would be nice if this worked with blocking and waiting for capacity to be available. I'm gussing at this point but I'm assuming I'm going to see a sleep with this value further down

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+    /// Duration in ms to keep an unresponsive agent on the cool-down list. Default: 10_000.
+    cool_down_interval: u64,
+}
+
+impl Default for CoreFetchConfig {
+    fn default() -> Self {
+        Self {
+            parallel_request_count: 2,
+            parallel_request_pause: 10,
+            fetch_loop_sleep: 1_000,
```
ThetaSinner, line 69: Please can all these values have units in the name? I get that we can't type them and simply receive them through some unknown interface but it'd be good to know if it's `_s` or `_ms`

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+impl Fetch for CoreFetch {
+    fn add_ops(
+        &mut self,
+        op_list: Vec<OpId>,
+        source: AgentId,
+    ) -> BoxFut<'_, K2Result<()>> {
+        Box::pin(async move {
```
ThetaSinner, line 114: Is this different to doing 

```rust
async move {
     // do stuff
}.boxed()
```

? I slightly prefer the less syntax but it'd be nice to be consistent with whichever we go with

crates/core/src/factories/core_fetch.rs
ThetaSinner: I'm struggling to get comfortable with this structure where we have sources for the same op separated. I know we agreed to go simpler on this so I'm fine for now but I'm hoping we'll try to optimize this a bit if we can demonstrate that we're going an unreasonable amount of duplicate fetches later

crates/core/src/factories/core_fetch.rs
neonphog: Yeah, actually probably a `while let Some(bla) = ` loop. The for was just so the example would end : )

crates/core/src/factories/core_fetch.rs
neonphog: They are identical. I've been preferring the Box::pin version because it doesn't require an external library, it's built into rust.

crates/core/src/factories/core_fetch.rs
jost-s: Yes, for sure. For the moment there are assertions for that in the tests even
https://github.com/holochain/kitsune2/pull/35/files#diff-befa5f16b403d32326c69f97ef3ead25c6d7a78f93845c5cda7e6d08d7865a1cR118
and
https://github.com/holochain/kitsune2/pull/35/files#diff-befa5f16b403d32326c69f97ef3ead25c6d7a78f93845c5cda7e6d08d7865a1cR207

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+                                                    agent_id,
+                                                    Instant::now(),
+                                                );
+                                        }
+                                    }
+                                });
+                            }
+
+                            // Pause to give other requests a chance to be processed before re-adding this op.
+                            tokio::time::sleep(Duration::from_millis(1)).await;
```
mattyg, line 246: Is this a meaningful amount of delay? It seems quite short.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+
+#[derive(Debug)]
+struct State {
+    ops: HashSet<FetchRequest>,
+    cool_down_list: HashMap<AgentId, Instant>,
+}
+
+#[derive(Debug)]
+struct Inner {
+    state: Arc<Mutex<State>>,
```
jost-s, line 138: One mutex for all state objects.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+                self.fetch_queue_tx.clone(),
+                self.fetch_queue_rx.clone(),
+                transport.clone(),
+                self.state.clone(),
+                config.cool_down_interval_ms,
+            ));
+        }
+    }
+
+    async fn fetch_task(
```
jost-s, line 178: If this function signature can be improved or easily converted into a method and that makes sense, please let me know.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+        cool_down_list: &mut HashMap<AgentId, Instant>,
+        cool_down_interval: u64,
+    ) -> bool {
+        let a = cool_down_list.get(agent_id);
+        match a {
+            Some(instant) => {
+                let now = Instant::now();
+                if (now - *instant).as_millis() > cool_down_interval as u128 {
+                    // Cool down interval has elapsed. Remove agent from list.
+                    cool_down_list.remove(agent_id);
```
jost-s, line 241: This is side-effect-esque, but I think it's fine because it's just purging a stale entry. Otherwise I'd need to run this check again in a separate function to purge, and I didn't want to because the function would be run on every fetch request in the queue.

jost-s:
@neonphog @ThetaSinner I'm happy with this now, it's simple and easy to follow. Thanks for the guidance!

You may find something to improve about the structure of the data objects. @neonphog I could refactor the transport interface now, or is this better done on a separate PR once your API stub is merged?

crates/core/src/factories/core_fetch.rs
jost-s: I ended up with a mutex for the data structures and additional fields of the inner struct. In the end it's a tokio mutex, because it's doing async work.

crates/core/src/factories/core_fetch.rs
jost-s: Yes, done

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,341 @@
...
+            op_list.clone().into_iter().for_each(|op_id| {
+                ops.insert((op_id, source.clone()), ());
+            });
+            drop(ops);
+
+            // Wake up fetch task in case it is sleeping. Time out if the channel buffer
+            // is full.
+            let _ = tokio::time::timeout(
+                Duration::from_millis(5),
+                self.0.fetch_request_tx.send(()),
```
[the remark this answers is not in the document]
jost-s: It's no longer needed to wake the tasks now.

crates/core/src/factories/core_fetch.rs
jost-s: There's no longer a delay.

Cargo.toml
```diff
@@ -3,8 +3,10 @@ members = [
   "crates/api",
   "crates/bootstrap_srv",
   "crates/dht",
+  "crates/fetch_queue",
   "crates/memory",
 ]
+
```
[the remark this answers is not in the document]
ThetaSinner: Done

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+//! - Fetch tasks that request tracked ops from agents
+//! - Incoming op task that processes incoming responses to op requests by
+//!     - persisting ops to the data store
+//!     - removing op ids from in-memory data object
+//!
+//! ### State object [CoreFetch]
+//!
+//! - Exposes public method [CoreFetch::add_ops] that takes a list of op ids and an agent id.
+//! - Stores pairs of ([OpId][AgentId]) in a set.
+//! - A hash set` is used to look up elements by key efficiently. Ops may be added redundantly
```
ThetaSinner, line 17:
```suggestion
//! - A hash set is used to look up elements by key efficiently. Ops may be added redundantly
```

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    async fn fetch_task(
+        fetch_request_tx: Sender<FetchRequest>,
+        fetch_request_rx: Arc<Mutex<Receiver<FetchRequest>>>,
+        transport: DynTransport,
+        state: Arc<Mutex<State>>,
+        cool_down_interval: u64,
+    ) {
+        while let Some((op_id, agent_id)) =
+            fetch_request_rx.lock().await.recv().await
```
ThetaSinner, line 186: The lock here is immediately dropped when the `recv` returns a value, so that the other task can proceed? It looks like it should, but is that guaranteed?

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+                    state
+                        .lock()
+                        .await
+                        .cool_down_list
+                        .insert(agent_id.clone(), Instant::now());
+                }
+            }
+
+            // Re-insert the fetch request into the queue.
+            if let Err(err) = fetch_request_tx
```
ThetaSinner, line 221: As above, this reads to me like it can deadlock. If gossip discovers a batch of ops and fills the channel, then both fetch tasks experience an error, this won't be able to send and it is also responsible for draining the channel.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+        source: AgentId,
+    ) -> BoxFut<'_, K2Result<()>> {
+        Box::pin(async move {
+            // Add ops to set.
+            let ops = &mut self.0.state.lock().await.ops;
+            op_list.clone().into_iter().for_each(|op_id| {
+                ops.insert((op_id, source.clone()));
+            });
+
+            // Pass ops to fetch tasks.
```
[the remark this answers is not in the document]
jost-s: It's a mutable reference and the compiler told me that dropping it has no effect.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+struct Inner {
+    state: Arc<Mutex<State>>,
+    fetch_queue_tx: Sender<FetchRequest>,
+    fetch_queue_rx: Arc<Mutex<Receiver<FetchRequest>>>,
+}
+
+impl Inner {
+    pub fn new() -> Self {
+        // Create a channel to send new ops to fetch to the tasks. This is in effect the fetch queue.
+        let (fetch_queue_tx, fetch_queue_rx) = channel::<FetchRequest>(1024);
```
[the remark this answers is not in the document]
jost-s: Yes, that could happen, I believe. The tasks would pick up an item from the channel. At that moment add_ops would add another item, which prevents the tasks to re-insert the just requested item into the channel and the queue would come to a halt.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+const MOD_NAME: &str = "Fetch";
+
+/// Configuration parameters for [CoreFetchFactory].
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CoreFetchConfig {
+    /// How many parallel op fetch requests can be made at once. Default: 2.  
+    pub parallel_request_count: u8,
+    /// Duration in ms to keep an unresponsive agent on the cool-down list. Default: 10_000.
+    pub cool_down_interval_ms: u64,
```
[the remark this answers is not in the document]
jost-s: done #259aac531c78dfe62df2d067dd636b28f249a284

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/d13f217f86acb8715a3e1fb314e33d83012b9091

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,301 @@
...
+            op_list.clone().into_iter().for_each(|op_id| {
+                ops.insert((op_id, source.clone()));
+            });
+
+            // Pass ops to fetch tasks.
+            let futures = op_list.into_iter().map(|op_id| async {
+                if let Err(err) =
+                    self.0.fetch_queue_tx.send((op_id, source.clone())).await
+                {
+                    eprintln!(
```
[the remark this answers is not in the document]
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/d4a056aca87dcc8bfeb91fc81d83257deb62ec80

crates/core/src/factories/core_fetch.rs
jost-s: done https://github.com/holochain/kitsune2/commit/d13f217f86acb8715a3e1fb314e33d83012b9091

jost-s:
> @neonphog I could refactor the transport interface now, or is this better done on a separate PR once your API stub is merged?

I've refactored to a naive implementation. Please let me know if that makes sense.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
neonphog:
```suggestion
        let (fetch_queue_tx, fetch_queue_rx) = channel::<FetchRequest>(16_384);
```

crates/core/src/factories/core_fetch.rs
neonphog: This needs to not be a tokio Mutex. Please switch to a std::sync::Mutex.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,328 @@
...
+struct State {
+    ops: HashSet<FetchRequest>,
+    cool_down_list: CoolDownList,
+}
+
+#[derive(Debug)]
+struct Inner {
+    state: Arc<Mutex<State>>,
+    fetch_queue_tx: Sender<FetchRequest>,
+    fetch_queue_rx: Arc<Mutex<Receiver<FetchRequest>>>,
```
neonphog, line 143: There is no reason the receiver side needs to be in your top level class. I'd spawn the tokio threads in your constructor so you don't need this here.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,328 @@
...
+        parallel_request_count: u8,
+        transport: DynTransport,
+    ) {
+        for _ in 0..parallel_request_count {
+            tokio::task::spawn(Inner::fetch_task(
+                self.fetch_queue_tx.clone(),
+                self.fetch_queue_rx.clone(),
+                transport.clone(),
+                self.state.clone(),
+            ));
```
neonphog, line 176: The join handles of these task should be tracked in your top-level struct like `Vec<tokio::task::JoinHandle<()>>` so that they can be aborted on drop:

```rust
impl Drop for Inner /* or CoreFetch */ {
    fn drop(&mut self) {
        for t in self.task_list.iter() {
            t.abort();
````

File: crates/core/src/factories/core_fetch.rs [hunk 42]
```diff
@@ -0,0 +1,328 @@
...
+            {
+                // Send fetch request to agent.
+                // TODO: update when transport is available
+                let mut data = bytes::BytesMut::new();
+                data.put(op_id.clone().0 .0);
+                data.put(agent_id.clone().0 .0);
+                let data = data.freeze();
+                if let Err(err) = transport
+                    .send_module(
+                        Url::from_str("wss://0.0.0.0:443").unwrap(),
```
neonphog, line 214: Heh... woo, there's a lot to figure out here : ) -- maybe doesn't have to be part of this PR, but we'll need to query the peer_store for a url to set here.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,328 @@
...
+                        SpaceId::from(bytes::Bytes::new()),
+                        "Mod".to_string(),
```
neonphog, line 216:
```suggestion
                        "Fetch".to_string(),
```

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,328 @@
...
+                // Send fetch request to agent.
+                // TODO: update when transport is available
+                let mut data = bytes::BytesMut::new();
+                data.put(op_id.clone().0 .0);
+                data.put(agent_id.clone().0 .0);
+                let data = data.freeze();
+                if let Err(err) = transport
+                    .send_module(
+                        Url::from_str("wss://0.0.0.0:443").unwrap(),
+                        SpaceId::from(bytes::Bytes::new()),
```
neonphog, line 215: This should be injected by your Factory create function and passed through so it's available here.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,328 @@
...
+            {
+                continue;
+            }
+
+            // Check if agent is on cool-down list.
+            if !state
+                .lock()
+                .await
+                .cool_down_list
+                .is_agent_cooling_down(&agent_id)
```
neonphog, line 204: Only lock once (assuming the switch to std::sync::Mutex is made):

```suggestion
            let is_agent_cooling_down = {
                // It is okay to unwrap std::sync::Mutexes
                // we *want* poisoned threads to panic
                // their calling threads!
                let lock = state.lock().unwrap();
                if lock.ops.contains(&(op_id.clone(), agent_id.clone())) {
                    continue;
                }
                lock.cool_down_list.is_agent_cooling_down(&agent_id)
            };
```

neonphog (changes requested):
Sorry, I haven't gotten to the tests yet, I'll try to look at those during our meetings tomorrow...

crates/core/src/factories/core_fetch.rs
jost-s: Ha, that's how it was before, and it looked a bit cleaner with the separate constructor.

crates/core/src/factories/core_fetch.rs
jost-s: Switched https://github.com/holochain/kitsune2/pull/35/commits/9add269447278ecaf2a1e4a4fad8142cf8c61011

Must the receiver use a tokio mutex?

crates/core/src/factories/core_fetch.rs
jost-s: Reverted https://github.com/holochain/kitsune2/pull/35/commits/5fd4e84b2ad874141ac96443f738d069fbe5d4f5

crates/core/src/factories/core_fetch.rs
jost-s: Alright! https://github.com/holochain/kitsune2/pull/35/commits/2a156d55b84470e058faf4366bbdd3a8b71a063e

crates/core/src/factories/core_fetch.rs
jost-s: I added the peer store to the factory create method, but got stuck in the tests. https://github.com/holochain/kitsune2/pull/35/commits/f90981e0b69f31213ec7f27e682ec7fd1e889d06

crates/core/src/factories/core_fetch.rs
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/dd7985158ec70cb59bf62bcaa12c17a819b69441

crates/core/src/factories/core_fetch.rs
jost-s: Yes, updated https://github.com/holochain/kitsune2/commit/f90981e0b69f31213ec7f27e682ec7fd1e889d06

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/872fc8cccd9adf5e2a15cf68d05dfea14ae0b21b

crates/core/src/factories/core_fetch.rs
jost-s: also done https://github.com/holochain/kitsune2/pull/35/commits/9add269447278ecaf2a1e4a4fad8142cf8c61011

jost-s:
Thanks for all the feedback! Peer store pending, the rest is modified as suggested.

crates/core/src/factories/core_fetch.rs
neonphog: Correct, when you need to await anything while a mutex is locked you need it to be a tokio mutex. But that makes them much slower system-wise, so should only be used in very calculated situations.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+
+impl ModConfig for CoreFetchConfig {}
+
+// TODO: Temporary trait object of a transport module to facilitate unit tests.
+type DynTransport = Arc<dyn Transport>;
+
+#[derive(Debug)]
+struct CoreFetch(Inner);
```
neonphog, line 129: If the mutex isn't going to be at this level, there's no need for Inner. You could implement Fetch on Inner. Then rename Inner to CoreFetch.

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+use rand::Rng;
+use tokio::sync::Mutex;
+
+use crate::default_builder;
+
+use super::{CoreFetch, CoreFetchConfig, Transport};
+
+#[derive(Debug)]
+pub struct MockTransport {
+    requests_sent: Arc<Mutex<Vec<(OpId, AgentId)>>>,
```
neonphog, line 14: This contains simple types, should use std::sync::Mutex intead of the tokio one.

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+        loop {
+            let requests_sent =
+                mock_transport.requests_sent.lock().await.clone();
+            if requests_sent.len() >= num_ops {
+                op_list.clone().into_iter().all(|op_id| {
+                    requests_sent.contains(&(op_id, agent.clone()))
+                });
+                break;
+            }
+        }
```
neonphog, line 164: this is another busy loop

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+            let requests_sent =
+                mock_transport.requests_sent.lock().await.clone();
+            if requests_sent.len() >= total_ops
+                && expected_ops
+                    .iter()
+                    .all(|expected_op| requests_sent.contains(expected_op))
+            {
+                break;
+            }
+        }
```
neonphog, line 233: this is another busy loop

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+                    .0
+                    .state
+                    .lock()
+                    .unwrap()
+                    .cool_down_list
+                    .is_agent_cooling_down(&agent)
+            {
+                break;
+            }
+        }
```
neonphog, line 270: this is another busy loop

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+                let cool_down_list =
+                    &mut fetch.0.state.lock().unwrap().cool_down_list;
+                if expected_agents
+                    .iter()
+                    .all(|agent| cool_down_list.is_agent_cooling_down(agent))
+                {
+                    break;
+                }
+            }
+        }
```
neonphog, line 380: this is another busy loop

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    // Clear set of ops to fetch to stop sending requests.
+    fetch.0.state.lock().unwrap().ops.clear();
+
+    let mut num_requests_sent = mock_transport.requests_sent.lock().await.len();
+
+    // Wait for tasks to settle all requests.
+    tokio::time::timeout(Duration::from_millis(10), async {
+        loop {
+            tokio::time::sleep(Duration::from_millis(1)).await;
```
neonphog, line 96: yay, this one is not a busy loop! : )

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+        _space: kitsune2_api::SpaceId,
+        _module: String,
+        mut data: bytes::Bytes,
+    ) -> kitsune2_api::BoxFut<'_, kitsune2_api::K2Result<()>> {
+        Box::pin(async move {
+            let op_id_bytes = data.split_to(data.len() / 2);
+            let agent_id_bytes = data;
+            let op_id = OpId::from(op_id_bytes);
+            let agent_id = AgentId::from(agent_id_bytes);
+            self.requests_sent.lock().await.push((op_id, agent_id));
```
neonphog, line 42: do we still want to push it if it is `self.fail`? Or maybe the fail-state should be tracked in the list too?

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,355 @@
...
+                    Err(err) => {
+                        tracing::warn!("could not get agent id {agent_id} from peer store: {err}");
+                        continue;
+                    }
+                };
+
+                let mut data = bytes::BytesMut::new();
+                data.put(op_id.clone().0 .0);
+                data.put(agent_id.clone().0 .0);
+                let data = data.freeze();
```
neonphog, line 278:
```suggestion
                // Using this simple "encoding" for now.
                // We can evaluate in the future if we need
                // anything else more complicated.
                let mut data = bytes::BytesMut::new();
                data.put(op_id.clone().0 .0);
                data.put(agent_id.clone().0 .0);
                let data = data.freeze();
```

crates/core/src/factories/core_fetch/test.rs
jost-s: lol, I have a deja-vu!
Yes, they were at 10 ms before and that was far too long, so I thought it doesn't make sense to use any sleeps. I'm not used to such low fractions of time, in the Holochain repo it doesn't matter whether you wait for a second or two. =)

crates/core/src/factories/core_fetch.rs
neonphog: Actually, now that I've commented on this, I'm not sure this is sufficient. The kitsune api doesn't provide any requirements that the ids are the same length, so we can't decode this with just this information.

crates/core/src/factories/core_fetch/test.rs
neonphog: hehe, yea, you could even use `tokio::task::yield_now().await`... really we just want to give the other task a chance to trigger, because when a task gets cpu time, it'll be optimized to get a mutex lock over and over, but the processing tasks won't get a chance to change anything in the mean time.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
ThetaSinner: Hmm, I mean that is better in the case that the agent is gone. It's probably still okay in the case of an interrupted connection. It would be less good in the case of a flaky connection.

I think we should come back to this later and put in a full exponential backoff

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
ThetaSinner: Hmm that's odd. Do you have to `drop(*ops)`?

I would expect the `ops` reference to persist to the end of this scope.

If it really fights back then maybe just use a block to scope that first operation?

crates/core/src/factories/core_fetch.rs
ThetaSinner: Yes, ideally we use serializable types where we're sending messages like this I think.

I was just looking for this code to say, can we make sure we make the "ops to get" part of the request a list. Then it won't be a breaking protocol change if we want to batch the gets

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,355 @@
...
+                        .lock()
+                        .unwrap()
+                        .cool_down_list
+                        .add_agent(agent_id.clone());
+                }
+            }
+
+            // Re-insert the fetch request into the queue.
+            if let Err(err) =
+                fetch_request_tx.try_send((op_id.clone(), agent_id.clone()))
```
ThetaSinner, line 300: What is the re-drive logic? If we fail to try-send then how do we make sure that ops sitting in the fetch set make it back into the channel? 

I still think this is necessary to avoid deadlocking, but it introduced a new problem too

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    pub fn add_agent(&mut self, agent_id: AgentId) {
+        self.state.insert(agent_id, Instant::now());
+    }
+
+    pub fn is_agent_cooling_down(&mut self, agent_id: &AgentId) -> bool {
+        match self.state.get(agent_id) {
+            Some(instant) => {
+                let now = Instant::now();
+                if (now - *instant).as_millis()
```
ThetaSinner, line 338: This can just use https://doc.rust-lang.org/std/time/struct.Instant.html#method.elapsed

crates/core/src/factories/core_fetch.rs
jost-s: Yep, simpler.

crates/core/src/factories/core_fetch.rs
```diff
@@ -0,0 +1,355 @@
...
+                op_list
+                    .clone()
+                    .into_iter()
+                    .map(|op_id| (op_id.clone(), source.clone())),
+            );
+
+            // Pass ops to fetch tasks.
+            op_list.into_iter().for_each(|op_id| {
+                if let Err(err) =
+                    self.0.fetch_queue_tx.try_send((op_id, source.clone()))
```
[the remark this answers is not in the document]
jost-s: Necessary, I guess not. Yes, that was the purpose.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: Ah, that's the difference between cool-down and back-off for you. Perhaps an exponential backoff is better. It would mean that there is repeated successful gossip with the peer, but when fetching ops the connection is flaky, then good again when gossiping. A problem when used maliciously.

crates/core/src/factories/core_fetch.rs
jost-s: It would make sense to only re-insert an op if the fetch request could be sent successfully. If the agent was unresponsive and then we add it back to the queue, chances are lower that the op will be received in the meantime until we reach the point of this op again in the queue. Whereas if the fetch request was successful, chances are higher that the op will be received while processing other elements in the queue, so that the op will be gone from the ops to fetch list the next we get to this point in the queue.

jost-s: That's got nothing to do with your question. I think all we can do is remove the item from the set if re-insertion fails.

crates/core/src/factories/core_fetch.rs
jost-s: https://github.com/holochain/kitsune2/pull/35/commits/e0c11665a8f6cb3518bf103c72f61fadaaffb418

crates/core/src/factories/core_fetch/test.rs
jost-s: Hidden in here https://github.com/holochain/kitsune2/pull/35/commits/05acb3110f09fffbe50ec38ad7d1ddc233e6f932

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,409 @@
...
+    fetch.add_ops(op_list, agent_id.clone()).await.unwrap();
+
+    // Let the fetch request be sent multiple times. As only 1 op was added to the queue,
+    // this proves that it is being re-added to the queue after sending a request for it.
+    tokio::time::timeout(Duration::from_millis(10), async {
+        loop {
+            if mock_transport.requests_sent.lock().await.len() >= 3 {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
jost-s: All busy loops changed here too https://github.com/holochain/kitsune2/pull/35/commits/05acb3110f09fffbe50ec38ad7d1ddc233e6f932

crates/core/src/factories/core_fetch/test.rs
jost-s: I'd still push it because a request was sent and that's what I want to check in the tests.

crates/core/src/factories/core_fetch.rs
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/9ac37438280fe7e14884b933d4d38a901aca2ab8

jost-s:
@neonphog I was about to merge `main` into this PR, but I don't get what to do with the transport. I had that question on the PR but forgot to ask it.
So there's [`send_module`](https://github.com/holochain/kitsune2/blob/main/crates/api/src/transport.rs#L249) which must be called by the fetch module, but it's an impl of `Transport` which is a struct. A struct that doesn't impl `DynTxImpl`. It has a member that implements it. I'm stuck here, if I use `Transport` I can't mock transport, but `DynTxImpl` doesn't implement `send_module`. What's the way forward?

crates/core/src/factories/core_fetch.rs
jost-s: Using protobuf, that is?

ThetaSinner: I think that'd be the best option yes, just to start a protocol definition for the fetch module

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
ThetaSinner: As a caller, I can't tell whether these have been queued or dropped. This is actually quite important information. The caller should be pausing their work until this queue drains a bit. Making this `await` space is a simple way to do that, alternatively if we want to try send then the dropped ops could be returned. Or just a count or something.

crates/core/src/factories/core_fetch.rs
ThetaSinner: That works. Then we don't have state building up the set that can't be reached.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: Good, I'll just make it wait.

crates/core/src/factories/core_fetch.rs
jost-s: To clarify, I didn't think of this quick'n dirty way as the final api to construct the data to pass in =D

Protobuf definition done here https://github.com/holochain/kitsune2/pull/35/commits/e27beedb5f7fb7fec536df84eb300b9115f6e58e

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: Done here https://github.com/holochain/kitsune2/pull/35/commits/c2c224db475de13471aca256d76e114e9afa6876

crates/core/src/factories/core_fetch.rs
jost-s: Re-insert ops only when agent responsive and
Remove item from ops set when re-insertion fails https://github.com/holochain/kitsune2/pull/35/commits/0e9d9e7bd79590f7ed13d0dbd4b3c5fb026285ca

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,104 @@
...
+    let bytes = OpIds::from(value).encode_to_vec();
+    bytes::Bytes::copy_from_slice(&bytes)
+}
+
+/// Deserialize list of op ids.
+pub fn deserialize_op_ids(value: bytes::Bytes) -> K2Result<Vec<OpId>> {
+    let op_ids = OpIds::decode(value).map_err(K2Error::other)?;
+    let vec = Vec::from(op_ids);
+    Ok(vec)
+}
```
jost-s, line 39: Added functions to convert back and forth from serialized to deserialized op ids.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: It didn't like `drop(*ops)`. Enclosed in a block https://github.com/holochain/kitsune2/pull/35/commits/94bbc7c32e58c2bcfdea48e90066b6836f5616e1

crates/api/proto/gen/kitsune2.fetch.rs
```diff
@@ -0,0 +1,8 @@
+// This file is @generated by prost-build.
+/// A list of op ids.
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct OpIds {
+    /// Op ids.
+    #[prost(bytes = "bytes", repeated, tag = "1")]
+    pub data: ::prost::alloc::vec::Vec<::prost::bytes::Bytes>,
```
jost-s, line 7: @ThetaSinner takes a vec of op_ids

crates/core/src/factories/mem_peer_store/test.rs
```diff
@@ -23,68 +22,6 @@ fn unsneak_url(u: &Url) -> String {
     u.peer_id().unwrap().into()
 }
 
-#[derive(Debug, Default)]
```
jost-s, line 26: Moved this to `test_utils` under this module and made public for re-use in fetch tests. Is that a good place for it?

jost-s:
Ready for review again. I suggest I add any other changes than corrections to the next PR. This one is already large.

jost-s:
I realized there's a crossing here and we need to decide which route to take. When an agent is unresponsive, they're put on a cool-down list (or exponential back-off at some stage), and ops to be requested from them still in the queue are removed from the ops set, to prevent build of state. But removing those ops makes the whole cool-down list redundant. Once an agent is unresponsive, the ops are removed and no longer considered for fetch requests independent of cool-down list.

So the decision is:
- delete cool-down list and just remove ops for an agent from the set of ops to fetch when they are unresponsive
- do not remove ops for an unresponsive agent from the set of ops to fetch, so that once the cool-down interval has elapsed, they're requested again

First option is definitely the simplest. But it depends on how gossip works. If we're gossiping with an agent in our neighborhood, then they become unresponsive, we stop gossiping and resume it once they become available again, we can just re-add missing ops to the queue.

Writing about this it seems that the queue does not need to manage a cool-down list at this stage and that it should be gossip that handles the case that a peer goes offline and then comes online again and we request missing ops again.

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
ThetaSinner: This will be joined up by the peer behavior module. I agree we should be tracking connectivity across modules and not having each module make different decisions.

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,600 @@
...
+async fn fetch_queue() {
+    let builder = Arc::new(default_builder());
+    let peer_store = builder.peer_store.create(builder.clone()).await.unwrap();
+    let mock_transport = MockTransport::new(false);
+    let config = CoreFetchConfig::default();
+
+    let op_id = random_op_id();
+    let op_list = vec![op_id.clone()];
+    let agent_id = random_agent_id();
+    let agent_info = AgentBuild {
```
ThetaSinner, line 102: `AgentBuilder`?

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,600 @@
...
+    // Op should have been removed from ops to fetch.
+    assert!(
+        fetch.state.lock().unwrap().ops.is_empty(),
+        "ops are {:?}",
+        fetch.state.lock().unwrap().ops.len()
+    );
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn agent_cooling_down_is_removed_from_list() {
```
ThetaSinner, line 533: Are they actually removed from anywhere? Or just tracked as being on cool down? Just checking that the name of the test is accurate, the test content looks good but I'm wondering if I'm missing something here

ThetaSinner (dismissed):

crates/core/src/factories/mem_peer_store/test.rs
neonphog: I'm okay with this for now. I may move this into the test_utils crate later when I've got some more agent building tests to implement : )

neonphog (dismissed):
:tada: nice work @jost-s !

crates/api/src/fetch.rs
```diff
@@ -0,0 +1,106 @@
...
+#[cfg(test)]
+mod test {
+    use crate::id::Id;
+
+    use super::*;
+    use prost::Message;
+
+    #[test]
+    fn happy_encode_decode() {
+        let op_id_1 = OpId(Id(bytes::Bytes::from_static(b"some_op_id")));
```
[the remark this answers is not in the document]
jost-s: True, this is from before I remembered that David had added the impls.

crates/core/src/factories/core_fetch/test.rs
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/8f0b94d0646c2d6c97ffb934a247c40b75b5cb1e

crates/api/src/fetch.rs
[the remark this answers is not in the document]
jost-s: done https://github.com/holochain/kitsune2/pull/35/commits/c024fadef594f128eb0cde5a91844d1779edde53
Instead of `OpId::from` I could just initialize bytes and then do `op_id_1.into()` later on. I prefer to use `OpId::from`.

jost-s:
> I wouldn't want that. I think it's okay for a few packets to go missing here and there. I'd prefer that we try a few times with an exponential back-off and then if they're still not responsive we can drop their op hashes. If we're too aggressive with dropping hashes with individual network timeouts then we risk putting extra work onto gossip to rediscover hashes

I agree with you. It's kind of a conundrum though. If ops are removed from the set of ops to fetch when an agent is unresponsive, there's no point to re-insert it into the queue.

If ops of an unresponsive agent are not re-insert into the queue, the cool-down list is pointless too. This is the one extreme.

The other direction is to always re-insert requests into the queue. The ops were removed from the set to prevent build-up of state, in case the agent never becomes responsive again. But, as said earlier, if the ops are not in the set, re-inserting makes no sense. So the middle path would keep the ops in the set, and add logic to remove them after some time if the agent remains unresponsive.

How about that?

crates/core/src/factories/core_fetch/test.rs
jost-s: They are removed, but no explicitly. As it is right now, `is_agent_cooling_down` removes the agent from the list if the cool-down interval has elapsed. But see below for an extended discussion.

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -0,0 +1,600 @@
...
+impl MockTransport {
+    fn new(fail: bool) -> Arc<MockTransport> {
+        Arc::new(Self {
+            requests_sent: Arc::new(Mutex::new(Vec::new())),
+            fail,
+        })
+    }
+}
+
+impl Transport for MockTransport {
```
jost-s, line 34: Ah, by the way I tried `mockall` for this, but I still had to write the `unimplemented` fns.

ThetaSinner:
> @ThetaSinner I went ahead and implemented that. I'm quite happy with it now, although the back off probably needs to become a stop at some stage. You can look at the code, if you don't agree, I'll revert.

I think what you're talking about is in the right direction. I think the queue needs to become a bit less pivotal in this design and the data structure needs to take on more responsibility. So I'm onboard with the changes but I'm struggling to follow changes at this point. How much of a pain would it be to put that change on another PR and I'll approve the rest so we can get it merged? :)

Otherwise I'll just get on with the review

jost-s:
Sure, I can do that.

ThetaSinner (approved):
Thank you, let's get this bit change in and iterate from there as required :)

PR #37, opened 2024-12-02
Eliminate the origin time

ThetaSinner:
Getting rid of the origin time means one less value that user has to understand and set. It was a limited value concept anyway because you have to set it to when you create the app so that everyone sees the value origin time. Then if somebody installs the app 6 months later then they have to gossip 6 months worth of empty hashes anyway.

This replaces the origin time concept with starting from the UNIX_TIMESTAMP (i.e. a 0 valued Timestamp, because that's how our Timestamp is defined). Any empty hashes are represented as a `(initial_time_slice_id, count)` pair.

So for a network with a slice period of 1 week, that might look something like:
- UNIX timestamp, through to 3 weeks ago, no data stored
- 3 weeks ago to 2 weeks ago, a combined hash stored
- 2 weeks ago to 1 week ago, a combined hash stored
- From some time this week, until the end of time, no data stored

The only thing I want to call out about this is that the Holochain host implementation for Kitsune will need to implement a persistent version of what I've implemented in memory. I've tried to make it reasonably straight forward but it will take a bit of thought to get the same thing done in a database.

### DIFF
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -512,7 +512,7 @@ mod tests {
         // The partial should be:
         //   - The minimum factor size, and
         //   - start at the UNIX_TIMESTAMP
-        //   - end at the UNIX_TIMESTAMP+ UNIT_TIME
+        //   - end at the UNIX_TIMESTAMP + UNIT_TIME
         assert_eq!(0, pt.partial_slices[0].size);
         assert_eq!(UNIX_TIMESTAMP, pt.partial_slices[0].start);
         assert_eq!(UNIX_TIMESTAMP + UNIT_TIME, pt.partial_slices[0].end());
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -287,7 +287,9 @@ impl PartitionedTime {
 
             let hash = combine_op_hashes(op_hashes);
 
-            store.store_slice_hash(self.full_slices, hash).await?;
+            if !hash.is_empty() {
+                store.store_slice_hash(self.full_slices, hash).await?;
+            }
 
             self.full_slices += 1;
             full_slices_end_timestamp += self.full_slice_duration;
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -1050,6 +1050,7 @@ mod tests {
         )
         .await
         .unwrap();
+        let initial_full_slices_count = pt.full_slices;
 
         // Quick calculation of how many full slices we can fit into the current time, leaving space
         // for the minimum recent time.
@@ -1069,27 +1070,25 @@ mod tests {
             store.retrieve_slice_hash(5203984823).await.unwrap();
         assert!(some_slice_hash.is_none());
 
-        println!("Full slices: {}", store.slice_hash_count().await.unwrap());
-
         // Now insert an op at the current time
         store
             .process_incoming_ops(vec![MetaOp {
                 op_id: OpId::from(bytes::Bytes::from(vec![7; 32])),
-                timestamp: Timestamp::now(),
+                timestamp: (Timestamp::now() - pt.full_slice_duration).unwrap(),
                 op_data: vec![],
             }])
             .await
             .unwrap();
         // and compute the new state in the future
-        pt.update(Timestamp::now() + 2 * pt.full_slice_duration, store.clone())
+        pt.update(Timestamp::now() + pt.full_slice_duration, store.clone())
             .await
             .unwrap();
 
-        // Then a single full slice should have been created as the current time
+        // Then a single full slice should have been created at the current time
         assert!(store.slice_hash_count().await.unwrap() > 15_000);
         // and the count should match the number of full slices that the time partition claims to
         // have created.
-        assert_eq!(store.slice_hash_count().await.unwrap(), pt.full_slices);
+        assert_eq!(store.slice_hash_count().await.unwrap(), initial_full_slices_count + 1);
     }
 
     fn validate_partial_slices(pt: &PartitionedTime) {
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -1085,8 +1085,11 @@ mod tests {
             .await
             .unwrap();
 
-        // Then a single full slice should have been created
+        // Then a single full slice should have been created as the current time
         assert!(store.slice_hash_count().await.unwrap() > 15_000);
+        // and the count should match the number of full slices that the time partition claims to
+        // have created.
+        assert_eq!(store.slice_hash_count().await.unwrap(), pt.full_slices);
     }
 
     fn validate_partial_slices(pt: &PartitionedTime) {
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -54,16 +54,14 @@
 //! slices to be stored but is less granular when comparing time slices with another peer.
 
 use crate::constant::UNIT_TIME;
-use kitsune2_api::{DynOpStore, K2Error, K2Result, OpId, Timestamp};
+use kitsune2_api::{
+    DynOpStore, K2Error, K2Result, OpId, Timestamp, UNIX_TIMESTAMP,
+};
 use std::time::Duration;
 
 #[derive(Debug)]
 #[cfg_attr(test, derive(Clone, PartialEq))]
 pub struct PartitionedTime {
-    /// The timestamp at which the time partitioning starts.
-    ///
-    /// This must be a constant among peers who need to compute the same time slices.
-    origin_timestamp: Timestamp,
     /// The factor used to determine the size of the time slices.
     ///
     /// The size of a time slice is 2^factor * [UNIT_TIME].
@@ -137,11 +135,11 @@ impl PartitionedTime {
     /// The resulting [PartitionedTime] will be consistent with the store and the current time.
     /// It should be updated again after [PartitionedTime::next_update_at].
     pub async fn try_from_store(
-        origin_timestamp: Timestamp,
         factor: u8,
+        current_time: Timestamp,
         store: DynOpStore,
     ) -> K2Result<Self> {
-        let mut pt = Self::new(origin_timestamp, factor)?;
+        let mut pt = Self::new(factor)?;
 
         pt.full_slices = store.slice_hash_count().await?;
 
@@ -149,7 +147,7 @@ impl PartitionedTime {
         let full_slice_end_timestamp = pt.full_slice_end_timestamp();
 
         // Given the time reserved by full slices, how much time is left to partition into smaller slices
-        let recent_time = (Timestamp::now() - full_slice_end_timestamp).map_err(|_| {
+        let recent_time = (current_time - full_slice_end_timestamp).map_err(|_| {
             K2Error::other("Failed to calculate recent time, either the clock is wrong or this is a bug")
         })?;
 
@@ -158,7 +156,7 @@ impl PartitionedTime {
         }
 
         // Update the state for the current time. The stored slices might be out of date.
-        pt.update(store, Timestamp::now()).await?;
+        pt.update(current_time, store).await?;
 
         Ok(pt)
     }
@@ -183,8 +181,8 @@ impl PartitionedTime {
     ///   - Update the `next_update_at` field to the next time an update is required.
     pub async fn update(
         &mut self,
-        store: DynOpStore,
         current_time: Timestamp,
+        store: DynOpStore,
     ) -> K2Result<()> {
         // Check if there is enough time to allocate new full slices
         self.update_full_slice_hashes(store.clone(), current_time)
@@ -210,9 +208,8 @@ impl PartitionedTime {
     ///
     /// This constructor just creates an instance with initial values, but it doesn't update the
     /// state with full and partial slices for the current time.
-    fn new(origin_timestamp: Timestamp, factor: u8) -> K2Result<Self> {
+    fn new(factor: u8) -> K2Result<Self> {
         Ok(Self {
-            origin_timestamp,
             factor,
             full_slices: 0,
             partial_slices: Vec::new(),
@@ -231,12 +228,12 @@ impl PartitionedTime {
 
     /// The timestamp at which the last full slice ends.
     ///
-    /// If there are no full slices, then this is the origin timestamp.
+    /// If there are no full slices, then this is the constant [UNIX_TIMESTAMP].
     fn full_slice_end_timestamp(&self) -> Timestamp {
         let full_slices_duration = Duration::from_secs(
             self.full_slices * self.full_slice_duration.as_secs(),
         );
-        self.origin_timestamp + full_slices_duration
+        UNIX_TIMESTAMP + full_slices_duration
     }
 
     /// Figure out how many new full slices need to be allocated.
@@ -290,9 +287,7 @@ impl PartitionedTime {
 
             let hash = combine_op_hashes(op_hashes);
 
-            store
-                .store_slice_hash(self.full_slices, hash.to_vec())
-                .await?;
+            store.store_slice_hash(self.full_slices, hash).await?;
 
             self.full_slices += 1;
             full_slices_end_timestamp += self.full_slice_duration;
@@ -478,9 +473,8 @@ mod tests {
 
     #[test]
     fn new() {
-        let origin_timestamp = Timestamp::now();
         let factor = 4;
-        let pt = PartitionedTime::new(origin_timestamp, factor).unwrap();
+        let pt = PartitionedTime::new(factor).unwrap();
 
         // Full slices would have size 2^4 = 16, so we should reserve space for at least one
         // of each smaller slice size
@@ -489,13 +483,11 @@ mod tests {
 
     #[tokio::test]
     async fn from_store() {
-        let origin_timestamp = Timestamp::now();
         let factor = 4;
         let store = Arc::new(Kitsune2MemoryOpStore::default());
-        let pt =
-            PartitionedTime::try_from_store(origin_timestamp, factor, store)
-                .await
-                .unwrap();
+        let pt = PartitionedTime::try_from_store(factor, UNIX_TIMESTAMP, store)
+            .await
+            .unwrap();
 
         assert_eq!(0, pt.full_slices);
         assert!(pt.partial_slices.is_empty());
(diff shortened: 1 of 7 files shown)


crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,530 @@
+use kitsune2_api::{K2Error, K2Result};
+
+/// In-memory store for time slice hashes.
+///
+/// The inner store will look something like this:
+/// ----, -, -, ----, -, ----, -, ----
```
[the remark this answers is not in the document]
ThetaSinner: Yep, that's fair. I did that when I was trying to picture what I was doing. You suggestion is much more readable!

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+/// contains at least 1 hash.
+///
+/// Gaps are not permitted, so the representation starts out as a single block that spans
+/// the entire range of possible slice ids. As hashes are inserted, the block is split around
+/// the inserted hash.
+#[derive(Debug)]
+#[cfg_attr(test, derive(Clone))]
+pub(super) struct TimeSliceHashStore {
+    pub(super) highest_stored_id: Option<u64>,
+    inner: Vec<SliceHash>,
```
neonphog, line 18: If we use a sparse BTreeMap here instead of a Vec, the inserting would be a lot simpler.

[And I don't think the empty block aggregation is too bad](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f5c3d005d7485b3f99f27b2a08ba80ad)

neonphog (changes requested):
Suggestion for an optimization in the test-only memory store, just because I think it simplifies the code. But otherwise, this looks awesome to me!

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
+use kitsune2_api::{K2Error, K2Result};
+
+/// In-memory store for time slice hashes.
+///
+/// The inner store will look something like this:
+/// `[ Block_0_3, Single_3_ABC, Block_4_2, Single_6_DEF ]`
```
mattyg, line 6:
```suggestion
/// `[ Block(0, 3), Single(3, [a, b, c]), Block(4, 2), Single(6, [d, e, f]) ]`
```
still took me a minute to parse this. Is this clearer to others?

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+                SliceHash::Block { count, .. } => *count,
+            }
+    }
+}
+
+impl TimeSliceHashStore {
+    pub(super) fn new() -> Self {
+        Self {
+            highest_stored_id: None,
+            inner: vec![SliceHash::Block {
```
mattyg, line 56: shouldn't this start empty? And then when the first Single is inserted we would also insert a Block before it? Otherwise we have to alter the size of this Block when the first Single is inserted, right?

mattyg: nevermind i see the logic for this below

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+                        break;
+                    }
+                }
+                SliceHash::Block { id, count, .. } => {
+                    if id <= slice_id && slice_id <= id + count {
+                        // Completely contained within a compressed block then either
+                        //   - It is empty, so we can just ignore it.
+                        //   - It is not empty, so we need to split the block into two.
+
+                        if !hash.is_empty() {
```
mattyg, line 113: should this check be moved to the top of the function and always error since there is no case where an empty hash would be used?

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+                        }
+
+                        handled = true;
+                        break;
+                    }
+                }
+            }
+        }
+
+        if !handled {
```
mattyg, line 178: Thinking out loud: If our new hash isn't replacing an existing hash, and isn't being inserting upon an existing block, then shouldn't it get inserted to the end of the list, with a block inserted/extended before it? But I guess that is never the case since we're starting with a single block covering all time. So it seems like this error is unreachable. Is that right?

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+                }
+                None => {
+                    self.highest_stored_id = Some(slice_id);
+                }
+                _ => {}
+            }
+        }
+
+        let mut handled = false;
+        for (i, slice_hash) in self.inner.iter().cloned().enumerate() {
```
mattyg, line 82: Not sure if it would be clearer or more efficient to use [position](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) rather than iterating from the start, or its effectively doing the same thing and you prefer to spell it out

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+
+        // Doesn't change the stored value, but does update the highest stored value
+        assert_eq!(original.inner, store.inner);
+        assert_eq!(None, store.highest_stored_id);
+
+        assert!(store.check());
+    }
+
+    #[test]
+    fn insert_single_hash_into_empty() {
```
mattyg, line 311: maybe one more test `insert_single_hash_into_block` where you explicitly create the first block, even though i know its covered here

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,532 @@
...
+        );
+        assert_eq!(
+            Some(&SliceHash::Block {
+                id: 101,
+                count: u64::MAX - 101
+            }),
+            store.get(u64::MAX)
+        );
+    }
+}
```
mattyg, line 532: do we want tests about how empty hashes get handled?

mattyg (dismissed):
love it

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Hmm, that's a valid operation. If you have received no data in a time window, then when you go to compute the combined hash for it you will get no ops from the host and compute an empty hash. Given that we know it's a no-op though, it's wasteful to call the host and ask them to filter it out.

Cool, I'll make that an error

ThetaSinner: Done: [86b826a](https://github.com/holochain/kitsune2/pull/37/commits/86b826a77b012baffaea258b7fdaa2100aef1eda)

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Assuming the algorithm is correct, this is unreachable yes. However, because the loop isn't guaranteed to do work if there is an integrity problem with the data structure, I like having this here to catch the problem.

That does mean it isn't tested because I can't find a way to hit the condition. I'm happy to remove this is that would be preferred

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: This was a mutable loop at some point and that became something the compiler couldn't reason about. I guess the slight downside with position is needing do do multiple match blocks, but it would simplify the code regardless

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Yes, now tested by `insert_empty_hash_into_empty`

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Done: [89d71f3b](https://github.com/holochain/kitsune2/pull/37/commits/89d71f3b1ff4be3f8e812a0d3ff33f515b7709cd)

ThetaSinner: I went with `rfind`, because most of the time I expect it to be newer time slices being updated, so it probably makes sense to search backwards.

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Now that I've used `rfind` below, it's much clearer what this `handled` was for and it's just part of a match condition after the find

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Jeez... yes it would be wouldn't it. I over-complicated this a bunch by trying to actually represent the missing elements. 

What I was thinking was that we want to know which empty spaces have been touched and which haven't. In reality though, that information is not that useful. We want to be able to say to a peer "I have X time slices and this is the combined hash". If we haven't received any data in some time window since the last time slice that did have data then it would be valid for the other peer to check if we're missing something. If we aren't and it's genuinely been quiet, then looking up 0 hashes is cheap... So we can do without tracking it and this becomes really simple.

I just wish I'd thought of this, because as soon as I saw this comment I was wondering what on earth I was thinking :D There was a thought process, but having been through the exercise, this can be way simpler :)

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Never mind, due to the comment above about sparse storage, this is now gone

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Also gone

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: Also gone :)

ThetaSinner:
We don't even need to store the highest seen id, it's just the last element in the BTreeMap

neonphog (approved):
Looks awesome!

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,140 @@
...
+/// response. Otherwise, you will get exactly what was most recently stored for that slice id.
+#[derive(Debug, Default)]
+#[cfg_attr(test, derive(Clone))]
+pub(super) struct TimeSliceHashStore {
+    inner: BTreeMap<u64, bytes::Bytes>,
+}
+
+impl TimeSliceHashStore {
+    /// Insert a hash at the given slice id.
+    pub(super) fn insert(
```
mattyg, line 19: wow that is dramatically simpler haha. So I see that a BTreeMap gives us efficient compression of empty space as its simply not included. But if we have a bunch of sequential hashes, then in theory we don't need the index of any but the first. Do you know if serializing a BTreeMap takes this into account, or is it worth writing a custom serializer? Or is this an unnecessary premature optimization?

mattyg (approved):

crates/dht/src/time.rs
```diff
@@ -489,67 +485,55 @@ mod tests {
...
         // Should allocate no full slices and one partial
         assert_eq!(0, pt.full_slices);
         assert_eq!(1, pt.partial_slices.len());
 
         // The partial should be:
         //   - The minimum factor size, and
-        //   - start at the origin time
-        //   - end at the origin time + UNIT_TIME
+        //   - start at the UNIX_TIMESTAMP
+        //   - end at the UNIX_TIMESTAMP+ UNIT_TIME
```
jost-s, line 515:
```suggestion
        //   - end at the UNIX_TIMESTAMP + UNIT_TIME
```

crates/memory/src/op_store/time_slice_hash_store.rs
```diff
@@ -0,0 +1,140 @@
...
+    #[test]
+    fn create_empty() {
+        let store = TimeSliceHashStore::default();
+
+        assert_eq!(None, store.highest_stored_id());
+        assert!(store.inner.is_empty());
+    }
+
+    #[test]
+    #[should_panic(expected = "Cannot insert empty combined hash")]
```
jost-s, line 60: Seems like this doesn't panic but returns an error.

crates/dht/src/time.rs
```diff
@@ -1075,8 +1037,58 @@ mod tests {
...
+            }])
+            .await
+            .unwrap();
+        // and compute the new state in the future
+        pt.update(Timestamp::now() + 2 * pt.full_slice_duration, store.clone())
+            .await
+            .unwrap();
+
+        // Then a single full slice should have been created
+        assert!(store.slice_hash_count().await.unwrap() > 15_000);
```
jost-s, line 1087: Store the initial slice hash count at the beginning of the test and then add one?
```suggestion
        assert_eq!(store.slice_hash_count().await.unwrap(),  initial_slice_hash_count + 1);
```

jost-s (approved):
Yay!

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: I expect that sequences of non-empty hashes that are exactly equal will be incredibly rare. I don't think we need to worry about compressing those. I'd tried to deal with compressing empty hashes but would have been better off with this sparse representation that only includes hashes with some real content.

Even if you published the same entry in a Holochain app once a day and there was no other DHT activity, it'd still be at a different timestamp and so the op hash would be different. Just occasionally the XOR of a bunch of hashes will match, but not often enough that I think it's important to compress

mattyg: No sorry I don't mean sequential hashes that are equal, I mean that I assume the BTreeMap is always storing the index for every hash, even if the indexes are sequential.

I.e.  [..., (index, hash), (index+1, hash2), (index+2, hash3), ...] 

But because they're sequential you don't actually need index+1 and index+2, so I'm wondering if BTreeMap takes this into account in serialization or we should for a little more compression, or if you think it low value premature optimization.

ThetaSinner: Oooh, I get what you're saying. When we do have a block of hashes because we've had data in each time slice.

Yes I think you could definitely write a data structure that does that. I don't think BTreeMap would.

However, this is really just a test implementation. We'll need a persistent version of this in SQLite and then we'd struggle to implement the same optimization. I'm quite happy to have the test implementation less than perfect on memory usage but really similar to the real implementation

crates/memory/src/op_store.rs
```diff
@@ -53,33 +56,59 @@ impl OpStore for Kitsune2MemoryOpStore {
...
         .boxed()
     }
 
+    /// Retrieve the count of time slice hashes stored.
+    ///
+    /// Note that this is not the number of hashes that have been provided at unique `slice_id`s.
+    /// Start from time slice id 0 and count up to the highest stored id. This value should be the
+    /// count based on the highest stored id.
+    /// This value is easier to compare between peers because it tracks an absolute number of time
+    /// slices that have had a slice hash computed and stored. The number of stored hashes would
```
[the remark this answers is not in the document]
ThetaSinner: Hmm, yeah this isn't clear and looking at it again, it's not entirely true... Because we'll sync from time 0 but also be doing "latest ops" from peers and we'll end up computing slices at both ends potentially during a catch up...

Let me have another think about this

crates/memory/src/op_store/time_slice_hash_store.rs
ThetaSinner: I can write it as a test that errors

crates/memory/src/op_store/time_slice_hash_store.rs
mattyg: Gotcha makes sense. I'm mostly interested in optimizing the data sent over network rather than memory/disk usage.

crates/memory/src/op_store.rs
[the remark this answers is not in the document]
ThetaSinner: I've had a go at justifying what I was thinking here. Please take a look at the updated documentation and if it doesn't make more sense then please make me try again, because it might just not make sense :)

crates/memory/src/op_store.rs
```diff
@@ -53,33 +56,63 @@ impl OpStore for Kitsune2MemoryOpStore {
...
 
+    /// Retrieve the count of time slice hashes stored.
+    ///
+    /// Note that this is not the number of hashes that have been provided at unique `slice_id`s.
+    /// Start from time slice id 0 and count up to the highest stored id. This value should be the
+    /// count based on the highest stored id.
+    /// This value is easier to compare between peers because it ignores sync progress. A simple
+    /// count cannot tell the difference between a peer that has synced the first 4 time slices,
+    /// and a peer who has synced the first 3 time slices and created one recent one. However,
+    /// using the highest stored id shows the difference to be 4 and say 300 respectively.
```
jost-s, line 87: I would leave it at that. Here's a suggestion that I find clearer:
```suggestion
    /// Note that this is not the total number of hashes of a time slice at a unique `slice_id`.
    /// This value is the highest stored id, starting from time slice id 0 and counting up to the highest stored id. In other words it is the id of the most recent time slice.
    ///
    /// This value is easier to compare between peers because it ignores sync progress. A simple
    /// count cannot tell the difference between a peer that has synced the first 4 time slices,
    /// and a peer who has synced the first 3 time slices and created one recent one. However,
    /// using the highest stored id shows the difference to be 4 and say 300 respectively.
```

ThetaSinner: Okay, happy with most of that, except it does return a genuine count rather than an the id so I've adjusted that a tiny bit further

crates/dht/src/time.rs
jost-s: Right, so shouldn't this work?
```suggestion
        assert_eq!(store.slice_hash_count().await.unwrap(), initial_pt.full_slices() + 1);
```

ThetaSinner: It does nearly work
- The `full_sices` is calculated on the first update which happens as part of construction. So I just used `pt` here
- It's actually equal and should be. The `slice_hash_count` is meant to report the same value as `full_slices` and that's a valid assertion that I have added
- I do want to keep the original assertion too because that checks that the numbers are sane which checking equality doesn't guarantee

jost-s: Right, definitely keep that assertion.
What the comment before the assertion suggests and what my idea therefore was:
```rust
        let mut pt = PartitionedTime::try_from_store(
            factor,
            current_time,
            store.clone(),
        )
        .await
        .unwrap();
        // Quick calculation of how many full slices we can fit into the current time, leaving space
        // for the minimum recent time.
        // Something like 15k full slices at the time of writing.
        assert_eq!(
            (current_time.as_micros() as u128 - pt.min_recent_time.as_micros())
                / (pt.full_slice_duration.as_micros()),
            pt.full_slices as u128
        );
        let initial_full_slices_count = pt.full_slices;
        ...
        pt.update(Timestamp::now() + 2 * pt.full_slice_duration, store.clone())
            .await
            .unwrap();

        // Then a single full slice should have been created as the current time
        assert!(store.slice_hash_count().await.unwrap() > 15_000);
        assert_eq!(store.slice_hash_count().await.unwrap(), initial_full_slices_count + 1);
```
Does that make sense?

ThetaSinner: Right, yes, I'm being slow. That does make sense. I have to adjust a couple of other bits because I'd added quite a bit of extra time to definitely cover the new op. But I think I now have this set up the way you're after and it does make sense :)

jost-s (approved):
Awesome-o! Thank you for your patience with my insistence.

Issue #26, opened 2024-11-27
[k2] top-level wire protocol spec and encoding implementation


neonphog:
This spec should include:
- the encoding format (messagepack)
- the exact encoding parameters (structs are maps, names are camelCase or whatever we agree on)
- a method for rejecting connections on implementer defined whim (i.e. holochain switch to dht v2, so it's no longer compatible)
- preflight validation (prove we have at least one space in common that we are both allowed to speak on)
- send/recv arbitrary opaque specific module implementation data. I.e. not just "the dht module" we need to know to send it to the actual "dht_v1" module.

PR #38, opened 2024-12-02
Kitsune2 Top-Level P2P Protocol

neonphog:
RESOLVES: https://github.com/holochain/kitsune2/issues/26

Consider before approving:
- Are we cool with protobuf as a departure from msgpack? I like the self-spec-ing nature, but it means ~~pulling in and building the protobuf-src crate~~ having to deal with needing the protoc binary. If we're switching to this, we should probably use it for the module message encoding as well (e.g. gossip messages / fetch queue messages / etc).
  - *sigh* [protobuf-src doesn't work on windows](https://github.com/MaterializeInc/rust-protobuf-native/issues/4) -- I can pretty trivially get protoc installed in windows ci and use that instead, but this is another downside, so let me know what you think before I move forward on that.
  - UPDATE: tried protoc-prebuilt package and it works locally, but the ci runner gets rate-limited from others downloading it... So I've made that an optional feature and am just running a gha to install protoc on ci
- Do we want this in the api crate? We could instead split this out into a separate protocol crate that only the tx module (??) would use?


crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
...
+    // requirements within the data payload of this message type.
+    // For example, if peers are required to include the "dht_v1" module
+    // to communicate with each other, they should reject preflight to
+    // peers that do not include that module.
+    TY_PREFLIGHT = 1;
+
+    // This is a notification or fire-and-forget message from a peer.
+    //
+    // This type requires that a "space" be specified.
+    TY_NOTIFY = 2;
```
neonphog, line 35: I named this NOTIFY because, especially in the protobuf domain, a "message" is something already. But in my other space PR i've got "incoming_message" and "send_message" which would use this type. So, need some naming help here.

crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
...
+
+    // This is a module communication.
+    //
+    // This type requires that a "space" be specified.
```
cdunster, line 39: Is this line a copy-paste error or does this type require both?

crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
+syntax = "proto3";
+
+package kitsune2.wire;
```
cdunster, line 3: Do we want to version it? I know that buf recommends versioning the packages as such:
```suggestion
package kitsune2.wire.v1;
```

crates/api/build.rs
```diff
@@ -0,0 +1,7 @@
+fn main() -> std::io::Result<()> {
+    std::env::set_var("PROTOC", protobuf_src::protoc());
+    prost_build::Config::new()
+        .bytes(["."])
+        .compile_protos(&["src/wire.proto"], &["src/"])?;
```
cdunster, line 5: I feel that a specific directory for protobuf definitions makes sense, what do you think?

crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
...
+//
+// This is the top-level encoding
+// that will be transferred between Kitsune2 peers. Most communications
+// between peers to make Kitsune2 actually function will be encoded
+// separately inside the payload of TY_MODULE type messages.
+message K2Proto {
+  // Enumeration of the types of messages that can be sent between peers.
+  //
+  // We are using this enum field to distinguish between top-level messages,
+  // rather than protobuf's oneof because of the downsides of upgrading oneofs.
```
cdunster, line 15: What are the downsides of `oneof`'s?

crates/api/src/wire.proto
cdunster: Based on the fact that they are "fire-and-forget" then I think that notification makes sense to me so how about just:
```suggestion
    TY_NOTIFICATION = 2;
```

cdunster:
@neonphog, have you ever worked with [flatbuffers](https://flatbuffers.dev)? I've never used them but I've heard that they're a better and more efficient version of protocol buffers.

Cargo.toml
```diff
@@ -41,6 +43,13 @@ tempfile = "3.14.0"
 thiserror = "2.0.3"
 # this is used by bootstrap_srv as the http server implementation.
 tiny_http = "0.12.0"
+# --- build-dependencies ---
+# The following workspace dependencies are thus-far only used in build-deps.
+# Please be careful to only include them in build dependencies or move them
+# above this section.
+# --- build-dependencies ---
+prost-build = "0.13.3"
+protobuf-src = "2.1.0"
```
ThetaSinner, line 52: This isn't a toolset I know for Rust but does this avoid the windows build issue? https://docs.rs/prost-build/latest/prost_build/

crates/api/src/wire.proto
ThetaSinner: I'm happy with NOTIFY, as long as the naming in method names matches

crates/api/src/wire.proto
ThetaSinner: If I'm understanding this then yes both are required. So you might send
- abc, gossip, please give me your new ops since we last spoke
- abc, fetch, please give me the op data for this list of op hashes

Or maybe we won't ask so nicely, but I think that's the idea here

crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
...
+    //
+    // This type requires that a "space" be specified.
+    // This type requires that a "module" be specified.
+    TY_MODULE = 3;
+
+    // This message indicates a general disconnect, with the reason
+    // or context specified in the data payload.
+    //
+    // We may add additional specific disconnect codes in the future.
+    TY_DISCONNECT_GENERAL = 900;
```
ThetaSinner, line 47: Probably just `TY_DISCONNECT`, if we're giving a reason anyway

crates/api/src/wire.proto
```diff
@@ -0,0 +1,61 @@
...
+  // The type of this message.
+  optional Ty ty = 1;
+
+  // The payload or content of this message.
+  optional bytes data = 2;
+
+  // If the Ty requires that a space be specified, this is it.
+  optional bytes space = 3;
+
+  // If the Ty requires a module impl be specified, this is it.
```
ThetaSinner, line 59: Might be good to mention some modules as examples, so it's clear what this field is for? Module is a fairly general term but we have a specific meaning in mind here

ThetaSinner:
Protobuf is good with me, I think it's easier to keep track of breaking changes because protobuf provide good guidance for how to think about that and this doesn't affect app developers anyway.

I can't think why somebody would want to integrate with Kitsune at the wire level but not want to use any of our types. I think that would only happen if they weren't using Rust and then what they'll actually want is the `.proto` files. I'm not sure how we best package those if somebody wants them but we can figure that out if somebody asks for it. 
Then for a program that is embedding Kitsune, it's going to have to build the protos however we split the libraries. So I'm not seeing a strong case for separating those out up front.

crates/api/src/wire.proto
neonphog: [Mostly that they are very fiddly to update if you can at all](https://protobuf.dev/programming-guides/proto3/#backward)

Cargo.toml
neonphog: We are indeed using prost-build in the build.rs. It in turn requires the protoc protobuf compiler binary to be on the path, which is where the windows issue arises. I'm going to try [protoc-prebuilt](https://crates.io/crates/protoc-prebuilt) as a test to see if it works better. It should, at the very least, be faster, since it won't have to compile it : )

neonphog:
@cdunster 

> have you ever worked with [flatbuffers](https://flatbuffers.dev/)? I've never used them but I've heard that they're a better and more efficient version of protocol buffers.

I have not, although I'd heard the same. My sense is that protobuf has the first mover advantage in popularity and tooling/rust integration. And I'm not too worried about the small amount of protocol overhead as the majority of the data we'll be transferring is the op content. But I'm totally up for experimenting with it if you think it is worth it!

crates/api/build.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/38/commits/c7c3632db3f838143cf3b0a1ea4583f0ffb04bb2

Cargo.toml
neonphog: That seems to have worked: https://github.com/holochain/kitsune2/pull/38/commits/bbeca4e9bbe256744a04379cec3fce6c2f1074e7 -- https://github.com/holochain/kitsune2/actions/runs/12143891872/job/33861875174?pr=38

crates/api/src/wire.proto
cdunster: It's just a [recommendation from buf](https://buf.build/docs/lint/rules/#package_version_suffix). I understand the same as you for the design of protobuf but I've only ever used them in production with https://buf.build. I'm not sure if I'd recommend it though but it's just what I'm familiar with so thought I'd ask :-)

cdunster:
@neonphog 

> I have not, although I'd heard the same. My sense is that protobuf has the first mover advantage in popularity and tooling/rust integration. And I'm not too worried about the small amount of protocol overhead as the majority of the data we'll be transferring is the op content. But I'm totally up for experimenting with it if you think it is worth it!

I don't know enough about flatbuffers to risk convincing you to use them :wink: all I know is that I trust the person who told me they are easier to work with and that it would avoid a lot of the boilerplate we added to support things like backwards compatibility.

crates/api/src/wire.proto
neonphog: I'd submit it's [possible](https://www.hmeid.com/blog/just-say-no-to-versioning) to add functionality without breaking previous clients / versioning the protocol.

crates/api/src/wire.proto
neonphog: Done: https://github.com/holochain/kitsune2/pull/38/commits/522892f79d9840fd718edc066bb69b8343a20422

Cargo.toml
neonphog: Nevermind, that was flaky because the runner is sometimes getting ratelimited for the binary download. Switched it to being an optional feature flag, and defaulted to having ci install protoc for use: https://github.com/holochain/kitsune2/pull/38/commits/208fec56f52462139df53d97fe1f5dd53c231193

ThetaSinner: I definitely saw that... :) Cool okay, well this has always been my sticking point with Avro and Protobuf. They're great to use but the tooling for them can be poor.

We have to consider the downstream for this. We'll need to provide this in all our CI that builds the Kitsune API, which will be most things that interact with Holochain. I've checked it's available for Nix, and it is so that's not an issue. We'll also introduce this as a build dependency on Windows for people who compile from source.

I'd prefer to use Protobuf if it could be done without external tools everywhere, but I'm wondering the value of msgpack is just that it doesn't need external build tools to work?

Does protobuf give us a better wire format? 
I believe it gives clearer way to evolve the format
Anything else that we should consider?

neonphog: @ThetaSinner - Another option would be to build a developer generator command that would build the rust module and we could check it into source control and not require the protoc binary at compile time.

neonphog: This works: https://github.com/holochain/kitsune2/pull/38/commits/dc8ca9ba37a84aea6025c66b0bd552b64ac31dd8 we'll just have to remember to run that binary any time we update the .proto file.

jost-s (commented):
Top-level protocol, does this mean that inside of this protobuf message encoding there'll be MessagePack encoded messages that are passed onto Holochain?

Cargo.toml
ThetaSinner: I like that! We can build some automated checks around that later too, like generate then check for a diff. Great, I'm happy then :)

ThetaSinner (approved):

neonphog:
@jost-s 

> Top-level protocol, does this mean that inside of this protobuf message encoding there'll be MessagePack encoded messages that are passed onto Holochain?

Setting asside holochain for the moment, we indeed *can* do anything for the sub-encoding. Messagepack, json, cbor, whatever. If we're picking protobuf for the top-level, though, it probably makes the most sense to use it also for our internal module protocols.

Now, on to holochain. The op encoding will be entirely defined by holochain. I don't think we want to change that yet, so that will remain messagepack. But that is opaque to kitsune and would be included in a sub-sub-encoding of the modules' protocols.

crates/api/Cargo.toml
```diff
@@ -13,10 +13,21 @@ edition = "2021"
...
+prost-build = { workspace = true }
+protoc-prebuilt = { workspace = true, optional = true }
+
 [dev-dependencies]
 tokio = { workspace = true, features = ["full"] }
+
+[features]
+default = []
+
+# fetch a prebuilt/vendored protoc for use in builting the protocol bindings
```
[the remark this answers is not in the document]
neonphog: This has been deleted : )

jost-s:
@neonphog 

> Setting asside holochain for the moment, we indeed can do anything for the sub-encoding. Messagepack, json, cbor, whatever. If we're picking protobuf for the top-level, though, it probably makes the most sense to use it also for our internal module protocols.

So there are 3 levels of encoding. Is the  `data` of a top-level message the 2nd layer?

neonphog:
@jost-s 

> So there are 3 levels of encoding. Is the data (= payload/content) of a top-level message the 2nd layer?

Yep : )

jost-s (approved):
Fine with it all =)

Issue #21, opened 2024-11-27
[k2] top-level module (the kitsune2 module)


neonphog:
This top level module will instantiate the tx module and the individual space modules. It contains the external api that implementers like holochain will make use of.

PR #40, opened 2024-12-03
First-iteration top-level kitsune module

neonphog:
RESOLVES: https://github.com/holochain/kitsune2/issues/21

### DIFF
--- a/crates/api/src/builder.rs
+++ b/crates/api/src/builder.rs
@@ -15,6 +15,10 @@ pub struct Builder {
     /// The [agent::Verifier] to use for this Kitsune2 instance.
     pub verifier: agent::DynVerifier,
 
+    /// The [kitsune::KitsuneFactory] to be used for creating
+    /// [kitsune::Kitsune] kitsune-level kitsune2 module instances.
+    pub kitsune: kitsune::DynKitsuneFactory,
+
     /// The [space::SpaceFactory] to be used for creating
     /// [space::Space] instances.
     pub space: space::DynSpaceFactory,
@@ -32,10 +36,12 @@ impl Builder {
         let Self {
             config,
             verifier: _,
+            kitsune,
             space,
             peer_store,
         } = self;
 
+        kitsune.default_config(config)?;
         space.default_config(config)?;
         peer_store.default_config(config)?;
 
--- a/crates/api/src/builder.rs
+++ b/crates/api/src/builder.rs
@@ -49,10 +49,11 @@ impl Builder {
     }
 
     /// This will generate an actual kitsune instance.
-    // TODO - the result type of this build function is temporarilly
-    //        an Arc of the builder itself. Once we have the Kitsune
-    //        factory, this will produce an actual Kitsune instance.
-    pub fn build(self) -> Arc<Self> {
-        Arc::new(self)
+    pub async fn build(
+        self,
+        handler: kitsune::DynKitsuneHandler,
+    ) -> K2Result<kitsune::DynKitsune> {
+        let builder = Arc::new(self);
+        builder.kitsune.create(builder.clone(), handler).await
     }
 }
--- a/crates/api/src/kitsune.rs
+++ b/crates/api/src/kitsune.rs
@@ -1,0 +1,55 @@
+//! Kitsune2 top-level module related types.
+
+use crate::*;
+use std::sync::Arc;
+
+/// Handler for events coming out of Kitsune2.
+pub trait KitsuneHandler: 'static + Send + Sync + std::fmt::Debug {
+    /// Gather preflight data to send to a new opening connection.
+    /// Returning an Err result will close this connection.
+    fn preflight_gather(&self, peer_url: String) -> K2Result<bytes::Bytes>;
+
+    /// Validate preflight data sent by a remote peer on a new connection.
+    /// Returning an Err result will close this connection.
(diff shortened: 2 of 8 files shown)


crates/api/src/builder.rs
```diff
@@ -15,6 +15,10 @@ pub struct Builder {
     /// The [agent::Verifier] to use for this Kitsune2 instance.
     pub verifier: agent::DynVerifier,
 
+    /// The [kitsune::KitsuneFactory] to be used for creating
+    /// [kitsune::Kitsune] kitsune-level kitsune2 module instances.
```
jost-s, line 19: kitsune kitsune kitsune
```suggestion
    /// [kitsune::Kitsune] module instances.
```

crates/api/src/kitsune.rs
```diff
@@ -0,0 +1,55 @@
+//! Kitsune2 top-level module related types.
+
+use crate::*;
+use std::sync::Arc;
+
+/// Handler for events coming out of Kitsune2.
+pub trait KitsuneHandler: 'static + Send + Sync + std::fmt::Debug {
+    /// Gather preflight data to send to a new opening connection.
+    /// Returning an Err result will close this connection.
+    fn preflight_gather(&self, peer_url: String) -> K2Result<bytes::Bytes>;
```
jost-s, line 10: What do you think of including `incoming`/`outgoing` in these function names, e.g. `preflight_outgoing_gather`?

crates/core/src/factories/core_kitsune.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            // This is quick, we don't hold the lock very long,
+            // because we're just constructing the future here,
+            // not awaiting it.
+            let fut = match self.map.lock().unwrap().entry(space.clone()) {
+                Entry::Occupied(e) => e.get().clone(),
+                Entry::Vacant(e) => {
+                    let builder = self.builder.clone();
+                    let handler = self.handler.clone();
+                    e.insert(futures::future::FutureExt::shared(Box::pin(
```
jost-s, line 91: Nothing else done with this future/map yet, right?

neonphog: Not sure anything else ever needs to be done with this map. Really space creation shouldn't ever error, and if it does it probably means the whole tokio worker thread died or something. It's just easier to propagate the errors. The shared future is just about race condition management of parallel requests for the same space that doesn't yet exist.

crates/api/src/kitsune.rs
neonphog: Done, and also gave these two default implementations: https://github.com/holochain/kitsune2/pull/40/commits/08d12d04e982d04c9211fb0a12827f440cc8b019

crates/core/src/factories/core_kitsune.rs
jost-s: And the shared future is not yet awaited anywhere, is it?

neonphog: Yep, it's awaited right here to get the result for this function call : )

jost-s: Ah, now I get the comment above. The future is constructed while the map is locked, but it's awaited afterwards when the map lock is released, and awaiting it constructs the space, and the future is shared in case some other thread created the same space in the meantime.

jost-s (dismissed):

crates/api/src/kitsune.rs
```diff
@@ -0,0 +1,70 @@
...
+        peer_url: String,
+        data: bytes::Bytes,
+    ) -> K2Result<()> {
+        drop(peer_url);
+        drop(data);
+        Ok(())
+    }
+
+    /// Kitsune would like to construct a space. Provide a handler.
+    fn create_space(
```
ThetaSinner, line 36: This is a hook so that the user can customize what pieces get used by a space? Wouldn't that be decided by what factories are plugged into already? Or are we thinking that some spaces might be set up differently to others?

This just seems slightly strange to me so I'm wanting to understand where the responsibilities are for setting up a space

neonphog: The space is the level at which messaging between peers is received. So the handler that does the accepting of those messages is the SpaceHandler. This hook at the top-level handler is so that conductor or whatever other implementor can then construct those SpaceHandlers to be able to receive messages from remote peers.

jost-s: Perhaps it could help if you added some integration test stub that shows how this would work with Holochain being the consumer.

neonphog: Done: https://github.com/holochain/kitsune2/pull/40/commits/fa0715ba92c9dd001b8e27629820ad5e392e3f5b

ThetaSinner: This is like the equivalent of the "add a new messaging handle" with the ghost actor in the current implementation, if I've understood this right :)

ThetaSinner (approved):

crates/api/src/kitsune.rs
neonphog: Yeah, exactly, when conductor hooks up the "event" type channel so it can receive events out of legacy kitsune.

jost-s: @neonphog Thanks for the test!

PR #41, opened 2024-12-03
Two way op flow

ThetaSinner:
We discussed needing to treat incoming `MetaOp`s separately from ops that we have persisted. We don't know the internal structure of incoming op data. It needs to go to the host to be decoded. Then within Kitsune, we do need to be able to organize persisted ops by timestamp and op ID. Plus we don't want to be gossiping data until we've been able to store it and will be able to respond to requests for it. 

So, this PR proposes that we have different types and require the host to be able to communicate the op data back to us with its timestamp.

See https://github.com/holochain/kitsune2/pull/20#discussion_r1857349218


crates/memory/src/op_store.rs
```diff
@@ -1,9 +1,28 @@
...
-use std::collections::{BTreeSet, HashMap};
+use kitsune2_api::{
+    K2Error, K2Result, MetaOp, OpId, OpStore, StoredOp, Timestamp,
+};
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, HashMap};
 use tokio::sync::RwLock;
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Kitsune2MemoryOp {
```
neonphog, line 11:
```suggestion
/// This is a stub implementation of an op that will be serialized
/// via serde_json (with inefficient encoding of the payload) to be
/// used for testing purposes.
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct Kitsune2MemoryOp {
```

neonphog (dismissed):

ThetaSinner:
Sorry, I extracted this from other changes and it's not as ready as I thought. I'll get this updated :)

ThetaSinner:
Okay, tests fixed, this should be good to go

crates/dht/src/time.rs
```diff
@@ -742,26 +750,34 @@ mod tests {
...
+                .unwrap(),
+                Kitsune2MemoryOp::new(
+                    OpId::from(bytes::Bytes::from(vec![23; 32])),
+                    origin_timestamp,
+                    vec![],
+                )
+                .try_into()
+                .unwrap(),
+                Kitsune2MemoryOp::new(
+                    OpId::from(bytes::Bytes::from(vec![11; 32])),
```
ThetaSinner, line 768: Note the change from 7 to 11, reusing the same op id isn't a good idea and this test broke with the changed op store

crates/dht/src/time.rs
```diff
@@ -811,26 +827,34 @@ mod tests {
...
+                .unwrap(),
+                Kitsune2MemoryOp::new(
+                    OpId::from(bytes::Bytes::from(vec![23; 32])),
+                    origin_timestamp,
+                    vec![],
+                )
+                .try_into()
+                .unwrap(),
+                Kitsune2MemoryOp::new(
+                    OpId::from(bytes::Bytes::from(vec![7; 32])),
```
neonphog, line 845: Is this one okay to still be duplicated?

neonphog (approved):
:+1:

crates/dht/src/time.rs
ThetaSinner: Apparently yes, the test passe whether I change that or not. That suggests I need less test data in there :)

crates/dht/src/time.rs
jost-s: In which way did it break?

ThetaSinner: The hash comparison later in the test (checking that hashes got XORed together) failed. Because the later 7 overwrote the original

PR #44, opened 2024-12-04
Need a Real Url Type

neonphog:
In starting work on the transport module, I've discovered that we need a real Url type. String isn't gonna cut it anymore : )

### DIFF
--- a/crates/api/src/url.rs
+++ b/crates/api/src/url.rs
@@ -1,0 +1,270 @@
+//! Url-related types.
+
+use crate::*;
+
+// We're using bytes::Bytes as the storage type for urls instead of String,
+// even though it adds a little complexity overhead to the accessor functions
+// here, because Bytes are more cheaply clone-able, and we need it to be bytes
+// for the protobuf wire message types.
+
+/// A validated Kitsune2 Url.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct Url(bytes::Bytes);
+
+impl serde::Serialize for Url {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(self.as_str())
+    }
+}
+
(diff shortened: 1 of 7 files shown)


crates/core/src/factories/mem_peer_store/test.rs
```diff
@@ -10,14 +10,22 @@ const AGENT_1: AgentId = AgentId(Id(bytes::Bytes::from_static(b"agent1")));
 const AGENT_2: AgentId = AgentId(Id(bytes::Bytes::from_static(b"agent2")));
 const SPACE_1: SpaceId = SpaceId(Id(bytes::Bytes::from_static(b"space1")));
 
+fn sneak_url(s: &str) -> Url {
```
ThetaSinner, line 13: Not following these function names?

neonphog: hehe. the test needs to sneak in some debug data to tally up which infos are still around or whatever, so I'm putting that data into the url and extracting it again. I can add some docs there.

neonphog: Done: https://github.com/holochain/kitsune2/pull/44/commits/81dcec59fe12ab9b915e59646055d70df8abd8f5

ThetaSinner (approved):

crates/api/src/url.rs
```diff
@@ -0,0 +1,270 @@
...
+    type Err = K2Error;
+
+    fn from_str(src: &str) -> Result<Self, Self::Err> {
+        Self::from_str(src)
+    }
+}
+
+impl Url {
+    /// Construct a new validated Kitsune2 Url.
+    pub fn new(src: bytes::Bytes) -> K2Result<Self> {
```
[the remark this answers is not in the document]
neonphog: I'd prefer not... but I guess I don't care too much if you'd rather.

jost-s (dismissed):
Nah, that's fine.

ThetaSinner (approved):

PR #45, opened 2024-12-04
Transport API and StubTransport Test Implementation

neonphog:


crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/// Trait-object [TxSpaceHandler].
+pub type DynTxSpaceHandler = Arc<dyn TxSpaceHandler>;
+
+/// Handler for module-related events.
+pub trait TxModuleHandler: TxBaseHandler {
+    /// The sync handler for receiving module messages sent by a remote
+    /// peer in reference to a particular space.
+    fn recv_module(
```
ThetaSinner, line 321: I wonder if this should be `recv_module_msg`. It's the message that gets received and not the module. I'm trying to get used to this but I keep tripping up on the meaning of this

crates/core/src/factories.rs
```diff
@@ -8,3 +8,6 @@ pub use core_space::*;
 
 mod mem_peer_store;
 pub use mem_peer_store::*;
+
+mod stub_transport;
```
ThetaSinner, line 12: Should this be a `mem_transport`? It does do transport work, just not networking.

crates/core/src/factories/stub_transport.rs
```diff
@@ -0,0 +1,341 @@
...
+    }
+}
+
+impl Listener {
+    pub fn url(&self) -> Url {
+        self.url.clone()
+    }
+}
+
+struct Stat {
```
ThetaSinner, line 273: Quick doc comment explaining the structs in here?

crates/core/src/factories/stub_transport.rs
```diff
@@ -0,0 +1,341 @@
...
+            },
+        };
+
+        let send = match self.con_map.lock().unwrap().get(&id) {
+            None => return None,
+            Some(send) => send.clone(),
+        };
+
+        let (ds1, dr1) = tokio::sync::mpsc::unbounded_channel();
+        let (ds2, dr2) = tokio::sync::mpsc::unbounded_channel();
```
ThetaSinner, line 323: Just checking the two channels here? The `listen` set up a channel for the Stat to message us when somebody sends it a message for us. Then we need a way to send message to the stat but also to respond?

crates/core/src/factories/stub_transport/test.rs
```diff
@@ -0,0 +1,348 @@
...
+    let hnd = TxImpHnd::new(hnd);
+    builder
+        .transport
+        .create(builder.clone(), hnd)
+        .await
+        .unwrap()
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn transport_notify() {
```
ThetaSinner, line 186: Nice, this test comes out really clearly with a fixture type extracted :)

crates/core/src/factories/stub_transport/test.rs
```diff
@@ -0,0 +1,348 @@
...
+                }
+            }
+        }
+        panic!(
+            "matching connect entry not found {peer}, {:#?}",
+            self.track.lock().unwrap()
+        );
+    }
+
+    pub fn disconnect(&self, peer: &Url, reason: Option<&str>) -> bool {
```
ThetaSinner, line 128:
```suggestion
    pub fn assert_disconnect(&self, peer: &Url, reason: Option<&str>) -> bool {
```

ThetaSinner (commented):
Looking good! I particularly like the testing with this one and that the traits come with a working and tested implementation so I can see how it all fits together :)

I have left quite a few comments about naming which is quite a pain. They are suggestions only in most cases!

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+            k2_proto::Ty::Module => {
+                if let (Some(space), Some(module)) = (space, module) {
+                    let space = SpaceId::from(space);
+                    if let Some(h) = self
+                        .mod_map
+                        .lock()
+                        .unwrap()
+                        .get(&(space.clone(), module.clone()))
+                    {
+                        h.recv_module(peer, space, module, data);
```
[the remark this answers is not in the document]
neonphog: Yes exactly. Probably not going to happen with modules, but for sure the naive implementation of conductor may use the same space handler for all spaces, so I included the SpaceId there. This is following that pattern.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+                    }
+                }
+                Ok(())
+            }
+            k2_proto::Ty::Disconnect => {
+                let reason = String::from_utf8_lossy(&data).to_string();
+                Err(K2Error::other(format!("Remote Disconnect: {reason}")))
```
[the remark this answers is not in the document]
neonphog: The rest of this code *can* error in the higher level handlers and that always indicates the connection should be closed too. I'm totally up for having a more complex return type, but the functionality would then boil down to "close the connection if you get a Close variant OR if you get an error" so I worry it would just be adding complexity for no reason.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub fn register_module_handler(
+        &self,
+        space: SpaceId,
+        module: String,
+        handler: DynTxModuleHandler,
+    ) {
+        self.mod_map
+            .lock()
+            .unwrap()
+            .insert((space, module), handler);
```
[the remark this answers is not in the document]
neonphog: Oh, yes. Good catch! Are you okay with it panicking in that case?

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+            ty: k2_proto::Ty::Notify as i32,
+            data,
+            space: Some(space.into()),
+            module: None,
+        })
+        .encode()?;
+        self.imp.send(peer, enc).await
+    }
+
+    /// Notify a remote peer module within a space. This is a fire-and-forget
```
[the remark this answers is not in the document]
neonphog: Nope, module messages don't currently have any response tracking in this implementation. Do you think we're going to need rpc-style module messages? If so, we'll have to sort out timeouts for them etc.

crates/api/src/transport.rs
neonphog: haha, fair : )

crates/core/src/factories.rs
neonphog: My thought here was to use "stub" as a distinguisher for module implementations that we don't expect to be production-ready. So when we move your mem-op-store into the core crate, I was going to suggest renaming it to Stub. Totally open to other naming suggestions.

crates/core/src/factories/stub_transport.rs
neonphog: Network connections are bi-directional. Channels are uni-directional. We need two of them so our peer and us can both send and receive messages.

crates/core/src/factories/stub_transport/test.rs
neonphog: This *used* to have an assert in it, but one of my tests needed it to not panic, so it now returns a bool, and doesn't assert internally.

neonphog: Maybe I should actually remove the asserts from the other helpers?

crates/core/src/factories/stub_transport.rs
```diff
@@ -0,0 +1,341 @@
...
+) {
+    let mut con_pool = HashMap::new();
+
+    while let Some(cmd) = cmd_recv.recv().await {
+        match cmd {
+            Cmd::RegCon(url, data_send, mut data_recv) => {
+                match handler.peer_connect(url.clone()) {
+                    Err(_) => continue,
+                    Ok(preflight) => {
+                        let (s, _) = tokio::sync::oneshot::channel();
```
jost-s, line 194: What's the point of this oneshot channel if there's no receiver?

jost-s (commented):
It looks good and is also difficult to comprehend. I need to come back to this tomorrow. It's not as second nature to me as it is to you. ;-)

crates/core/src/factories.rs
ThetaSinner: Understood. My default would probably be to name them as closely as possible to what they do and solve the "where to use it" question with documentation. People should read what a type is for and what its limitations are before using it ideally.

If we want different names so that there aren't a mixture of "memory" things that should and shouldn't be used in production then my thoughts: Stub to me has no depth to it, like a "shuffle" that just returns its input or a "random_bytes" that just returns a zeroed array. This is a valid implementation that should only be used in a single process. So if not "MemoryXXX" then perhaps "IntraProcessXXX" or something like that? I'm not sure about the op store in this case, I don't really fancy "NoPersistenceOpStore"

crates/core/src/factories/stub_transport.rs
ThetaSinner: Cool, yes that makes sense. Thank you

crates/core/src/factories/stub_transport/test.rs
ThetaSinner: I see, yes that's probably a good idea. Maybe something like "check_xxx" instead of "assert_xxx" and have them return what they find for assertion?

crates/api/src/transport.rs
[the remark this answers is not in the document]
ThetaSinner: Yes, this should all happen at startup. We just need to be sensible about not having environment variables or anything else dynamic that can affect the registration of modules unless we're really careful to ensure that we only register one of each regardless of the config.

crates/api/src/transport.rs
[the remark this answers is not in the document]
neonphog: Just because the remote sent us a graceful disconnect message doesn't mean they actually disconnected from us : ) We need to actually disconnect from them so at least *we* are in a consistent state.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    /// Call this when you establish an outgoing connection and
+    /// when you establish an incoming connection. If this call
+    /// returns an error, the connection should be closed immediately.
+    /// On success, this function returns bytes that should be
+    /// sent as a preflight message for additional connection validation.
+    /// (The preflight data should be sent even if it is zero length).
+    pub fn peer_connect(&self, peer: Url) -> K2Result<bytes::Bytes> {
+        for h in self.mod_map.lock().unwrap().values() {
```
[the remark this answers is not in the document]
neonphog: Yeah, exactly. Gossip may want to know a peer is around so it can gossip with it. And some rate-limiting module may want to be able to kill incoming connections from bad peers.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+            h.peer_connect(peer.clone())?;
+        }
+        for h in self.space_map.lock().unwrap().values() {
```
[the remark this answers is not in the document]
neonphog: Exactly, there is only one transport for all of kitsune. I don't know that the spaces have any need of this information, but I liked the symmetry of all of the handler types inheriting from the same core handler trait.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+            ..
+        } = data;
+
+        match ty {
+            k2_proto::Ty::Unspecified => Ok(()),
+            k2_proto::Ty::Preflight => {
+                self.handler.preflight_validate_incoming(peer, data)
+            }
+            k2_proto::Ty::Notify => {
+                if let Some(space) = space {
```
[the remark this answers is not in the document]
neonphog: If somebody's spamming us, hopefully the rate-limiting will catch it. We can add some debug messages in tracing later, but I'm not worrying about tracing much until I need it for debugging or testing.

crates/core/src/factories/stub_transport.rs
neonphog: The command variant requires the sender, we just don't care about awaiting the receiver in this case.

crates/api/src/transport.rs
```diff
@@ -0,0 +1,350 @@
...
+    pub fn new(handler: DynTxHandler) -> Arc<Self> {
+        Arc::new(Self {
+            handler,
+            space_map: Arc::new(Mutex::new(HashMap::new())),
+            mod_map: Arc::new(Mutex::new(HashMap::new())),
+        })
+    }
+
+    /// When constructing a [Transport] from a [TransportFactory],
+    /// This function does the actually wrapping of your implemementation
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/8cad7435fde032eea076b927db291eee3f797f86

crates/api/src/transport.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/f5bf55b6564d64f4e2144fe2ec52e7fdb7f791de

crates/core/src/factories/stub_transport.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/117fa4d9c58a6fcc2bf095c26c8c706c69e031c4

Cargo.toml
```diff
@@ -37,6 +37,9 @@ serde = { version = "1.0.215", features = ["derive"] }
 # kitsune2 agent info is serialized as json to improve debugability of
 # bootstrapping. So, we need a json library.
 serde_json = "1.0.132"
+# We've abandoned the ideal of an async runtime agnostic library
+# in favor of expediency.
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/eba5c8ec2fc164b891e982ed312842dcaa056947

crates/api/src/transport.rs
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/8cad7435fde032eea076b927db291eee3f797f86

crates/core/src/factories.rs
neonphog: That's fair. I spent the requisite 2 minutes trying to think of other names and didn't come up with anything better. Let's go with your original suggestion of just using Mem for this and your op store. We can always change them later if needed.

neonphog: Done: https://github.com/holochain/kitsune2/pull/45/commits/f3fbf7b5e92e13992725de3a3f3b6781787ecc92

crates/core/src/factories/mem_transport/test.rs
```diff
@@ -0,0 +1,368 @@
...
+    )
+    .await
+    .unwrap();
+
+    h2.check_connect(&u1).unwrap();
+
+    t1.disconnect(u2.clone(), Some("test-reason".into())).await;
+
+    h1.check_disconnect(&u2, Some("test-reason")).unwrap();
+    h2.check_disconnect(&u1, Some("test-reason")).unwrap();
```
jost-s, line 300: What happens after a disconnect if you send another space_notify?

jost-s (approved):
I'm starting to see how this is coming together. Which may sound as if this is getting to a stage where I wanted it to be, but it rather means that I'm beginning to understand the full picture.

crates/core/src/factories/mem_transport/test.rs
neonphog: It would reconnect.

crates/core/src/factories.rs
ThetaSinner: I'd genuinely rather debug a memory leak than come up with names for things :) This looks good, thank you!

PR #46, opened 2024-12-05
Add test utils with tracing helper

ThetaSinner:


jost-s (approved):
The tracing subscriber is way under-engineered. Look at Holochain!
I'm just joking.

neonphog (approved):
Nice! :tada:

PR #47, opened 2024-12-05
Dht hash partitioning

ThetaSinner:
Implements hash partitioning in terms of the existing time partitioning.

Things to pay particular attention to:
- The arc changes, looking at naming and whether we like the simple typesdefs vs structured types. E.g. `arc_contains` has to be defined as an associated function rather than a member method.
- The documentation on the hash space. It's a pretty significant deviation from how the hash space was managed in Kitsune2. My reasoning is that having the hash space change its partitioning is extra complexity and endless extra computation for little value. We can vary the argent size, so what really matters is just having small enough chunks that an agent covering one chunk is holding a small enough amount of data.
- All time slices are created up front, even if the combined agent arcs for this node won't need them. This could be made more efficient but is simple for this point in time. I think that will need to change because really numbers of partitions of the hash space take a while to create.

### DIFF
--- a/crates/api/src/agent.rs
+++ b/crates/api/src/agent.rs
@@ -165,8 +165,8 @@ pub struct AgentInfo {
     pub url: Option<Url>,
 
     /// The arc over which this agent claims authority.
-    #[serde(default = "StorageArc::default")]
-    pub storage_arc: StorageArc,
+    #[serde(default = "DhtArc::default")]
+    pub storage_arc: DhtArc,
 }
 
 /// Signed agent information.
@@ -305,7 +305,7 @@ mod test {
         let now = Timestamp::from_micros(1731690797907204);
         let later = Timestamp::from_micros(now.as_micros() + 72_000_000_000);
         let url = Some(Url::from_str("ws://test.com:80/test-url").unwrap());
-        let storage_arc = StorageArc::Arc(42, u32::MAX / 13);
+        let storage_arc = DhtArc::Arc(42, u32::MAX / 13);
 
         let enc = AgentInfoSigned::sign(
             &TestCrypto,
@@ -348,7 +348,7 @@ mod test {
     async fn fills_in_default_fields() {
         let dec = AgentInfoSigned::decode(&TestCrypto, br#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false}","signature":"ZmFrZS1zaWduYXR1cmU"}"#).unwrap();
         assert!(dec.url.is_none());
-        assert_eq!(StorageArc::Empty, dec.storage_arc);
+        assert_eq!(DhtArc::Empty, dec.storage_arc);
     }
 
     #[tokio::test(flavor = "multi_thread")]
--- a/crates/api/src/agent.rs
+++ b/crates/api/src/agent.rs
@@ -316,7 +316,7 @@ mod test {
                 expires_at: later,
                 is_tombstone: false,
                 url: url.clone(),
-                storage_arc: storage_arc.clone(),
+                storage_arc,
             },
         )
         .await
--- a/crates/api/src/agent.rs
+++ b/crates/api/src/agent.rs
@@ -115,31 +115,6 @@ pub trait LocalAgent: Signer + 'static + Send + Sync + std::fmt::Debug {
 /// Trait-object [LocalAgent].
 pub type DynLocalAgent = Arc<dyn LocalAgent>;
 
-/// A basic definition of a storage arc compatible with the concept of
-/// storage and querying of items in a store that fall within that arc.
-///
-/// This is intentionally a type definition and NOT a struct to prevent
-/// the accumulation of functionality attached to it. This is intended
-/// to transmit the raw concept of the arc, and ensure that any complexity
-/// of its usage are hidden in the modules that need to use this raw data,
-/// e.g. any store or gossip modules.
-///
-/// - If None, this arc does not claim any coverage.
-/// - If Some, this arc is an inclusive range from the first loc to the second.
-/// - If the first bound is larger than the second, the claim wraps around
-///   the end of u32::MAX to the other side.
-/// - A full arc is represented by `Some((0, u32::MAX))`.
-pub type BasicArc = Option<(u32, u32)>;
-
-/// An empty basic arc (`None`) is used for tombstone entries and for
-/// light-weight nodes that cannot afford the storage and bandwidth of being
-/// an authority.
-pub const BASIC_ARC_EMPTY: BasicArc = None;
-
-/// A full basic arc (`Some((0, u32::MAX))`) is used by nodes that wish to
-/// claim authority over the full DHT.
-pub const BASIC_ARC_FULL: BasicArc = Some((0, u32::MAX));
-
 mod serde_string_timestamp {
     pub fn serialize<S>(
         t: &crate::Timestamp,
--- a/crates/api/src/arc.rs
+++ b/crates/api/src/arc.rs
@@ -193,6 +193,15 @@ mod tests {
     }
 
     #[test]
+    fn contains_empty_arc_no_locations() {
+        let arc = DhtArc::Empty;
+
+        assert!(!arc.contains(0));
+        assert!(!arc.contains(u32::MAX));
+        assert!(!arc.contains(u32::MAX / 2));
+    }
+
+    #[test]
     fn arc_dist_edge_cases() {
         type Dist = u32;
         type Loc = u32;
--- a/crates/api/src/arc.rs
+++ b/crates/api/src/arc.rs
@@ -1,7 +1,6 @@
 //! Kitsune2 arcs represent a range of hash locations on the DHT.
 //!
-//! An arc can exist in its own right and refer to a range of locations on the DHT.
-//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! When used in the context of an agent, it represents the range of hash locations
 //! that agent is responsible for.
 
 use serde::{Deserialize, Serialize};
@@ -10,7 +9,7 @@ use serde::{Deserialize, Serialize};
 /// storage and querying of items in a store that fall within that arc.
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Default)]
 #[serde(untagged)]
-pub enum StorageArc {
+pub enum DhtArc {
     /// No DHT locations are contained within this arc.
     #[default]
     Empty,
@@ -20,9 +19,9 @@ pub enum StorageArc {
     Arc(u32, u32),
 }
 
-impl StorageArc {
+impl DhtArc {
     /// A full arc that contains all DHT locations.
-    pub const FULL: StorageArc = StorageArc::Arc(0, u32::MAX);
+    pub const FULL: DhtArc = DhtArc::Arc(0, u32::MAX);
 
     /// Get the min distance from a location to an arc in a wrapping u32 space.
     /// This function will only return 0 if the location is covered by the arc.
@@ -62,8 +61,8 @@ impl StorageArc {
     /// ```
     pub fn dist(&self, loc: u32) -> u32 {
         match self {
-            StorageArc::Empty => u32::MAX,
-            StorageArc::Arc(arc_start, arc_end) => {
+            DhtArc::Empty => u32::MAX,
+            DhtArc::Arc(arc_start, arc_end) => {
                 let (d1, d2) = if arc_start > arc_end {
                     // this arc wraps around the end of u32::MAX
 
@@ -134,12 +133,12 @@ impl StorageArc {
     ///
     /// |---b--a-A--B---|
     /// ```
-    pub fn overlaps(&self, other: &StorageArc) -> bool {
+    pub fn overlaps(&self, other: &DhtArc) -> bool {
         match (&self, &other) {
-            (StorageArc::Empty, _) | (_, StorageArc::Empty) => false,
+            (DhtArc::Empty, _) | (_, DhtArc::Empty) => false,
             (
-                this @ StorageArc::Arc(a_beg, a_end),
-                other @ StorageArc::Arc(b_beg, b_end),
+                this @ DhtArc::Arc(a_beg, a_end),
+                other @ DhtArc::Arc(b_beg, b_end),
             ) => {
                 // The only way for there to be overlap is if
                 // either of a's start or end points are within b
@@ -155,11 +154,11 @@ impl StorageArc {
 
 #[cfg(test)]
 mod tests {
-    use crate::StorageArc;
+    use crate::DhtArc;
 
     #[test]
     fn contains_full_arc_all_values() {
-        let arc = StorageArc::FULL;
+        let arc = DhtArc::FULL;
 
         // Contains bounds
         assert!(arc.contains(0));
@@ -171,7 +170,7 @@ mod tests {
 
     #[test]
     fn contains_includes_bounds() {
-        let arc = StorageArc::Arc(32, 64);
+        let arc = DhtArc::Arc(32, 64);
 
         assert!(arc.contains(32));
         assert!(arc.contains(64));
@@ -179,7 +178,7 @@ mod tests {
 
     #[test]
     fn contains_wraps_around() {
-        let arc = StorageArc::Arc(u32::MAX - 32, 32);
+        let arc = DhtArc::Arc(u32::MAX - 32, 32);
 
         assert!(!arc.contains(u32::MAX - 33));
         assert!(arc.contains(u32::MAX - 32));
@@ -195,37 +194,37 @@ mod tests {
     fn arc_dist_edge_cases() {
         type Dist = u32;
         type Loc = u32;
-        const F: &[(Dist, Loc, StorageArc)] = &[
+        const F: &[(Dist, Loc, DhtArc)] = &[
             // Empty arcs contain no values, distance is always u32::MAX
-            (u32::MAX, 0, StorageArc::Empty),
-            (u32::MAX, u32::MAX / 2, StorageArc::Empty),
-            (u32::MAX, u32::MAX, StorageArc::Empty),
+            (u32::MAX, 0, DhtArc::Empty),
+            (u32::MAX, u32::MAX / 2, DhtArc::Empty),
+            (u32::MAX, u32::MAX, DhtArc::Empty),
             // Unit length arcs at max value
-            (1, 0, StorageArc::Arc(u32::MAX, u32::MAX)),
+            (1, 0, DhtArc::Arc(u32::MAX, u32::MAX)),
             (
                 u32::MAX / 2 + 1,
                 u32::MAX / 2,
-                StorageArc::Arc(u32::MAX, u32::MAX),
+                DhtArc::Arc(u32::MAX, u32::MAX),
             ),
             // Unit length arcs at min value
-            (1, u32::MAX, StorageArc::Arc(0, 0)),
-            (u32::MAX / 2, u32::MAX / 2, StorageArc::Arc(0, 0)),
+            (1, u32::MAX, DhtArc::Arc(0, 0)),
+            (u32::MAX / 2, u32::MAX / 2, DhtArc::Arc(0, 0)),
             // Lower bound is inclusive
-            (0, 0, StorageArc::Arc(0, 1)),
-            (0, u32::MAX - 1, StorageArc::Arc(u32::MAX - 1, u32::MAX)),
+            (0, 0, DhtArc::Arc(0, 1)),
+            (0, u32::MAX - 1, DhtArc::Arc(u32::MAX - 1, u32::MAX)),
             // Distance from lower bound, non-wrapping
-            (1, 0, StorageArc::Arc(1, 2)),
-            (1, u32::MAX, StorageArc::Arc(0, 1)),
+            (1, 0, DhtArc::Arc(1, 2)),
+            (1, u32::MAX, DhtArc::Arc(0, 1)),
             // Distance from upper bound, non-wrapping
-            (1, 0, StorageArc::Arc(u32::MAX - 1, u32::MAX)),
+            (1, 0, DhtArc::Arc(u32::MAX - 1, u32::MAX)),
             // Distance from upper bound, wrapping
-            (0, 0, StorageArc::Arc(u32::MAX, 0)),
-            (1, 1, StorageArc::Arc(u32::MAX, 0)),
+            (0, 0, DhtArc::Arc(u32::MAX, 0)),
+            (1, 1, DhtArc::Arc(u32::MAX, 0)),
             // Distance from lower bound, wrapping
-            (1, u32::MAX - 1, StorageArc::Arc(u32::MAX, 0)),
-            (1, u32::MAX - 1, StorageArc::Arc(u32::MAX, 1)),
+            (1, u32::MAX - 1, DhtArc::Arc(u32::MAX, 0)),
+            (1, u32::MAX - 1, DhtArc::Arc(u32::MAX, 1)),
             // Contains, wrapping
-            (0, 0, StorageArc::Arc(u32::MAX, 1)),
+            (0, 0, DhtArc::Arc(u32::MAX, 1)),
         ];
 
         for (dist, loc, arc) in F.iter() {
@@ -242,54 +241,34 @@ mod tests {
     #[test]
     fn arcs_overlap_edge_cases() {
         type DoOverlap = bool;
-        const F: &[(DoOverlap, StorageArc, StorageArc)] = &[
-            (false, StorageArc::Arc(0, 0), StorageArc::Arc(1, 1)),
-            (
-                false,
-                StorageArc::Arc(0, 0),
-                StorageArc::Arc(u32::MAX, u32::MAX),
-            ),
-            (true, StorageArc::Arc(0, 0), StorageArc::Arc(0, 0)),
+        const F: &[(DoOverlap, DhtArc, DhtArc)] = &[
+            (false, DhtArc::Arc(0, 0), DhtArc::Arc(1, 1)),
+            (false, DhtArc::Arc(0, 0), DhtArc::Arc(u32::MAX, u32::MAX)),
+            (true, DhtArc::Arc(0, 0), DhtArc::Arc(0, 0)),
             (
                 true,
-                StorageArc::Arc(u32::MAX, u32::MAX),
-                StorageArc::Arc(u32::MAX, u32::MAX),
+                DhtArc::Arc(u32::MAX, u32::MAX),
+                DhtArc::Arc(u32::MAX, u32::MAX),
             ),
-            (true, StorageArc::Arc(u32::MAX, 0), StorageArc::Arc(0, 0)),
+            (true, DhtArc::Arc(u32::MAX, 0), DhtArc::Arc(0, 0)),
             (
                 true,
-                StorageArc::Arc(u32::MAX, 0),
-                StorageArc::Arc(u32::MAX, u32::MAX),
+                DhtArc::Arc(u32::MAX, 0),
+                DhtArc::Arc(u32::MAX, u32::MAX),
             ),
             (
                 true,
-                StorageArc::Arc(u32::MAX, 0),
-                StorageArc::Arc(u32::MAX, u32::MAX),
+                DhtArc::Arc(u32::MAX, 0),
+                DhtArc::Arc(u32::MAX, u32::MAX),
             ),
-            (true, StorageArc::Arc(0, 3), StorageArc::Arc(1, 2)),
-            (true, StorageArc::Arc(1, 2), StorageArc::Arc(0, 3)),
-            (true, StorageArc::Arc(1, 3), StorageArc::Arc(2, 4)),
-            (true, StorageArc::Arc(2, 4), StorageArc::Arc(1, 3)),
-            (
-                true,
-                StorageArc::Arc(u32::MAX - 1, 1),
-                StorageArc::Arc(u32::MAX, 0),
-            ),
-            (
-                true,
-                StorageArc::Arc(u32::MAX, 0),
-                StorageArc::Arc(u32::MAX - 1, 1),
-            ),
-            (
-                true,
-                StorageArc::Arc(u32::MAX - 1, 0),
-                StorageArc::Arc(u32::MAX, 1),
-            ),
-            (
-                true,
-                StorageArc::Arc(u32::MAX, 1),
-                StorageArc::Arc(u32::MAX - 1, 0),
-            ),
+            (true, DhtArc::Arc(0, 3), DhtArc::Arc(1, 2)),
+            (true, DhtArc::Arc(1, 2), DhtArc::Arc(0, 3)),
+            (true, DhtArc::Arc(1, 3), DhtArc::Arc(2, 4)),
+            (true, DhtArc::Arc(2, 4), DhtArc::Arc(1, 3)),
+            (true, DhtArc::Arc(u32::MAX - 1, 1), DhtArc::Arc(u32::MAX, 0)),
+            (true, DhtArc::Arc(u32::MAX, 0), DhtArc::Arc(u32::MAX - 1, 1)),
+            (true, DhtArc::Arc(u32::MAX - 1, 0), DhtArc::Arc(u32::MAX, 1)),
+            (true, DhtArc::Arc(u32::MAX, 1), DhtArc::Arc(u32::MAX - 1, 0)),
         ];
 
         for (do_overlap, a, b) in F.iter() {
--- a/crates/api/src/arc.rs
+++ b/crates/api/src/arc.rs
@@ -1,0 +1,46 @@
+//! Kitsune2 arcs represent a range of hash locations on the DHT.
+//!
+//! An arc can exist in its own right and refer to a range of locations on the DHT.
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+/// The definition of an arc for Kitsune.
+///
+/// The lower bound is inclusive, the upper bound is exclusive.
+pub type ArcLiteral = (u32, u32);
+
+/// A full arc is represented by `(0, u32::MAX)`.
+pub const ARC_LITERAL_FULL: ArcLiteral = (0, u32::MAX);
+
+/// Check if a location is contained within the arc.
+pub fn arc_contains(arc_literal: ArcLiteral, location: u32) -> bool {
+    if arc_literal.0 < arc_literal.1 {
+        location >= arc_literal.0 && location < arc_literal.1
+    } else {
+        location >= arc_literal.0 || location < arc_literal.1
+    }
+}
+
+/// A basic definition of a storage arc compatible with the concept of
+/// storage and querying of items in a store that fall within that arc.
+///
+/// This is intentionally a type definition and NOT a struct to prevent
+/// the accumulation of functionality attached to it. This is intended
+/// to transmit the raw concept of the arc, and ensure that any complexity
+/// of its usage are hidden in the modules that need to use this raw data,
+/// e.g. any store or gossip modules.
+///
+/// - If None, this arc does not claim any coverage.
+/// - If Some, this arc is an inclusive range from the first loc to the second.
+/// - If the first bound is larger than the second, the claim wraps around
+///   the end of u32::MAX to the other side.
+/// - A full arc is represented by `Some((0, u32::MAX))`.
+pub type BasicArc = Option<ArcLiteral>;
+
+/// An empty basic arc (`None`) is used for tombstone entries and for
+/// light-weight nodes that cannot afford the storage and bandwidth of being
+/// an authority.
+pub const BASIC_ARC_EMPTY: BasicArc = None;
+/// A full basic arc (`Some((0, u32::MAX))`) is used by nodes that wish to
+/// claim authority over the full DHT.
+pub const BASIC_ARC_FULL: BasicArc = Some(ARC_LITERAL_FULL);
--- a/crates/api/src/arc.rs
+++ b/crates/api/src/arc.rs
@@ -17,6 +17,7 @@ pub fn arc_contains(arc_literal: ArcLiteral, location: u32) -> bool {
     if arc_literal.0 < arc_literal.1 {
         location >= arc_literal.0 && location < arc_literal.1
     } else {
+        // Arc wraps around u32::MAX.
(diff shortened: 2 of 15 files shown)


crates/test_utils/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 /// This is intended to be used in tests, so it defaults to DEBUG level.
 pub fn enable_tracing() {
     let _ = tracing_subscriber::fmt()
+        .with_test_writer()
```
ThetaSinner, line 6: I missed this on the first pass, but this is super useful. It is documented to understand libtest's `-- --nocapture` behaviour somehow and makes this much nicer to use in tests. You don't get any log output unless you ask for it

crates/api/src/arc.rs
```diff
@@ -0,0 +1,46 @@
...
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+/// The definition of an arc for Kitsune.
+///
+/// The lower bound is inclusive, the upper bound is exclusive.
+pub type ArcLiteral = (u32, u32);
+
+/// A full arc is represented by `(0, u32::MAX)`.
+pub const ARC_LITERAL_FULL: ArcLiteral = (0, u32::MAX);
```
neonphog, line 13: How can the upper bound be exclusive? The location *at* u32::MAX is not covered by this "full" arc.

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+//! more information about this. Each agent is responsible for storing and serving some part of
+//! the hash space. This module provides a structure that partitions the hash space by some factor.
+//!
+//! The factor must be chosen up front and cannot be changed. The factor determines the number of
+//! partitions to create. A factor of 32 will create a single partition. Each unit decrease from 32
+//! will double the number of partitions. A good default for a network that is expected to grow to
+//! a few thousand nodes is 21. This will create 2^11 partitions, which is 2048. Given that data is
+//! also stored redundantly, say 10 times, this is means that 20,000 nodes are needed for the data
+//! to be maximally distributed. With higher replication, say 50 times, and a hash factor of 20,
+//! over 200,000 nodes are needed for maximal distribution.
```
neonphog, line 13: Not saying no to this. Still digesting. Consider the rest of this message musings:

Part of the hope of u32 was that if an app really took off we'd still be a long way off from every peer having an arc-length of 1 and still covering the whole space. I.e. room to grow.

A subset of happs being written will know they won't ever hit that scale, and may be fine with a smaller bound. In that case, maybe we should use a u16, and not have a factor at all?

Another subset of happs being written will probably never hit that scale, but they'll have the hope of it. In which case, maybe we should go ahead and use a u32 and not have a factor at all, and figure out some way to compress empty locations within a time slice the way we did with empty time slices?

crates/api/src/arc.rs
```diff
@@ -0,0 +1,46 @@
+//! Kitsune2 arcs represent a range of hash locations on the DHT.
+//!
+//! An arc can exist in its own right and refer to a range of locations on the DHT.
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+/// The definition of an arc for Kitsune.
+///
+/// The lower bound is inclusive, the upper bound is exclusive.
+pub type ArcLiteral = (u32, u32);
```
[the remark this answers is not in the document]
neonphog: The range type doesn't handle wrapping around the end...

crates/dht/src/hash.rs
neonphog: As you say, the agents pick the number of chunks they want to cover. That is the scaling metric. The factor here boils down to a choice (presumably at app bundling time) of how much network overhead the developer is willing to accept in exchange for being able to shard more if their app is well used.

u32 was chosen before the time partitioning was a thing... but while the individual op count within any time-bounded arc range will probably be kinda small for most apps, we do still have to consider the bandwidth requirement for full-syncing new nodes into the network.

If an app had a total history burden of 1 Terabyte, if you had an arc size of 1 out of u16, you would only have to store 16 Megabytes of that. So it seems (to me) that we would never need to shard beyond 1:65536.

The same calculation for a u8 yields individual nodes needing to be responsible for 4 Gigabytes. While that doesn't seem entirely unreasonable, it does feel on the edge of maybe there are use-cases that might want to shard beyond that.

Correct me if I'm wrong about the following assumptions:

- We don't care about the network overhead in a full-sync operation. It will be insignificant compared to the op data that needs to be transferred.
- We don't care about the network overhead in a these-nodes-have-last-seen-timestamps of each other, because there is essentially no overhead there, we just send the hashes we've seen since then.
- We DO care about the network overhead in the needing-to-compare-hashes-of-partitions case... especially if there are a lot of partitions--say 65536 per time-slice because the devs are hoping their app will grow crazy. Because with 32-byte hashes, that is already 2M per time slice.

crates/api/src/arc.rs
```diff
@@ -0,0 +1,46 @@
...
+
+/// A full arc is represented by `(0, u32::MAX)`.
+pub const ARC_LITERAL_FULL: ArcLiteral = (0, u32::MAX);
+
+/// Check if a location is contained within the arc.
+pub fn arc_contains(arc_literal: ArcLiteral, location: u32) -> bool {
+    if arc_literal.0 < arc_literal.1 {
+        location >= arc_literal.0 && location < arc_literal.1
+    } else {
+        location >= arc_literal.0 || location < arc_literal.1
```
jost-s, line 20:
```suggestion
        // Arc wraps around u32::MAX.
        location >= arc_literal.0 || location < arc_literal.1
```

crates/api/src/arc.rs
ThetaSinner: That's a mistake. I ran into the top of the range not being covered elsewhere and I think I have this mostly sorted but this is an oversight. I'll take a look but probably focus on the musing below first :)

crates/dht/src/hash.rs
ThetaSinner: I expected this to be a controversial point so musing is good :)

Does full-sync here refer to new agents joining or all agents holding everything? 
I think in the new agent case it's a yes, there would be some overhead to communicate what the top hashes should be to aim for but that should be minimal compared with the data itself.
In the case of agents holding everything, communicating top hashes at a point in time can be combined right the way down to one 32-bit hash as a summary. The problem is when there's a mismatch for some reason, which could quite easily happen if an op gets lost. If we have a really large number of hash partitions and are managing time slices for each then we have to send at least a top level time hash for each hash partition... which is already thousands of hashes at the scale I'm talking about in these docs.
To make that work, I think the network does need to freeze time slices after a while and stop trying to sync them with peers. Ideally a few weeks. At least that way we aren't going to keep going through that loop in gossip of discovering some ancient mismatch and having to repeatedly send lots of hashes in each gossip round.
So what I'm talking about here is the third point I think, about how many partitions we have to compare hashes over.
So that's some first thoughts from me on the topic of "how much sharding should we support". I think once it goes beyond a certain point, full sync nodes are going to be comparing a lot. If you get right down to agents only holding once chunk then it's very cheap. So there is a bit of a tradeoff there depending on how you expect the network to behave. But if you gain enough data quickly for the network to shard, then where are the limits...

Taking something like a 1 Terrabyte network. How much of that do I care about? If I'm storing 4GB but I'm also a participating node that isn't just providing storage. I need to be able to get and cache whatever data I'm interacting with. So say it's something like a social app and I produce something like 200Mb worth of text and images but then I have a couple of hundred friends doing the same. The size of the cache to participate is going to grow rapidly too. Sure I don't have to provide the processing power to validate that data and I can aggressively prune my cache. But I could easily end up with 5-10Gb of data that's in active use if my friends are similarly active and we assume I'm commenting back on threads that people have posted a few weeks ago.

So I'm thinking that in a network that active, there will likely be such a cache requirement to participate that it's acceptable to have a reasonable amount of DHT data to be responsible for. 

Then if we want to hit mega scale. I'm not sure that every participant providing processing power is the right way to go. I think a deployment model where most people have a pinned 0 arc and rely on their cache is going to be the best option. Then run many machines that have 100Gb free for storage. We have to expect some validation to require gets of data that is going to be cached rather than DHT data. So let's be conservative and say that a node with 100Gb can really only offer 60Gb of valuable storage, with aggressive cache cleaning. Then I think that with this level of partitioning the app can actually be scaled to a good level.

All just musings and trying to get across the way I'm thinking about this. I'm in agreement that the same thing would probably work with a u16 and we'd just have smaller numbers to deal with but I'm mostly interested in figuring out whether this works to solve the "how the network scales" questions.

I'm also happy to sit down and do the calculations properly rather than roughly if that would add some value to the discussion. Everything here is pretty much from my head :)

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+//!
+//! Each hash partition manages a [PartitionedTime] structure that is responsible for managing the
+//! time slices for that hash partition. The interface of this module is largely responsible for
+//! delegating the updating of time slices to the inner time partitions. This ensures that all the
+//! time partitions are updated in lockstep. This makes reasoning about the hash-time partitioning
```
jost-s, line 18:
```suggestion
//! time partitions are updated in lockstep. This makes reasoning about the space-time partitioning
```

crates/api/src/arc.rs
ThetaSinner: Changed and added tests

crates/api/src/arc.rs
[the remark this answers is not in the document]
ThetaSinner: Happy to look at range, I also thought it didn't do ends well. But then neither do I apparently :)

crates/dht/src/hash.rs
neonphog: yep sorry, that was a bad use of full-sync. I meant the thing where a new node joins the data and need to catch up to everything for the idea arc size.

crates/api/src/arc.rs
```diff
@@ -0,0 +1,83 @@
...
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+/// The definition of an arc for Kitsune.
+///
+/// The lower bound is inclusive, the upper bound is exclusive.
+pub type ArcLiteral = (u32, u32);
+
+/// A full arc is represented by `(0, u32::MAX)`.
+pub const ARC_LITERAL_FULL: ArcLiteral = (0, 0);
```
neonphog, line 13: Doc is now wrong. Also, I did this with the upper bound being inclusive for the logic in the peer store because that was more intuitive to me. We need to either change that logic or change this logic.

crates/dht/src/hash.rs
neonphog: Okay, I think I've identified what my concern here is. You are correct, the bit width of the location/arc bounds is irrelevant because it just affects local calculations.

Rather than making the factor configurable, I vote for hard-coding it. I'm just going to throw out a number and say 512. I assert that all small and medium scale apps will still ideally hold arc ranges of more than 1 in that 512. Beyond that, a large scale app will function just fine with everyone holding exactly 1 in 512. And you are correct, there are other mitigations that can happen if individual nodes are starting to get overwhelmed in even that scenario. And 512 is small enough that I feel like it will be manageable in terms of coordination overhead, although we should get some numbers on that.

We can always change this factor by protocol version bump. We don't want to make developers make a decision on this that they won't really understand. And if we want to get more sophisticated in later iterations, I think we could probably come up with a way to derive a consensus factor from the data in the dht, or negotiate it in connection preflights even if that just means always picking the higher partition count number.

neonphog: So, having said that, your code here shouldn't need to change at all, so I'll continue my review : )

neonphog:
> The arc changes, looking at naming and whether we like the simple typesdefs vs structured types. E.g. arc_contains has to be defined as an associated function rather than a member method.

So, this again was about my PTSD of the insanity of the Arq stuff in legacy kitsune. I was thinking I don't want to have to see any of that, so if the gossip module needs something more complex, it should do it internally and not expose it, and it can convert whatever struct it is using into a BasicArc when crossing the module boundary into op store or whatever.

We now have the opposite problem where I used two inclusive bounds in the peer store, and you used an inclusive/exclusive bound in the dht logic. Also, the `arc_contains` function *could* be built off my calc_dist function (just check for dist==0).

So, long story short, I think at this point, we should go ahead and make it a struct, and just not let it get out of hand functionality-wise.

crates/api/src/arc.rs
neonphog: also see https://github.com/holochain/kitsune2/pull/47#issuecomment-2524252635

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+            partitioned_hashes.len()
+        );
+
+        Ok(Self {
+            size,
+            partitioned_hashes,
+        })
+    }
+
+    /// Get the next update time the inner time partitions.
```
neonphog, line 102:
```suggestion
    /// Get the next update time of the inner time partitions.
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+use std::collections::HashMap;
+
+/// A partitioned hash structure.
+///
+/// Partitions the hash structure into a configurable number of partitions. Each partition is
+/// responsible for managing the time slices for that partition.
+#[derive(Debug)]
+pub struct PartitionedHashes {
+    size: u32,
+    partitioned_hashes: Vec<HashPartition>,
```
neonphog, line 40:
```suggestion
    /// This is just a convenience for internal function use.
    /// This should always be exactly `((u32::MAX + 1) / self.partitioned_hashes.len()`.
    size: u32,
    /// The partition count here (length of Vec) should always be a power of 2.
    /// (2**0, 2**1, etc).
    partitioned_hashes: Vec<HashPartition>,
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+}
+
+#[derive(Debug)]
+struct HashPartition {
+    _arc: ArcLiteral,
+    partitioned_time: PartitionedTime,
+}
```
neonphog, line 47: What's the point of this struct? Should `PartitionedTime` be directly in the Vec above?

crates/dht/src/hash.rs
neonphog: > To make that work, I think the network does need to freeze time slices after a while and stop trying to sync them with peers. Ideally a few weeks.

By design, there's no consensus in holochain, and by extension kitsune. The same dna hash could be partitioned on two LANs for multiple years, and then come together some day. The nodes will see each other, but if freezing is happening, they won't be able to validate ops properly.

That would certainly be a rare scenario, but it'd be awesome if we could easily support it.

Once we have the hashes in memory, it should be trivial (computationally) to build up a cross-cutting merkle-like tree of them for all arc buckets but a single time-slice. Then (the hope would be) that the older stuff wouldn't change very much, and we could exchange our top-level time slice hashes without much overhead.

No need to do this now, but if you think something like this (or something else entirely) might someday be able to solve this problem, I'd be a lot more comfortable : )

neonphog (changes requested):
Nice work! I've come around to being happy with this : )

Just some outstanding questions about hard-coding the factor, future optimizations, and some details to sort out with my implementation of the arc bounds.

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+//! a few thousand nodes is 21. This will create 2^11 partitions, which is 2048. Given that data is
+//! also stored redundantly, say 10 times, this means that 20,000 nodes are needed for the data
+//! to be maximally distributed. With higher replication, say 50 times, and a hash factor of 20,
+//! over 200,000 nodes are needed for maximal distribution.
+//!
+//! Each hash partition manages a [PartitionedTime] structure that is responsible for managing the
+//! time slices for that hash partition. The interface of this module is largely responsible for
+//! delegating the updating of time slices to the inner time partitions. This ensures that all the
+//! time partitions are updated in lockstep. This makes reasoning about the space-time partitioning
+//! easier.
```
jost-s, line 19: WDYT about using "space partition" for arcs and "time slice" and "time partition" for the time divisions?
```suggestion
//! Each space partition manages a [PartitionedTime] structure that is responsible for managing the
//! time slices for that space partition. The interface of this module is largely responsible for
//! delegating the updating of time slices to the inner time partitions. This ensures that all the
//! time partitions are updated in lockstep. This makes reasoning about the space-time partitioning
//! easier.
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug)]
+struct HashPartition {
+    _arc: ArcLiteral,
+    partitioned_time: PartitionedTime,
+}
+
+impl PartitionedHashes {
+    /// Create a new partitioned hash structure.
+    ///
+    /// The `hash_factor` determines the number of partitions to create. A value of 32 will create
```
jost-s, line 52: Would be the same here, `space_factor` etc.

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+        hash_factor: u8,
+        time_factor: u8,
+        current_time: Timestamp,
+        store: DynOpStore,
+    ) -> K2Result<Self> {
+        if hash_factor > 32 {
+            return Err(K2Error::other("Hash factor must be 32 or lower"));
+        }
+
+        let (size, num_buckets) = if hash_factor == 32 {
```
jost-s, line 68:
```suggestion
        let (size, num_partitions) = if hash_factor == 32 {
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// create the [PartitionedTime] structure.
+    pub async fn try_from_store(
+        hash_factor: u8,
+        time_factor: u8,
+        current_time: Timestamp,
+        store: DynOpStore,
+    ) -> K2Result<Self> {
+        if hash_factor > 32 {
+            return Err(K2Error::other("Hash factor must be 32 or lower"));
+        }
```
jost-s, line 66:
```suggestion
        if hash_factor == 0 || hash_factor > 32 {
            return Err(K2Error::other("Hash factor must be between 1 and 32"));
        }
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+
+        let (size, num_buckets) = if hash_factor == 32 {
+            (u32::MAX, 1)
+        } else {
+            let size = 1u32 << hash_factor;
+
+            // We will always be one bucket short because u32::MAX is not a power of two. It is one less
+            // than a power of two, so the last bucket is always one short.
+            (size, (u32::MAX / size) + 1)
```
jost-s, line 75: This will overflow with hash_factor = 0.
```suggestion
            let compensation = if size == 1 {
                0
            } else {
                1
            }
            (size, (u32::MAX / size) + compensation)
```

crates/api/src/arc.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn arc_wraps_around() {
+        let arc = (u32::MAX - 32, 32);
+
+        assert!(super::arc_contains(arc, u32::MAX - 1));
+        assert!(super::arc_contains(arc, u32::MAX));
+        assert!(super::arc_contains(arc, 20));
+
+        assert!(!super::arc_contains(arc, 32));
+    }
+}
```
mattyg, line 83: maybe a test for BASIC_ARC_EMPTY?

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+                .update(store.clone(), current_time)
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    /// Inform the time partitions of ops that have been stored.
+    ///
+    /// The ops are split into the right hash partition based on the location of the op. Then the
```
jost-s, line 130:
```suggestion
    /// The ops are placed into the right space partition based on the location of the op. Then the
```

crates/dht/src/time.rs
```diff
@@ -200,6 +211,98 @@ impl PartitionedTime {
...
+
+    /// Inform the [PartitionedTime] that some ops have been stored.
+    ///
+    /// It is expected that the caller ensures that the ops belong to the hash range managed by this
+    /// [PartitionedTime]. This method will update the hashes of the full and partial slices that
+    /// the incoming ops belong to.
+    ///
+    /// If the op happens to be new enough to not belong in a slice, then it will be ignored. The
+    /// op will be discovered later when a partial slice update adds a slice that includes the op.
+    pub(crate) async fn inform_ops_stored(
```
jost-s, line 223:
```suggestion
    /// op will be discovered later when a partial slice update adds a slice that includes the op.
```

crates/dht/src/time.rs
```diff
@@ -200,6 +211,98 @@ impl PartitionedTime {
...
+    ) -> K2Result<()> {
+        let full_slice_end = self.full_slice_end_timestamp();
+
+        for op in stored_ops {
+            if op.timestamp < full_slice_end {
+                // This is a historical update. We don't really expect this to happen too often.
+                // If we're syncing because we've been offline then it's okay and we should
+                // try to detect that when it's happening but otherwise it'd be good to log a warning
+                // here.
+
```
jost-s, line 236: So add a `eprintln!` for now?

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+
+        let hash = store.retrieve_slice_hash((0, ph.size), 0).await.unwrap();
+        assert!(hash.is_some());
+        assert_eq!(op_id_bytes_1, hash.unwrap());
+
+        let count = store
+            .slice_hash_count((ph.size, 2 * ph.size))
+            .await
+            .unwrap();
+        // Note that this is because we've stored at id 1, not that two hashes ended up in this
```
jost-s, line 294: Yeah, this naming is confusing. Perhaps it should be `slice_hash_id` instead?

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+        )
+        .await
+        .unwrap();
+
+        let op_id_bytes_1 = bytes::Bytes::from_static(&[100, 0, 0, 0]);
+        let op_id_bytes_2 = bytes::Bytes::from(ph.size.to_le_bytes().to_vec());
+        ph.inform_ops_stored(
+            store.clone(),
+            vec![
+                // Stored in the first partial
```
jost-s, line 325:
```suggestion
                // Stored in the first time slice of the first space partition.
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+                StoredOp {
+                    op_id: OpId::from(op_id_bytes_1.clone()),
+                    timestamp: ph.partitioned_hashes[0]
+                        .partitioned_time
+                        .full_slice_end_timestamp(),
+                },
+                // Stored in the second partial
```
jost-s, line 332:
```suggestion
                // Stored in the second time slice of the first space partition.
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+                    now,
+                    h._arc.1.to_be_bytes().to_vec(),
+                )
+                .try_into()
+                .unwrap()])
+                .await
+                .unwrap();
+        }
+
+        // Check nothing is current stored in the partials
```
jost-s, line 420:
```suggestion
        // Check nothing is currently stored in the partials
```

crates/api/src/arc.rs
```diff
@@ -0,0 +1,46 @@
...
+/// The definition of an arc for Kitsune.
+///
+/// The lower bound is inclusive, the upper bound is exclusive.
+pub type ArcLiteral = (u32, u32);
+
+/// A full arc is represented by `(0, u32::MAX)`.
+pub const ARC_LITERAL_FULL: ArcLiteral = (0, u32::MAX);
+
+/// Check if a location is contained within the arc.
+pub fn arc_contains(arc_literal: ArcLiteral, location: u32) -> bool {
```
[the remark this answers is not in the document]
ThetaSinner: Resolving because of the wraparound problem above

crates/dht/src/hash.rs
ThetaSinner: It is a count, it's the `highest_stored_slice_id + 1`. I believe the naming is correct here and it's the value that's potentially confusing. That's just because we've stored a combined hash at slice_id 1 and that means we consider ourselves to have 2 total slices.

I can try to rewrite the comment to make that clearer?

crates/dht/src/time.rs
ThetaSinner: There is tracing, but there wasn't when I wrote myself a note, honest :) I've added a log message now

crates/dht/src/time.rs
ThetaSinner: Not clear about this diff, possibly on the wrong line?

crates/dht/src/hash.rs
ThetaSinner: You are correct. This will perform poorly long before reaching 0 though, so I'm just going to reject inputs that set the number of partitions too high. Well above 0

crates/dht/src/hash.rs
ThetaSinner: Needs renaming in a few places, done without committing this suggestion :)

crates/dht/src/hash.rs
ThetaSinner: I was thinking it'd be nice to have the `arc` pre-calculated later. I haven't needed it yet so I'll just get rid of this type :)

crates/dht/src/hash.rs
jost-s: Is it really a count? There can be gaps in the vector of stored slice ids for the empty slices, so could it be something like `[_, _, _, _, 4, _, 6]`. The count would be `2` for me. The highest slice id is `6`.

crates/dht/src/time.rs
jost-s: No idea what happened here. I did this in the IDE probably and made a bollocks of it.

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,443 @@
...
+//!
+//! The factor must be chosen up front and cannot be changed. The factor determines the number of
+//! partitions to create. A factor of 32 will create a single partition. Each unit decrease from 32
+//! will double the number of partitions. A good default for a network that is expected to grow to
+//! a few thousand nodes is 21. This will create 2^11 partitions, which is 2048. Given that data is
+//! also stored redundantly, say 10 times, this means that 20,000 nodes are needed for the data
+//! to be maximally distributed. With higher replication, say 50 times, and a hash factor of 20,
+//! over 200,000 nodes are needed for maximal distribution.
+//!
+//! Each hash partition manages a [PartitionedTime] structure that is responsible for managing the
```
jost-s: This still says "hash partition", so I suspect you haven't pushed your changes yet.

crates/dht/src/hash.rs
jost-s: I would hold off on assuming we're going to freeze fixed time slices. We have no data on this, and it may turn out to not be an issue to compare and modify them.

crates/api/src/arc.rs
ThetaSinner: Superseded by arc PR

crates/dht/src/hash.rs
ThetaSinner: Over time it does become an issue I believe. I am fine if we say there's no consensus by design and that means we can't actually refuse to accept updates. However, gossip will degenerate when there is a permanent historical mismatch.

Say we have 1k full time slices. What we'd like to be able to do is aggregate those and expect mismatches only in the partial time slices. We have to expect that in some gossip rounds that isn't what will happen, we're going to run into somebody who we aren't in sync with. In that case, we have to go more granular on the full time slices but we don't want to send more data than we have to, to find the problem. So we fall back to something like telescoping the full time slices into combined hashes. That's not so bad, we're still sending a reasonable amount of data at that point still. We're still relying on probability, where we expect more recent time slices to be more likely to contain a mismatch. So it's okay to roll up more older time slices because we don't think they're so likely to mismatch. Finally though, if we find a mismatch in one of these telescoped slices, we either need to send all hashes in that slice or we need to go round again and send all the top hashes for the time slices that made up that telescoped slice.

That all works fine and won't happen that often. The problem comes when we have a peer who mismatches with us at a low slice and we can't get one or more problem hashes. Then every single gossip round will degenerate down to the worst case and we'll have to send them a large batch of top hashes.

I think that assuming we can always get every op is a mistake. If the system is consensus free then we don't need to match every hash. We just need to do our best to match. At some point, surely it makes sense to protect bandwidth?

I'm fine with treating this as an edge case and not doing anything with it for now. However, I think a potential solution might look like pinning some time hashes for peer where we have a mismatch which is "ancient". Then when we compute a top hash for gossip with that peer, we substitute some full slice hashes of our own, with their value for that time slice. That trades a small amount of storage for a whole lot of bandwidth.
Sure it creates a new problem that a peer can force us to use disk space. But that can be solved too.

crates/api/src/arc.rs
ThetaSinner: Superseded by arc PR

crates/dht/src/hash.rs
ThetaSinner: Hmm, I'm not seeing this any more and the comment is saying "Outdated". Possibly I hadn't pushed, I believe I have got this updated now :)

crates/dht/src/hash.rs
zippy: Can you describe your thinking about how to compute the "top_hash"?

jost-s: > That all works fine and won't happen that often. The problem comes when we have a peer who mismatches with us at a low slice and we can't get one or more problem hashes. Then every single gossip round will degenerate down to the worst case and we'll have to send them a large batch of top hashes.

When we can't get one or more problem combined hashes, we exchange top hashes at the next lower level, right? And if that continues and we don't have any combined hashes we agree on, we start comparing individual op hashes, and since we have the same number of full time slices but don't agree on any combined hashes for any of those slices, we need to compare all op hashes of historical gossip. That is the case you're describing here, right?

I would like to know the probability of this case. It means that another peer has sufficiently gossiped around to know about all the time slices there have been since the beginning of the app, but does not have a single combined hash that matches ours, meaning there's a diff for every of the full slices. Is that what we're talking about?

zippy: Is "top hash" the same as `PartialSlice::hash` in `/dht/src/time.rs`?

jost-s: @zippy yes

zippy: I see, so it's actually the result of `combine_op_hashes` for both partial and full slices.   I notice that this conversation has also included the word "chunk" which appears to by synonymous with "slice" which is what's in the code.  Maybe "chunk" existed in the previous implementation?

ThetaSinner: There's a huge amount of information in this thread now. I'm going to respond about terminology but I think we should split out the conversation about historical mismatches. It's not actually relevant to how we partition, it's relevant to the gossip process that's built on top of it later. For me the link is how granular the partitioning is. The more granular we are with time and space, the more important it is that we don't have to keep re-negotiating historical mismatches. So for this conversation let's just say that we want to keep the number of partitions as low as possible while also allowing Kitsune2 to scale appropriately.

As we iterate, the terms we're using are changing so there's going to have to be some shared adopting/dropping of terms and I hope that we can land on something consistent as this comes together. I'm trying to use the following terms:
- Time slice - Some period of time that is a right-open interval (e.g. May 5, from and including midnight to May 9th, up to but not including midnight)
- Combined hash - The reduce/fold of a set of op hashes, using an XOR over those hashes. When we're talking about a time slice that has no implied "tree" structure, then this is the term I'm using.
- DHT arc - Refers to some part of the DHT hash space. Because of how we're partitioning it can be thought of as a contiguous list of segments which are each the same size. So say we break the hash space into 4 pieces then a valid DHT arc would be 1, 2, 3 or all 4 of those pieces but 2 and 4 wouldn't be possible to represent, it would have to be 1 and 2 or 2 and 3 or 2, 3 and 4.
- Full time slice - A historical time slice that is supposedly complete, bar any network partitioning or offline nodes
- Partial time slice - A recent time slice where we expect regular updates
- Top hash - When there is some implied structure by considering some topology over the space and time. So that could be segments where we roll up full time slices within DHT arcs. It could be a disc where we combine over DHT arcs and full time slices. It could be a ring where we combine over all DHT arcs for a single partial time slice. It is intended to convey the meaning of a "summary of summaries" to compress information that we think is likely to match somebody else's view of the DHT.

ThetaSinner: So `PartialSlice::hash` is not a top hash, it is a "combined hash" as documented and I do want to keep those terms separate. I think "chunk" is confusing and where I've used it then it was an accidental use of terminology from Kitsune 1. I mean holding a single DHT arc, at the defined unit size.

crates/api/src/arc.rs
ThetaSinner: Yes, it's implicitly tested in one of the `dist` tests but not for contains. I've added a new test `contains_empty_arc_no_locations`

crates/dht/src/hash.rs
zippy: Thanks that's really helpful.  It's really helpful if that king of terminology glossary can make it into the spec, or a readme.

crates/dht/src/hash.rs
ThetaSinner: This is now a fixed value

crates/api/src/agent.rs
```diff
@@ -165,8 +165,8 @@ pub struct AgentInfo {
     pub url: Option<Url>,
 
     /// The arc over which this agent claims authority.
-    #[serde(default = "StorageArc::default")]
-    pub storage_arc: StorageArc,
+    #[serde(default = "DhtArc::default")]
```
ThetaSinner, line 168: I missed pushing this rename on the previous PR, my mistake!

crates/dht/src/hash.rs
jost-s: Mmh, something is off, this is the same after refreshing the page.

crates/dht/src/hash.rs
jost-s: I say we keep it and a better term will emerge with usage.

crates/dht/src/hash.rs
ThetaSinner: I see it in this comment view, and if I click the file link at the top of the thread's box then it takes me to the commit hash that the comment was made at and I see it.

I can't find it on the latest version of the code on the review?

ThetaSinner: The latest version is here https://github.com/holochain/kitsune2/pull/47/files#diff-f10db395c584cade20b3f0b8cc59c314af201b415596138055cf7e77a099a6d7R8

jost-s: It's gone now.

jost-s (approved):

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,440 @@
...
+//! hash. Generally, nodes will be able to just sync "what's new" but for catch-up, this is true.
+//! As a consequence, a new network that hasn't yet gained enough members to start reducing how
+//! many partitions are covered by each node will have a higher overhead for each node.
+//! It is the responsibility of the gossip module to work out how to be efficient about sending
+//! the minimum amount of data required to keep all nodes up to date. However, the worst case is
+//! that 512 combined hashes must be sent to find out which partition has a mismatch for a given
+//! time slice. Over time, there will be more time slices to check.
+//! Therefore, the choice of 512 should be thought of as a best-effort choice to ask a reasonable
+//! amount of work and storage from each node in a large network but being no larger than that
+//! so we don't add overhead to gossip.
```
neonphog, line 50: :guitar:

neonphog (approved):
Yay, happy with this!

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,440 @@
...
+//! store to participate in the network and the amount of data that must be sent over the network
+//! to discover what op data needs to be fetched. Every node will store between 0 and 512
+//! partitions. For nodes that are choosing to store data, the minimum they can store is 1
+//! partition. Or in other words, 1/512th of the total data. To put this in context, an application
+//! that has a complete data set of 5TB would require a minimum of 10GB of storage dedicated to
+//! that app to be able to participate in the network.
+//! It is important to note though, that each partition is managing a [PartitionedTime] structure
+//! which requires some memory and processing to maintain. Nodes that take on more partitions will
+//! have more work to do and will have to send more data during gossip rounds. That happens when
+//! there are  changes to data that has made it into a time slice and become part of a combined
```
jost-s, line 40:
```suggestion
//! there are changes to data that has made it into a time slice and become part of a combined
```

crates/dht/src/hash.rs
```diff
@@ -0,0 +1,440 @@
...
+//! which requires some memory and processing to maintain. Nodes that take on more partitions will
+//! have more work to do and will have to send more data during gossip rounds. That happens when
+//! there are changes to data that has made it into a time slice and become part of a combined
+//! hash. Generally, nodes will be able to just sync "what's new" but for catch-up, this is true.
+//! As a consequence, a new network that hasn't yet gained enough members to start reducing how
+//! many partitions are covered by each node will have a higher overhead for each node.
+//! It is the responsibility of the gossip module to work out how to be efficient about sending
+//! the minimum amount of data required to keep all nodes up to date. However, the worst case is
+//! that 512 combined hashes must be sent to find out which partition has a mismatch for a given
+//! time slice. Over time, there will be more time slices to check.
```
jost-s, line 47: One more question, this module is about partitioning hash space and arcs. This worst case is when there's a mismatch of the top hash of space partitions, I believe. If the top hash mismatches, we'll need to compare all 512 combined hashes of all the space partitions, to find out which of them is mismatching. Why is the time slice here?

jost-s (approved):

crates/dht/src/hash.rs
ThetaSinner: I could do with drawing some pictures to go with what I'm writing :) Maybe we can embed SVGs into docs. I'll have a look as part of the gossip work.

What I'm getting at here is thinking about space like a disc. Then for the time slices you draw rings on that disc starting with the start time at the center of the disc and with time moving outwards.

There are then several ways to select from that data. You could take a disc out of the disc. That would involve collecting across all space segments and all time slices from the start time up to some limit of time slices.

You could take out a ring which would involve picking a single time slice id and then collecting across all space segments.

The answer is yes though, once you're collecting in some way and you find a mismatch at the "top level" of that structure, you will be forced to send a whole lot more data. How much data you have to send will depend on how much was included in the collection that you found a mismatch for. I think that's going to be the crux of designing the diffing logic. How do we make it likely that we send smaller amounts of data more often than we have to send a lot.

ThetaSinner: There's more work to be done on terminology and documentation here. I'm going to move forwards here and continue refining this writing as my ideas for how the diffing should work materialize

PR #48, opened 2024-12-09
BasicArc convert to type

ThetaSinner:
It's worth a quick scan to see what is moved vs new before digging into the code. However, what has moved has also changed due to a changed definition of the arc. See https://github.com/holochain/kitsune2/pull/48#discussion_r1876541143

### DIFF
--- a/crates/api/src/agent.rs
+++ b/crates/api/src/agent.rs
@@ -165,8 +165,8 @@ pub struct AgentInfo {
     pub url: Option<Url>,
 
     /// The arc over which this agent claims authority.
-    #[serde(default = "BasicArc::default")]
-    pub storage_arc: BasicArc,
+    #[serde(default = "StorageArc::default")]
+    pub storage_arc: StorageArc,
 }
 
 /// Signed agent information.
@@ -305,7 +305,7 @@ mod test {
         let now = Timestamp::from_micros(1731690797907204);
         let later = Timestamp::from_micros(now.as_micros() + 72_000_000_000);
         let url = Some(Url::from_str("ws://test.com:80/test-url").unwrap());
-        let storage_arc = BasicArc::Literal(42, u32::MAX / 13);
+        let storage_arc = StorageArc::Arc(42, u32::MAX / 13);
 
         let enc = AgentInfoSigned::sign(
             &TestCrypto,
@@ -325,7 +325,7 @@ mod test {
         .unwrap();
 
         assert_eq!(
-            r#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false,\"url\":\"ws://test.com:80/test-url\",\"storageArc\":{\"Literal\":[42,330382099]}}","signature":"ZmFrZS1zaWduYXR1cmU"}"#,
+            r#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false,\"url\":\"ws://test.com:80/test-url\",\"storageArc\":[42,330382099]}","signature":"ZmFrZS1zaWduYXR1cmU"}"#,
             enc
         );
 
@@ -341,14 +341,14 @@ mod test {
 
     #[tokio::test(flavor = "multi_thread")]
     async fn ignores_future_extension_fields() {
-        AgentInfoSigned::decode(&TestCrypto, br#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false,\"url\":\"ws://test.com:80/test-url\",\"storageArc\":{\"Literal\":[42,330382099]},\"fakeField\":\"bla\"}","signature":"ZmFrZS1zaWduYXR1cmU","fakeField2":"bla2"}"#).unwrap();
+        AgentInfoSigned::decode(&TestCrypto, br#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false,\"url\":\"ws://test.com:80/test-url\",\"storageArc\":[42,330382099],\"fakeField\":\"bla\"}","signature":"ZmFrZS1zaWduYXR1cmU","fakeField2":"bla2"}"#).unwrap();
     }
 
     #[tokio::test(flavor = "multi_thread")]
     async fn fills_in_default_fields() {
         let dec = AgentInfoSigned::decode(&TestCrypto, br#"{"agentInfo":"{\"agent\":\"dGVzdC1hZ2VudA\",\"space\":\"dGVzdC1zcGFjZQ\",\"createdAt\":\"1731690797907204\",\"expiresAt\":\"1731762797907204\",\"isTombstone\":false}","signature":"ZmFrZS1zaWduYXR1cmU"}"#).unwrap();
         assert!(dec.url.is_none());
-        assert_eq!(BasicArc::Empty, dec.storage_arc);
+        assert_eq!(StorageArc::Empty, dec.storage_arc);
     }
 
     #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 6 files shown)


crates/api/src/agent.rs
```diff
@@ -190,6 +165,7 @@ pub struct AgentInfo {
     pub url: Option<Url>,
 
     /// The arc over which this agent claims authority.
+    #[serde(default = "BasicArc::default")]
```
ThetaSinner, line 168: Should this be a required field? It was naturally optional due to being an `Option` type

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// Arc does not wrap around, loc < arc_start
+    /// |-----l---s------e-----|
+    /// 0                      u32::MAX
+    ///
+    /// Arc does not wrap around, loc >= arc_end
+    /// |---------s------e--l--|
+    /// 0                      u32::MAX
+    /// ```
+    pub fn dist(&self, loc: u32) -> u32 {
```
ThetaSinner, line 94: This has changed, because the distance to the upper bound needs to take into account the upper bound being exclusive. So with an upper bound of 4, a location of 5 is actually at distance 2.

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        assert!(arc.contains(u32::MAX - 1));
+        assert!(arc.contains(u32::MAX));
+        assert!(arc.contains(20));
+
+        assert!(!arc.contains(32));
+    }
+
+    #[test]
+    fn arc_dist_edge_cases() {
```
ThetaSinner, line 272: These two tests have been modified accordingly to the arc definition change

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Default)]
+pub enum BasicArc {
+    /// No DHT locations are contained within this arc.
+    #[default]
+    Empty,
+    /// All DHT locations are contained within this arc.
+    Full,
+    /// A specific range of DHT locations are contained within this arc.
+    ///
+    /// The lower bound is inclusive, the upper bound is exclusive.
```
ThetaSinner, line 20: To justify the reasoning for this, which I didn't do on the PR this is split from. Ops should always belong in exactly one arc range when partitioning the hash space. Otherwise you end up with a constant number of single-location overlaps between agents.

I didn't think this through carefully enough when reviewing the original agent definition PR. But I think having closed+open intervals for the arc is the correct logic.

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+
+/// A basic definition of a storage arc compatible with the concept of
+/// storage and querying of items in a store that fall within that arc.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Default)]
+pub enum BasicArc {
+    /// No DHT locations are contained within this arc.
+    #[default]
+    Empty,
+    /// All DHT locations are contained within this arc.
+    Full,
```
neonphog, line 17: This is a duplicate state. It can be represented by the Literal variant. I'd be inclined to get rid of Full, even though it makes some comparison logic simpler, just to avoid multiple variants meaning the same thing.

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// A specific range of DHT locations are contained within this arc.
+    ///
+    /// The lower bound is inclusive, the upper bound is exclusive.
+    Literal(u32, u32),
```
neonphog, line 21: I don't like how this serializes. Is this a case where `serde(untagged)` would be acceptable? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5fe9bcf2c03e3f5f2c41e2f6bc82be16

crates/api/src/arc.rs
```diff
@@ -0,0 +1,386 @@
...
+//! An arc can exist in its own right and refer to a range of locations on the DHT.
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+use serde::{Deserialize, Serialize};
+
+/// A basic definition of a storage arc compatible with the concept of
+/// storage and querying of items in a store that fall within that arc.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Default)]
+pub enum BasicArc {
```
neonphog, line 12: If there is no longer going to be more complicated Arc types, does "BasicArc" still make sense as a name? At the very least, it distinguishes it from std::sync::Arc...

crates/core/src/factories/mem_peer_store/test.rs
```diff
@@ -51,7 +51,7 @@ impl AgentBuild {
         });
         let is_tombstone = self.is_tombstone.unwrap_or(false);
         let url = self.url.unwrap_or(None);
-        let storage_arc = self.storage_arc.unwrap_or(BASIC_ARC_FULL);
+        let storage_arc = self.storage_arc.unwrap_or(BasicArc::Full);
```
neonphog, line 54: Per my previous comment about getting rid of the Full variant, we could add a const to the impl:

```rust
impl BasicArc {
    pub const FULL: Self = Self::Arc(0, 0);
}
```

So that this would become `BasicArc::FULL`.

crates/api/src/agent.rs
neonphog: See my other comment about the untagged serialization. But I think it's still correct if a peer doesn't specify an arc that we assume that peer does not claim any authority == Empty.

crates/api/src/arc.rs
neonphog: I also vote for changing this variant name to `Arc` instead of Literal.

neonphog (changes requested):
Thanks for pulling this out! Some open questions and suggestions.

crates/api/src/arc.rs
zippy: `HashArc` ?

crates/core/src/factories/mem_peer_store/test.rs
ThetaSinner: Done

crates/api/src/arc.rs
ThetaSinner: Renamed to `StorageArc`. That's what the field using this type was already called. Seems like a reasonable name to me

crates/api/src/arc.rs
ThetaSinner: Can do yes, it requires not having two fields with no values so it forces the Full to go

crates/api/src/arc.rs
ThetaSinner: It can be, and I didn't like that `Full` didn't have a canonical representation but I've gone back to inclusive bounds, so there is a single representation of full with `(0, u32::MAX)`

crates/api/src/arc.rs
ThetaSinner: You are correct. What's in my head is that I can't set up the arc in an entirely natural way with overlapping ends if they are inclusive bounds. It's not a big adjustment to my code on the other PR to fix that though. I was also thinking that it's nice to have the same boundary logic for time and hash space but time is continuous and the space is not. I'm just going to revert my boundary changes and I'll adjust my partitioning code on the other PR :)

crates/api/src/arc.rs
ThetaSinner: No longer modified, just re-organized and commented

crates/api/src/arc.rs
ThetaSinner: No longer changed

crates/api/src/arc.rs
```diff
@@ -0,0 +1,305 @@
...
+//! It can also be used in context, such as an agent. Where it represents the range of locations
+//! that agent is responsible for.
+
+use serde::{Deserialize, Serialize};
+
+/// The definition of a storage arc compatible with the concept of
+/// storage and querying of items in a store that fall within that arc.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Default)]
+#[serde(untagged)]
+pub enum StorageArc {
```
jost-s, line 13: We're talking about locations on the DHT, so how about
```suggestion
pub enum DhtArc {
```

jost-s (approved):
LGTM with suggestions

crates/api/src/arc.rs
jost-s: I see this has already been discussed.

ThetaSinner: I'd be happy with this name too, any other votes either way?

ThetaSinner: If not, I'll just do the rename :)

crates/api/src/arc.rs
```diff
@@ -0,0 +1,305 @@
+//! Kitsune2 arcs represent a range of hash locations on the DHT.
```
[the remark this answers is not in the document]
jost-s: I was thinking that locations and arcs are not tied to hashes. Especially in this module, as is said in the following paragraph of the module docs, an arc can exist in its own right and just delimits a range of locations. Either way is fine with me tho.

neonphog (approved):
Woot, looks awesome!

crates/api/src/arc.rs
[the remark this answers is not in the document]
ThetaSinner: Right okay, yes that second paragraph is confusing. I've simplified that text because it wasn't really useful as it was. I did mean that it always refers to hash locations but it might mean "this time slice is for some DhtArc" or it might mean "this agent is an authority for this DhtArc". We can come back and add more documentation when this type is actually used that way and it can all be linked together properly

PR #49, opened 2024-12-10
Add Taplo

ThetaSinner:


.github/workflows/test.yaml
```diff
@@ -6,29 +6,44 @@ on:
...
 
 jobs:
-  test:
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-      fail-fast: false
-
-    runs-on: ${{ matrix.os }}
+  static:
```
ThetaSinner, line 9: Moved static checks to only run once on Ubuntu. It could be argued that `clippy` should run everywhere but we don't have any platform specific code in Kitsune itself yet. Happy to move it back if that changes or if people just think that's more correct

.github/workflows/test.yaml
```diff
@@ -6,29 +6,44 @@ on:
...
       - uses: Swatinem/rust-cache@v2
 
+      - name: Toml format
+        run: make static-toml
+
       - name: Format
         run: make fmt
 
       - name: Lint
         run: make clippy
```
cdunster, line 28: Future work: There are some nice GitHub actions that do these checks but add comments and suggestions to the code instead of just failing. It might be good to look into them like https://github.com/marketplace/actions/clippy-check and https://github.com/marketplace/actions/check-formatting-of-rust-code-with-rustfmt

Makefile
```diff
@@ -6,6 +6,10 @@
 
 all: fmt clippy doc test
 
+static-toml:
+	taplo format --check ./*.toml
+	taplo format --check ./crates/**/*.toml
```
cdunster, line 11: Could also use the taplo config file to add included dirs: https://taplo.tamasfe.dev/configuration/file.html but we use `make` so I guess it's no different.

cdunster (approved):
A couple of ideas but nothing to block this PR :slightly_smiling_face:

.github/workflows/test.yaml
ThetaSinner: My opinion is that with a project this size, I should always `make` before I push and if i forget then the time spent opening the build failure and looking at what I did wrong is a just punishment for wasting CI time :D 

The second action does look nice, as the amount of code here grows it could be useful to reviewers to see that instead of a build log

neonphog (approved):

PR #50, opened 2024-12-10
Bootstrap Client Module

neonphog:
DEPENDS ON https://github.com/holochain/kitsune2/pull/58

- Adds a bootstrap test server helper to the test_utils crate that *should* let us have zero config testing bootstrap servers.
- The test_utils crate is now a dependency of the core crate if the `test_utils` feature is enabled. Let me know if we should do this differently.
- Bootstrap module api written
- Bootstrap module core implementation that can serve both for production and for testing written.
- I only ended up with two tests... one that exercises the TestBootstrapSrv helper, and one that proves out the client retry loops by having a server that errors until we tell it not to. -- let me know if there are more tests I should write.


crates/api/src/config.rs
```diff
@@ -61,6 +61,19 @@ impl Config {
         Ok(())
     }
 
+    /// Update a module config with new values. Note, this will overwrite
+    /// ALL config values. If you would like to only update some, please
+    /// first call [Self::get_module_config], update the appropriate values,
+    /// then call this function again to set them.
+    pub fn set_module_config<M: ModConfig>(
```
ThetaSinner, line 68: This means that it's going to be tricky to report which config items that have been changed don't have any effect when changed at runtime. It might be easier to receive patch/update request when dealing with that kind of thing?

crates/bootstrap_srv/src/server.rs
```diff
@@ -68,7 +71,10 @@ impl BootstrapSrv {
 
         // get the address that was assigned
         let addrs = server.server_addrs().to_vec();
-        println!("Listening at {:?}", addrs);
+        for addr in addrs.iter() {
+            // print these separately incase someone wants to parse them
+            tracing::info!(addr = format!("{:?}", addr), "Listening");
```
ThetaSinner, line 76: Would that be easier if they went to stdout rather than out with the tracing?

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+
+/// Configuration parameters for [CoreBootstrapFactory].
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CoreBootstrapConfig {
+    /// The url of the kitsune2 bootstrap server. E.g. `https://boot.kitsu.ne`.
+    ///
+    /// # Feature `test_utils`
+    ///
+    /// If the `test_utils` feature is enabled, the default will change.
```
ThetaSinner, line 19: I'd quite like to avoid this. As much as possible, features should add new functionality and not change existing functionality. I'm just thinking about later when we want to have a mixture of tests that use the test bootstrapping and others that use a "real" local bootstrap server. We don't be able to do both in the same crate with `test_utils` turned on.

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+
+impl Bootstrap for CoreBootstrap {
+    fn put(&self, info: Arc<agent::AgentInfoSigned>) {
+        // ignore puts outside our space.
+        if info.space != self.space {
+            return;
+        }
+
+        // if we can't push onto our large buffer channel... we've got problems
+        let _ = self.push_send.try_send(info);
```
ThetaSinner, line 190: Similarly, I'd be tempted to log here, this isn't expected to happen

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+    push_send: PushSend,
+    mut push_recv: PushRecv,
+) {
+    let mut wait = None;
+
+    while let Some(info) = push_recv.recv().await {
+        let url =
+            format!("{server_url}/bootstrap/{}/{}", &info.space, &info.agent);
+        let enc = match info.encode() {
+            Err(_) => continue,
```
ThetaSinner, line 206: log please

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+                        let mut p = p * 2;
+                        if p > config.backoff_max() {
+                            p = config.backoff_max();
+                        }
+                        wait = Some(p);
+                    }
+                }
+
+                // wait for the backoff time
+                if let Some(wait) = &wait {
```
ThetaSinner, line 240: This is okay because we just have one bootstrapped agent per space? Otherwise we're blocking consuming the queue for the other agent

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+                let _ = peer_store.insert(list).await;
+            }
+        }
+
+        wait *= 2;
+        if wait > config.backoff_max() {
+            wait = config.backoff_max();
+        }
+
+        tokio::time::sleep(wait).await;
```
ThetaSinner, line 291: Should `wait` be reset to `config.backoff_min()` after the sleep? Otherwise it gets set to max on the first success and stays there forever

Makefile
```diff
@@ -2,9 +2,9 @@
 # tasks both in automation and locally until we figure out better
 # release automation tools.
 
-.PHONY: all fmt clippy doc build test
+.PHONY: all static-toml fmt clippy doc build test
 
-all: fmt clippy doc test
+all: static-toml fmt clippy doc build test
```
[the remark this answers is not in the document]
neonphog: I like to run `make` locally and know it's highly likely to pass CI. That's just my strategy, though. Happy to find common ground if this is going to throw other's processes off : )

crates/api/src/agent.rs
```diff
@@ -105,6 +105,17 @@ pub trait Verifier: std::fmt::Debug {
 /// Trait-object [Verifier].
 pub type DynVerifier = Arc<dyn Verifier + 'static + Send + Sync>;
 
+impl Verifier for DynVerifier {
```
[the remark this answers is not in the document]
neonphog: Certainly seems like it should work, right? But: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f7593dc0905e05c21c25dd607aa5b3b3

crates/api/src/config.rs
neonphog: I'm not sure I fully understand your suggestion, although I'm also not super happy with the interface here.

My initial thought was to leave the structure completely up to the module implementations, but perhaps this would all be more intuitive if we exposed a map of properties for each module in the interface. Then only those properties that were included in the set call would be overwritten?

crates/bootstrap_srv/src/server.rs
neonphog: :+1: Yeah, that's probably better. Will update.

crates/core/src/factories/core_bootstrap.rs
neonphog: This module supports both `test:` bootstrap urls and real actual urls at the same time. It's just the default config that changes when the feature is enabled.

I do agree though that this is a little strange. If you would rather there be clear separate modules for test bootstrapping and production bootstrapping, I'm happy to separate them out.

crates/core/src/factories/core_bootstrap.rs
neonphog: The other push task loop does that. I feel like this one is less dire... once we're already up and running (i.e. we've had a single bootstrap success), we should be gossiping agents and not rely on bootstrapping anymore. I was thinking even if we missed a poll once in five minutes, it'd be fine to just try again five minutes later.

Happy to change this if you think it is needed.

crates/api/src/agent.rs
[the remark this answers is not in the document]
ThetaSinner: Weird, sounds like that's one for the Rust team to figure out :)

crates/api/src/config.rs
ThetaSinner: I'm thinking if I'm a module receiving new configuration and I see:

```
mymod.config1 = "hello",
mymod.config2 = 20,
```

Then I get an update saying:


```
mymod.config1 = "hallo",
mymod.config2 = 25,
```

Then I have to figure out that `mymod.config1` changed which let's say isn't supported. Then I also have to figure out that `mymod.config2` changed and potentially take some action to start using that new config.

If instead I only get an input that says:


```
mymod.config2 = 25,
```

Then I can easily check that only properties that are allowed to be changed are changing and I can take an actions required on other config values changing as they're applied to my config set.

crates/core/src/factories/core_bootstrap.rs
ThetaSinner: I see the logic at the moment and it makes sense. I'm cautious about it because I'm not sure it will be maintained as cleanly as it's been written and it's one of the patterns that causes me a lot of headache time around Holochain. We have quite a few places where we do things like this that aren't easy to work out with, even if they started out that way.

I'd lean towards separating even if it means a bit of extra code

crates/core/src/factories/core_bootstrap.rs
ThetaSinner: Yep fair enough, that makes sense!

crates/core/src/factories/core_bootstrap.rs
ThetaSinner: Makes sense to me. Errors 5 minutes apart seems relatively unlikely unless something is also going to be wrong during that interval too

crates/api/src/config.rs
neonphog: This is not for runtime changes in config. It all happens before the module receives the config in the constructor. Actually, this could be considered a test function, since in a production use-case this should all be set via config file. I only needed this so I could alter the defaults for specific tests.

But this is certainly pointing to a need for the api to be different : )

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -23,13 +26,28 @@ pub struct Args {
...
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::builder()
+                .with_default_directive(tracing::Level::DEBUG.into())
+                .from_env_lossy(),
+        )
+        .with_file(true)
+        .with_line_number(true);
+
+    if args.json {
+        let _ = t.json().try_init();
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/50/commits/2b7b29719aa4dd993ceaac4fb5454ca15b7de413

crates/bootstrap_srv/src/server.rs
neonphog: https://github.com/holochain/kitsune2/pull/50/commits/b063bf1c9d044f41e9cedc681e1081f8048b16b7

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,296 @@
...
+            _test_server,
+        }
+    }
+}
+
+impl Bootstrap for CoreBootstrap {
+    fn put(&self, info: Arc<agent::AgentInfoSigned>) {
+        // ignore puts outside our space.
+        if info.space != self.space {
+            return;
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/50/commits/4fc6f10d14fd32cdce4df82d1ba8ed9369b8da13

crates/core/src/factories/core_bootstrap.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/50/commits/4fc6f10d14fd32cdce4df82d1ba8ed9369b8da13

Makefile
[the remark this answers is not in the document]
jost-s: I'm fine with this too. I think keeping the installation out of the make file will help with not having to rebuild everything on every invocation of the make command.

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,263 @@
...
+    config: CoreBootstrapConfig,
+    server_url: Arc<str>,
+    space: SpaceId,
+    peer_store: peer_store::DynPeerStore,
+) {
+    let mut wait = config.backoff_min();
+
+    loop {
+        let url = format!("{server_url}/bootstrap/{space}");
+        if let Ok(Ok(data)) = tokio::task::spawn_blocking(move || {
```
jost-s, line 226: Should failure be logged here, in case the polling fails continuously and no one notices?

crates/core/src/factories/core_bootstrap.rs
```diff
@@ -0,0 +1,263 @@
...
+            ureq::get(&url)
+                .call()
+                .map_err(std::io::Error::other)?
+                .into_string()
+        })
+        .await
+        {
+            if let Ok(list) = agent::AgentInfoSigned::decode_list(
```
jost-s, line 234: Is this worth logging when it fails? Could be that the bootstrap server is filled with invalid data?

jost-s (dismissed):
LGTM
I can't think of more tests to write either.

crates/api/src/config.rs
ThetaSinner: Ahh okay, so can/should this go behind `#[cfg(test)]` for now?

neonphog: I'm working on a new config strategy. If it works out we can merge that first and I'll update this code to use it and hopefully it won't be as awkward.

neonphog: Okay, please see this: https://github.com/holochain/kitsune2/pull/58

crates/api/src/config.rs
```diff
@@ -61,6 +61,19 @@ impl Config {
         Ok(())
     }
 
+    /// Update a module config with new values. Note, this will overwrite
+    /// ALL config values. If you would like to only update some, please
+    /// first call [Self::get_module_config], update the appropriate values,
+    /// then call this function again to set them.
```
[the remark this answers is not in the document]
neonphog: This has been deleted.

crates/core/src/factories/core_bootstrap.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/50/commits/2ddd8fe0a2bf0798767993c30e306957b2520138

crates/core/src/factories/core_bootstrap.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/50/commits/2ddd8fe0a2bf0798767993c30e306957b2520138

ThetaSinner (approved):
LGTM!

PR #51, opened 2024-12-11
DHT module docs

ThetaSinner:
This PR now does three things. Some minor code cleanup and documentation consistency for existing code. Then I've also added some top level crate docs with diagrams and started a glossary. Initially this PR just had the glossary.

---

It might be too early for this, just thinking about where to put the glossary of terms that @zippy asked for. I think it's a good idea, we've already got a bunch of terms we use that wouldn't be clear to anybody who didn't work on Kitsune1 :)

I'm thinking the API is a decent place. Everything will have a trail back to the API, it's the starting point for anybody integrating with Kitsune2.
A better place might be the main module (the equivalent of kitsune_p2p) because that's where we'll have the maximum number of dependencies available to use for type references.

Doing a glossary like this, we get nice easy to reference links to terms -> file:///<my secret source code path>/kitsune2/target/doc/kitsune2_api/doc/glossary/index.html#time-slice


crates/api/src/doc/glossary.rs
```diff
@@ -0,0 +1,9 @@
+//! A glossary of terms used in the Kitsune2 project.
+//!
+//! ## Time slice
+//! A time slice is a period of time. It is defined to be a an interval from a start time,
```
jost-s, line 4:
```suggestion
//! A time slice is a period of time. It is defined to be an interval from a start time,
```

crates/api/src/doc/glossary.rs
```diff
@@ -0,0 +1,9 @@
...
+//! inclusive, up to some end time, exclusive.
+//!
+//! Time slices are used to group data by time, so that a single hash describing that data can be
```
jost-s, line 7:
```suggestion
//! Time slices are used to group hashed data by time, so that a single hash describing that data can be
```

crates/api/src/doc/glossary.rs
```diff
@@ -0,0 +1,9 @@
+//! A glossary of terms used in the Kitsune2 project.
+//!
+//! ## Time slice
+//! A time slice is a period of time. It is defined to be a an interval from a start time,
+//! inclusive, up to some end time, exclusive.
```
jost-s, line 5:
```suggestion
//! inclusive, up to some end time, exclusive. It can be called a time partition as well.
```

jost-s (approved):
Yeah, that's good.

neonphog (approved):

crates/api/src/doc/glossary.rs
ThetaSinner: I will go back through my other doc comments and be more careful about this. A partition isn't intended to refer to an individual slice within the partition, the partition is meant to refer to the collection of time slices

ThetaSinner:
I think I prefer the `main` crate too. I will commit to moving this module there once it exists and for now, I'll add to this? :)

jost-s:
sure!

crates/api/src/doc/glossary.rs
jost-s: Ah, okay. And a space partition is then the collection of space slices? Does that sound right?

ThetaSinner: We need a term for this. I wasn't going to use slice. I've been using segment but that's actually wrong and what I should be saying is "sector"... Dear me, evolving new terminology is messy. I'm just going to have to keep going over these docs and making them consistent with themselves

jost-s: Yeah, sector is the geometrical term for a space partition with its time slices. I like the circle analogy. We could also use a line with stacks on it. Anyhow, off this goes.

ThetaSinner:
I'll continue this after #59

crates/dht/art/dht-circle.svg
ThetaSinner, line 1: Ooh, clever GitHub. I thought these would be a pain to review :)

crates/dht/build.rs
```diff
@@ -0,0 +1,32 @@
+use handlebars::Handlebars;
+use std::collections::HashMap;
+use std::io::BufRead;
+use std::path::Path;
+
+fn main() {
```
ThetaSinner, line 6: No support for embedding files in docs by default. Some light hackery to get this working. I'm just templating a markdown file to embed SVGs and then including the generated markdown file into the crate level docs.

crates/dht/src/arc_set.rs
```diff
@@ -18,15 +19,11 @@ pub struct ArcSet {
 impl ArcSet {
     /// Create a new arc set from a list of arcs.
     ///
-    /// The size parameter determines the size of each sector. When divided into U32::MAX + 1, the
-    /// resulting factor must be a power of 2. This is the same sizing logic found in
-    /// [PartitionedHashes::try_from_store](crate::hash::PartitionedHashes::try_from_store).
-    ///
     /// The resulting arc set represents the union of the input arcs.
-    pub fn new(size: u32, arcs: Vec<DhtArc>) -> K2Result<Self> {
```
ThetaSinner, line 26: This didn't need to be a parameter while the sector size is hard-coded in the hash model

crates/dht/src/constant.rs
```diff
@@ -4,3 +4,8 @@ use std::time::Duration;
 
 /// The smallest unit of time in the DHT.
 pub const UNIT_TIME: Duration = Duration::from_secs(15 * 60); // 15 minutes
+
+/// The size of a hash sector in the hash partition of the DHT.
+///
+/// Results in 512 sectors, because 32 - 23 = 9, and 2^9 = 512.
+pub const SECTOR_SIZE: u32 = 1u32 << 23;
```
ThetaSinner, line 11: Extracted existing constant from hash module

crates/dht/src/dht.rs
```diff
@@ -57,16 +61,18 @@ pub enum DhtSnapshotNextAction {
...
+    /// be approximately 5.3 days.
     pub async fn try_from_store(
         current_time: Timestamp,
         store: DynOpStore,
     ) -> K2Result<Dht> {
         Ok(Dht {
-            partition: PartitionedHashes::try_from_store(
-                14,
+            partition: HashPartition::try_from_store(
+                9,
```
ThetaSinner, line 75: This was a mistake on my part. 14 creates seriously large full time slices. I'd originally documented 9 intending to use that. It means you get full time slices that are about 5 days.

neonphog:
Suggestion: https://github.com/holochain/kitsune2/pull/66

crates/dht/build.rs
```diff
@@ -0,0 +1,32 @@
...
+        .expect("Failed to register template");
+
+    let mut data = HashMap::new();
+    data.insert("dht-graph", load_svg("art/dht-graph.svg"));
+    data.insert("dht-circle", load_svg("art/dht-circle.svg"));
+
+    let mut f =
+        std::fs::File::create("DOCS.md").expect("Failed to create DOCS.md");
+    reg.render_to_write("DOCS.md", &data, &mut f)
+        .expect("Failed to render template");
```
neonphog, line 22: Rather than writing a file in the source directory that has to be git ignored, it is canonical to write it to the `OUT_DIR` directory, so it ends up in the target folder.

crates/dht/src/time.rs
```diff
@@ -60,10 +64,10 @@ use kitsune2_api::{
 };
 use std::time::Duration;
 
-/// The partitioned time structure.
+/// The time partition structure.
 #[derive(Debug)]
 #[cfg_attr(test, derive(Clone, PartialEq))]
-pub struct PartitionedTime {
+pub struct TimePartition {
```
neonphog, line 70: Oo I like that better : )

crates/dht/build.rs
neonphog: It looks like this works if you want to get rid of the build.rs:

```rust
//! <details><summary>bla</summary>
#![doc = include_str!("../art/dht-circle.svg")]
//! </details>
```

neonphog (dismissed):
Some different options for the build.rs and svgs. All the text and changes look good to me!

crates/dht/build.rs
ThetaSinner: This is gone, but yes, you're right

crates/dht/build.rs
ThetaSinner: Oh sweet, I didn't realise from the docs that you could mix these. Yes that works nicely, done

PR #52, opened 2024-12-11
tiny_http -> axum

neonphog:


crates/bootstrap_srv/src/http.rs
```diff
@@ -0,0 +1,254 @@
...
+                    "/bootstrap/:space/:agent",
+                    routing::put(handle_boot_put),
+                )
+                .layer(extract::DefaultBodyLimit::max(1024))
+                .with_state(h_send.clone());
+
+            let receiver = HttpReceiver(h_recv);
+
+            let listener =
+                match tokio::net::TcpListener::bind(config.addr).await {
```
ThetaSinner, line 136: It was less obvious before and I missed asking about it, but ipv4/v6. Does supporting both matter here?

ThetaSinner (approved):

crates/bootstrap_srv/Cargo.toml
```diff
@@ -11,6 +11,11 @@ categories = ["network-programming"]
 edition = "2021"
 
 [dependencies]
+async-channel = { workspace = true }
+axum = { workspace = true, default-features = false, features = [
+  "http1",
```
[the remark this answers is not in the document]
neonphog: We can easily add it. I imagine this server would almost always want to be fronted by a reverse proxy (cloudflare) so not strictly required. I could expose it as a feature to be enabled, though : )

crates/bootstrap_srv/src/http.rs
neonphog: Yes, this is another area where tiny_http was immature. It let you specify multiple bindings but only reported a single one, so was not super useful... I'll update the api to be more reasonable now that we can.

crates/bootstrap_srv/Cargo.toml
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/52/commits/ec0554c8625e2f128bab5b88525ded508862bfd0

crates/bootstrap_srv/src/http.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/52/commits/a08fdf59d9acac72434cea726d800d6a99152db7

crates/bootstrap_srv/Cargo.toml
```diff
@@ -11,16 +11,28 @@ categories = ["network-programming"]
...
 clap = { workspace = true, features = ["derive", "wrap_help"] }
 ctrlc = { workspace = true }
 ed25519-dalek = { workspace = true }
+futures = { workspace = true }
 num_cpus = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tempfile = { workspace = true }
-tiny_http = { workspace = true }
+tokio = { workspace = true, features = ["time", "rt", "rt-multi-thread"] }
```
jost-s, line 29: Actually one question - what is "time" needed for?

neonphog: [https://github.com/holochain/kitsune2/pull/52/files#diff-058df3dc5027b744df2256bd61b759aa85597ee130343b8f3f8af53cc9600034R183-R191](https://github.com/holochain/kitsune2/pull/52/files#diff-058df3dc5027b744df2256bd61b759aa85597ee130343b8f3f8af53cc9600034R183-R191)

```rust
// /crates/bootstrap_srv/src/http.rs
/*191*/     match tokio::time::timeout(std::time::Duration::from_secs(10), async move {
```

PR #53, opened 2024-12-12
Transport Trait

neonphog:
Instead of `Transport` directly being a struct, use our trait+trait-object pattern for test-ability.

There is no functional change in this PR, it is purely moving code around + renaming.


jost-s (approved):
:man_dancing:

PR #54, opened 2024-12-13
Transport Integration - Space Notify Send & Receive

neonphog:
- Cleans up some of the apis & adds some events to the top-level handler
- Kitsune top-module creates a tx instance and passes it to spaces
- Space module registers space handlers with the tx module
- Protocol and implementation of space send & recv


crates/api/src/kitsune.rs
```diff
@@ -5,6 +5,18 @@ use std::sync::Arc;
 
 /// Handler for events coming out of Kitsune2.
 pub trait KitsuneHandler: 'static + Send + Sync + std::fmt::Debug {
+    /// A notification that a new listening address has been bound.
+    /// Peers should now go to this new address to reach this node.
+    fn new_listening_address(&self, this_url: Url) {
+        drop(this_url);
```
jost-s, line 11: What are these `drop`s for?

jost-s (approved):

crates/api/src/kitsune.rs
neonphog: Underscores at the beginning of parameters show up in the docs, so I don't like them there. The explicit drops prevent the unused parameter warning. These default implementations don't need to do anything.

jost-s: But you want a default implementation that does nothing, I guess. Alright, got it.

jost-s (approved):

ThetaSinner (approved):
LGTM!

PR #55, opened 2024-12-13
Signing Agent Infos

neonphog:
Working on finishing off the space module. However, since this is so cross-cutting, I'm not quite there yet, and would like to get this part reviewed and merged before I make it even more unwieldy.

This PR:
- Adds some current and target storage arc APIs to the LocalAgent type which gossip and sharding will make use of.
- Adds hooks in the space module to sign agent infos on JOIN and sign tombstones on LEAVE.
- Moves the peer store agent builder into test_utils crate and adds some TestLocalAgent helper structs.

Follow-on work:
- Still need a task to check for about-to-expire local infos and sign new ones.
- Still need to inform bootstrap/gossip/sharding modules of updates.


crates/core/src/factories/core_fetch/test.rs
```diff
@@ -4,15 +4,19 @@ use std::{
...
 use bytes::Bytes;
+
 use kitsune2_api::{
     fetch::{deserialize_op_ids, Fetch},
     id::Id,
     transport::Transport,
     AgentId, K2Error, OpId, SpaceId, Url,
 };
+
+use kitsune2_test_utils::agent::*;
```
ThetaSinner, line 15: Why spaces between imports? Can keep them in a block and let `cargo fmt` keep them tidy?

crates/core/src/factories/core_space.rs
```diff
@@ -92,13 +134,142 @@ impl Space for CoreSpace {
 
     fn local_agent_join(
         &self,
-        _local_agent: agent::DynLocalAgent,
+        local_agent: agent::DynLocalAgent,
     ) -> BoxFut<'_, K2Result<()>> {
-        Box::pin(async move { todo!() })
+        Box::pin(async move {
+            // set some starting values
+            local_agent.set_tgt_storage_arc_hint(DhtArc::Empty);
```
ThetaSinner, line 141: This is fine for now, I'd like to come back to this per the comments I've left above

crates/core/src/factories/core_space.rs
```diff
@@ -92,13 +134,142 @@ impl Space for CoreSpace {
...
+                        let expires_at = created_at
+                            + std::time::Duration::from_secs(60 * 20);
+                        let info = agent::AgentInfo {
+                            agent: local_agent2.agent().clone(),
+                            space,
+                            created_at,
+                            expires_at,
+                            is_tombstone: false,
+                            url: Some(url),
+                            storage_arc: local_agent2.get_cur_storage_arc(),
```
ThetaSinner, line 181: The target storage arc is only available on local for now which I think makes sense. The host needs to have some influence on that value and Kitsune needs to be able to adjust it. Is it not useful to other peers though? I guess they can't take it into account for arc sizing calculations because that would be wrong. I do think it's wanted in gossip though, I'm still puzzling that out. So for gossip maybe we need to be sending that information as part of sharing agent info

crates/core/src/factories/core_space.rs
```diff
@@ -92,13 +134,142 @@ impl Space for CoreSpace {
...
+
+                        // add it to the peer_store.
+                        if let Err(err) = peer_store.insert(vec![info]).await {
+                            tracing::warn!(
+                                ?err,
+                                "failed to add agent info to peer store"
+                            );
+                        }
+
+                        // TODO - send the new agent info to bootstrap.
```
ThetaSinner, line 208: Depends on the bootstrap client which has been merged right? Is this TODO just being left to keep this review under control or was it just not possible yet when the TODO was written?

crates/core/src/factories/core_space/test.rs
```diff
@@ -1,56 +1,89 @@
...
+        active_peer_count = 0;
+        for peer in s1.peer_store().get_all().await.unwrap() {
+            if !peer.is_tombstone {
+                active_peer_count += 1;
+            }
+        }
+        if active_peer_count == 2 {
+            break;
+        }
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
```
ThetaSinner, line 58: These are accumulating a bit... I really like how fast the tests are at the moment and no matter how slow it gets to run them all, I'll be able to run per module so it won't be too bad.

Is it worth the effort to think about how else this could be tested without a sleep?

crates/test_utils/src/agent.rs
```diff
@@ -0,0 +1,164 @@
...
+
+struct TestLocalAgentInner {
+    cb: Option<Arc<dyn Fn() + 'static + Send + Sync>>,
+    cur: DhtArc,
+    tgt: DhtArc,
+}
+
+/// A test local agent is generated from incrementing id keys,
+/// and always signs with the signature `b"test-signature"`.
+pub struct TestLocalAgent {
```
ThetaSinner, line 37: Ahh, clearly not. This is the test agent. So I'm unclear about the use of the agent type above then

ThetaSinner (commented):
Lots to ask and discuss, not many changes to request though

crates/api/src/agent.rs
```diff
@@ -121,11 +121,83 @@ impl Verifier for DynVerifier {
...
+    /// Invoke the registered cb if one has been set.
+    /// This can be ignored if [Self::register_cb] has not yet been called.
+    fn invoke_cb(&self);
+
+    /// Access the current storage arc for this local agent.
+    /// This will be used by the space module to construct [AgentInfoSigned].
+    fn get_cur_storage_arc(&self) -> DhtArc;
+
+    /// Set the current storage arc for this local agent.
+    /// This will be initially set to zero on space join,
```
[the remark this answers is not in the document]
neonphog: Yep, this exactly.

crates/api/src/agent.rs
```diff
@@ -121,11 +121,83 @@ impl Verifier for DynVerifier {
...
+    /// This is a chance for the implementor to influence how large
+    /// a storage arc should be for this agent. The gossip module will
+    /// attempt to collect enough data for claiming storage authority
+    /// over this range.
+    fn get_tgt_storage_arc(&self) -> DhtArc;
+
+    /// The sharding module will attempt to determine an ideal target
+    /// arc for this agent. An implementation is free to use or discard
+    /// this information when returning the arc in [Self::get_tgt_storage_arc].
+    /// This will initially be set to zero on join, but the sharding module
```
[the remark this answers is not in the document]
neonphog: The thought here was to allow the "factor" i.e. holoport hosts that are hosting web users with zero arcs may want a factor 4 so they are gossiping on their behalf if not their exact arc locations. I.e. the "hint" would come in, and the specific LocalAgent implementation would actually store 4x that hint.

crates/core/src/factories/core_space.rs
neonphog: For sure gossip needs to be dealing with the arcs as we're trying to sync them, not the currently signed agent storage arcs. I don't believe the targets belong in the signed agent infos, but rather should be communicated directly in the gossip messages.

crates/core/src/factories/core_space.rs
neonphog: Correct, this means notify the bootstrap module, the module itself will decide if it sends to the server or not.

The PR was getting crazy, I just picked an arbitrary place to break off new work : )

crates/core/src/lib.rs
```diff
@@ -37,6 +38,104 @@ impl agent::Verifier for Ed25519Verifier {
...
 }
 
+struct Ed25519LocalAgentInner {
+    cb: Option<Arc<dyn Fn() + 'static + Send + Sync>>,
+    cur: DhtArc,
+    tgt: DhtArc,
+}
+
+/// A default in-memory [kitsune2_api::agent::LocalAgent] based on ed25519.
+pub struct Ed25519LocalAgent {
```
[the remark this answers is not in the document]
neonphog: For me, this is a minimum viable implementation that someone not using lair might choose to use in a production application. That's why I put it in core. But happy to consider alternatives : )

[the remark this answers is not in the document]
neonphog: (there is a `from_seed` constructor, so you can use your own seed persisted outside the scope of kitsune)

Cargo.toml
```diff
@@ -32,11 +32,13 @@ clap = "4.5.21"
 # on shutdown.
 ctrlc = { version = "3.4.5", features = ["termination"] }
 # bootstrap_srv for signature verification.
-ed25519-dalek = "2.1.1"
+ed25519-dalek = { version = "2.1.1", features = ["rand_core"] }
```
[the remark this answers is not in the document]
neonphog: Pretty sure I'm setting this in the module too : ) I wonder if I just edited the wrong file : )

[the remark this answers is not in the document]
neonphog: I was not, in fact, setting it in the crate. Weird: https://github.com/holochain/kitsune2/pull/55/commits/df970190128282d25ecde03bacd75f11f00ba3c2

crates/core/src/factories/core_space/test.rs
neonphog: Maybe a little bit better at least? https://github.com/holochain/kitsune2/pull/55/commits/f5120be2a6969b652225d5c17dabc066e9bc448f

crates/core/src/factories/core_fetch/test.rs
neonphog: I touched this code, I can fix it : ) -- https://github.com/holochain/kitsune2/pull/55/commits/db795946a666d7f60689ebd293f007589e538734

crates/api/src/agent.rs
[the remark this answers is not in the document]
ThetaSinner: I see, okay. Maybe this is a bit far out to be worry about too much then. That sounds reasonable and then we have to figure out what the defaults and clamping look like so we can update documentation later if we stray outside that. I think the logic that's being introduced here is sound.

crates/core/src/factories/core_space.rs
ThetaSinner: Sweet, that lines up nicely with my thinking out loud :)

crates/core/src/factories/core_space/test.rs
ThetaSinner: Yes that is better, I'm happy to trade a bit of extra heat for faster tests :)

crates/core/src/lib.rs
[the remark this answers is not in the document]
ThetaSinner: Gotcha, yeah that's good then. I'd rather use something like this than Lair in Wind Tunnel tests for example, just because it's simpler and less to go wrong and update. Now I know what this is for, I think it makes sense for it to be here

ThetaSinner (dismissed):
Good to go! :)

crates/core/src/factories/core_space.rs
```diff
@@ -92,13 +134,142 @@ impl Space for CoreSpace {
...
+                        tracing::warn!(?err, "failed to sign agent info");
+                        return;
+                    }
+                    Ok(info) => info,
+                };
+
+                if let Err(err) = self.peer_store.insert(vec![info]).await {
+                    tracing::warn!(
+                        ?err,
+                        "failed to add agent info to peer store"
```
jost-s, line 266: The method name is shown in the trace as well, I believe, but this describes the operation a bit clearer
```suggestion
                        "failed to tombstone agent info in peer store"
```

jost-s (approved):
LGTM

crates/api/src/agent.rs
```diff
@@ -83,11 +83,11 @@ use std::sync::Arc;
 /// Defines a type capable of cryptographic signatures.
 pub trait Signer {
     /// Sign the encoded data, returning the resulting detached signature bytes.
-    fn sign(
-        &self,
-        agent_info: &AgentInfo,
-        message: &[u8],
-    ) -> BoxFut<'_, K2Result<bytes::Bytes>>;
+    fn sign<'a, 'b: 'a, 'c: 'a>(
```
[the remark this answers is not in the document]
neonphog: If you used a `'static` `b""` type in calling this message, it would pollute the requirements and make it seem as if the reference to self also had to be `'static`. As far as I can tell, this now says `'b` and `'c` can be different than the `'a` used in self, but they have to be valid at least as long as self.

Although I still find lifetimes pretty fiddly...

crates/core/src/factories/mem_peer_store/test.rs
```diff
@@ -11,8 +11,8 @@ fn create() -> Inner {
     )
 }
 
-const AGENT_1: AgentId = AgentId(Id(bytes::Bytes::from_static(b"agent1")));
-const AGENT_2: AgentId = AgentId(Id(bytes::Bytes::from_static(b"agent2")));
+const AGENT_1: AgentId = AgentId(id::Id(bytes::Bytes::from_static(b"agent1")));
```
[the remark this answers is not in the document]
neonphog: I don't think this works for a const... the from trait doesn't mark that fn as const.

crates/api/src/agent.rs
```diff
@@ -121,11 +121,84 @@ impl Verifier for DynVerifier {
...
+
+    /// Access the current storage arc for this local agent.
+    ///
+    /// This will be used by the space module to construct [AgentInfoSigned].
+    fn get_cur_storage_arc(&self) -> DhtArc;
+
+    /// Set the current storage arc for this local agent.
+    /// This will be initially set to zero on space join,
+    /// then will be updated by the gossip module as best effort to
+    /// collect data for the range is made.
```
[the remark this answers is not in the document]
neonphog: Fixed: https://github.com/holochain/kitsune2/pull/55/commits/d198551853b2c51769998819ba3a8b0b2256e884

ThetaSinner (approved):

jost-s (approved):

jost-s (approved):

PR #56, opened 2024-12-17
refactor(fetch): increment back off times

jost-s:
Previously fetch requests were only re-inserted into the queue when successfully sent. That renders the cool down/back off list ineffective, because unresponsive agents are put on the list but requests to them are never retried.

Requests are always added to the queue now. The concern about state building up if agents on the back off list never become responsive again is addressed by introducing an incremental back off time with a maximum. When the maximum back off is hit for an agent, the request is removed from the set of fetch requests to make. This will effectively prevent the build-up of state and leads to giving up on completely unresponsive agents.


crates/core/src/factories/core_fetch.rs
```diff
@@ -276,38 +282,36 @@ impl CoreFetch {
...
-                        {
-                            tracing::warn!("could not re-insert fetch request for op {op_id} to agent {agent_id} into queue: {err}");
-                            // Remove op id/agent id from set to prevent build-up of state.
-                            state
-                                .lock()
-                                .unwrap()
-                                .ops
-                                .remove(&(op_id, agent_id));
-                        }
-                    }
```
jost-s, line 291: Always re-insert requests back into queue, not only when successful.

crates/core/src/factories/core_fetch.rs
```diff
@@ -106,15 +107,19 @@ impl FetchFactory for CoreFetchFactory {
 pub struct CoreFetchConfig {
     /// How many parallel op fetch requests can be made at once. Default: 2.  
     pub parallel_request_count: u8,
-    /// Duration in ms to keep an unresponsive agent on the cool-down list. Default: 120_000.
-    pub cool_down_interval_ms: u64,
+    /// Duration in ms to keep an unresponsive agent on the back off list. Default: 120_000.
+    pub back_off_interval_ms: u64,
+    /// Maximum exponent for back off interval. Back off duration is calculated
+    /// back_off_interval * 2^back_off_exponent. Default: 4.
+    pub max_back_off_exponent: u32,
```
ThetaSinner, line 114: That gives a maximum back-off of 32 minutes right? I wonder if there's a library we could use here because this a fiddly thing to implement

Ideally we'd back off by a few seconds and probably not by more than say 5 minutes

crates/core/src/factories/core_fetch.rs
```diff
@@ -228,22 +236,19 @@ impl CoreFetch {
...
 
-                // Do nothing if op id is no longer in the set of ops to fetch.
-                if !lock.ops.contains(&(op_id.clone(), agent_id.clone())) {
+                // Do nothing if op id is no longer in the set of requests to send.
+                if !lock.requests.contains(&(op_id.clone(), agent_id.clone())) {
                     continue;
                 }
 
-                lock.cool_down_list.is_agent_cooling_down(&agent_id)
+                lock.back_off_list.is_agent_backing_off(&agent_id)
```
ThetaSinner, line 247: Slightly confusing method name because we've put the agent on back off, they aren't doing the back off. Perhaps `is_agent_on_back_off`? Or `should_back_off_agent` or something like that?

crates/core/src/factories/core_fetch.rs
```diff
@@ -348,40 +352,53 @@ impl Drop for CoreFetch {
...
+                    < (self.back_off_interval * 2_u64.pow(*exponent)) as u128
             }
             None => false,
         }
     }
+
+    pub fn is_agent_at_max_back_off(&self, agent_id: &AgentId) -> bool {
+        self.state
+            .get(agent_id)
+            .map(|v| v.1 == self.max_back_off_exponent)
```
ThetaSinner, line 395: Slightly confusing here too, because we'll actually only wait for delay * 2^(exponent - 1) as the last back off. The agent will report being at max back off before they have necessarily waited for the whole period. I think the logic actually works here but I'd be more confident if this was a check that the final window had expired. 

Though, see the comment above about a library that does this?

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -374,11 +374,46 @@ async fn ops_are_cleared_when_agent_not_in_peer_store() {
...
+
+#[test]
+fn back_off() {
+    let back_off_interval_ms = 10;
+    let mut back_off_list = BackOffList::new(back_off_interval_ms, 2);
+    let agent_id = random_agent_id();
+    back_off_list.back_off_agent(&agent_id);
+    assert!(back_off_list.is_agent_backing_off(&agent_id));
+
+    std::thread::sleep(Duration::from_millis(back_off_interval_ms + 1));
```
ThetaSinner, line 388: Can this be done by manipulating the clock with Tokio?

jost-s: It can't. The function `is_agent_backing_off` would have to use `tokio::time::Instant::now()` too for this to work. `std::time::Instant`s are unaffected by it.

crates/core/src/factories/core_fetch.rs
jost-s: Changed default to 20 sec with exp 4. That gives us 20 sec initially and goes up to 11:40 min.

crates/core/src/factories/core_fetch.rs
jost-s: Yes, name changed to `is_agent_on_back_off`.

crates/core/src/factories/core_fetch.rs
jost-s: Fixed name and also check of whether the back off interval has elapsed.

jost-s:
@ThetaSinner Regarding using backon, I've started doing that and am happy to change to it. It is not much less custom code, because backon just gives us a series of durations. We still need to store at which back off interval the agent currently is, when the interval started and if all configured intervals have elapsed. I'll push an alternative branch with it tomorrow.

crates/core/src/factories/core_fetch.rs
```diff
@@ -276,38 +282,36 @@ impl CoreFetch {
...
-                            .collect();
+                        lock.back_off_list.back_off_agent(&agent_id);
+
+                        if lock
+                            .back_off_list
+                            .is_agent_at_max_back_off(&agent_id)
+                        {
+                            lock.requests
+                                .remove(&(op_id.clone(), agent_id.clone()));
+                        }
```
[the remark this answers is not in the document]
ThetaSinner: I'd expect to see a remove all for the ops here and taking the agent off the back off list?

If we're done waiting, then I think what we want is to drop all ops from this source and let the loop filter the incoming channel based on the op set. 

That means if the source comes back online then it can resume fetching before the next max timeout elapses. I'm also thinking we'll end up making a connection request for each outstanding op/source pair for this source at at the max backoff interval?

ThetaSinner (commented):
Question about terminating sources. Otherwise looking good.

If `backon` doesn't work as well as I hoped in this case then that's okay. Up to you whether that's adding value or just dependencies we can do without :)

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: Beautiful idea! Yes, now it makes sense to remove all ops of the unresponsive agent.
The agent could be removed from the list explicitly, but any successful request to them will do that anyway.

neonphog (approved):

jost-s:
superseded by #60

PR #57, opened 2024-12-17
Consider persisting tombstones

maackle:
Hey, your friendly neighborhood model checker here. Model checking showed me this seemingly unhandled case and I was wondering if you had considered it:

It looks to me like tombstones don't persist, they just remove an entry and "expire" immediately. If another entry later comes in which predates the tombstone, it is allowed to be stored. This test fails.

If this is true, and the current behavior is desired, it might be good as an extra line in the spec, since the usual meaning of "tombstone" would imply that this test should pass.

### DIFF
--- a/crates/bootstrap_srv/src/test.rs
+++ b/crates/bootstrap_srv/src/test.rs
@@ -253,6 +253,33 @@ fn tombstone_old_is_ignored() {
 }
 
 #[test]
+fn tombstone_persists() {
+    let s = BootstrapSrv::new(Config::testing()).unwrap();
+
+    let _ = PutInfo {
+        addr: s.listen_addrs()[0],
+        is_tombstone: true,
+        ..Default::default()
+    }
+    .call()
+    .unwrap();
+
+    let _ = PutInfo {
+        addr: s.listen_addrs()[0],
+        created_at: now()
+            - std::time::Duration::from_secs(60).as_micros() as i64,
+        ..Default::default()
+    }
+    .call()
+    .unwrap();
+
+    let addr = format!("http://{:?}/bootstrap/{}", s.listen_addrs()[0], S1);
+    let res = ureq::get(&addr).call().unwrap().into_string().unwrap();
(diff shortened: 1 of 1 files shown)


neonphog (changes requested):
The tombstones are persisted in the gossiped peer stores of the individual nodes. This is why the term tombstone is in the data structure.

We have to sign those and send them anyway. The bootstrap server needs a way of verifying that the same client which published a non-tombstone entry is the one that is requesting a deletion of that entry. Rather than coming up with a totally parallel set of data to sign, we're just using the gossip tombstone entry as the delete mechanism in the bootstrap server.

If it becomes a real problem that legitimate nodes are copying pre-tombstone entries and re-publishing them to grief systems in the short time before they expire we can add some extra cache to maintain tombstones on the server... but that will come with downsides of memory usage. We don't want to push out other non-tombstone entries from the space memory.

I'm happy to merge some docs to this affect as suggested into the spec.

maackle:
That design decision makes sense. Yeah, a line in the spec sounds good, for completeness, and to keep in mind the potential griefing vector.

crates/bootstrap_srv/src/lib.rs
```diff
@@ -45,6 +45,16 @@
 ///
 /// Any other properties on these objects will be ignored and pass through.
 ///
+/// A note on the "isTombstone" field: This is named "tombstone" because the
+/// p2p gossip this bootstrap service is designed to facilitate must keep
```
ThetaSinner, line 49:
```suggestion
/// p2p gossip this bootstrap service must keep
```

ThetaSinner (dismissed):

crates/bootstrap_srv/src/lib.rs
neonphog: Attempted to fix this via: https://github.com/holochain/kitsune2/pull/57/commits/6e482e3d231d14b62e108016c1240bfb9c2400e3, let me know if that works.

neonphog (approved):

ThetaSinner (approved):

PR #58, opened 2024-12-18
Config V2

neonphog:
- Does away with the `ModConfig` trait.
- We now just take anything Serializable for the `set_module_config` api.
- and can get anything that implements DeserializeOwned for the `get_module_config` api.
- Now supports runtime config alteration through `register_entry_update_cb` api.
- And adds some warnings for "this param may be unused" and "nobody's listening to this runtime update".

### DIFF
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -77,7 +77,7 @@ impl Config {
         let mut updates = Vec::new();
         {
             let mut lock = self.0.lock().unwrap();
-            let old_map: &mut ConfigMap = &mut *lock;
+            let old_map: &mut ConfigMap = &mut lock;
             let new_map: &ConfigMap = &in_map;
             fn apply_map(
                 updates: &mut Vec<(ConfigUpdateCb, serde_json::Value)>,
@@ -101,7 +101,7 @@ impl Config {
                     },
                     ConfigMap::ConfigEntry(new_entry) => match old_map {
                         ConfigMap::ConfigMap(m) => {
-                            if m.len() > 0 {
+                            if !m.is_empty() {
                                 return Err(K2Error::other(
                                     "attempted to insert an entry where a map exists",
                                 ));
@@ -142,7 +142,7 @@ impl Config {
     ) -> K2Result<()> {
         let value = {
             let mut lock = self.0.lock().unwrap();
-            let mut cur: &mut ConfigMap = &mut *lock;
+            let mut cur: &mut ConfigMap = &mut lock;
             for path in path {
                 let key = path.to_string();
                 match cur {
@@ -156,7 +156,7 @@ impl Config {
             }
             match cur {
                 ConfigMap::ConfigMap(m) => {
-                    if m.len() > 0 {
+                    if !m.is_empty() {
                         return Err(K2Error::other(
                             "attempted to insert an entry where a map exists",
                         ));
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -46,23 +46,61 @@ impl Default for ConfigMap {
     }
 }
 
+struct Inner {
+    map: ConfigMap,
+    are_defaults_set: bool,
+    is_runtime: bool,
+}
+
 /// Kitsune configuration.
-#[derive(Debug, serde::Serialize)]
-pub struct Config(Mutex<ConfigMap>);
+pub struct Config(Mutex<Inner>);
+
+impl serde::Serialize for Config {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.0.lock().unwrap().map.serialize(serializer)
+    }
+}
+
+impl std::fmt::Debug for Config {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.lock().unwrap().map.fmt(f)
+    }
+}
 
 impl Default for Config {
     fn default() -> Self {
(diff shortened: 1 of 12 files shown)


crates/api/src/config.rs
```diff
@@ -14,68 +15,236 @@ fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
...
+#[serde(transparent, rename_all = "camelCase")]
+struct ConfigEntry {
+    pub value: serde_json::Value,
+    #[serde(skip, default)]
+    pub update_cb: Option<ConfigUpdateCb>,
+}
+
+impl std::fmt::Debug for ConfigEntry {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.value.fmt(f)
```
ThetaSinner, line 32: Might be nice to include the presence or absence of the `update_cb` even though it's not meaningful to log the value?

crates/api/src/config.rs
```diff
@@ -14,68 +15,236 @@ fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
...
+impl Default for ConfigMap {
+    fn default() -> Self {
+        Self::ConfigMap(BTreeMap::new())
+    }
+}
+
+struct Inner {
+    map: ConfigMap,
+    are_defaults_set: bool,
+    is_runtime: bool,
```
ThetaSinner, line 52: Does this mean that the runtime logic is module scoped? So "you can replace fetch config at runtime" but "you cannot replace space config at runtime"?

If I'm understanding that right, I think this belongs at the `ConfigEntry` level?

ThetaSinner: Or maybe I'm misunderstanding this field because an update handler would have to be registered per config entry for it to be possible to update it anyway. So is this just about "some values in this map can be updated"?

crates/api/src/config.rs
```diff
@@ -14,68 +15,236 @@ fn tc<S: serde::Serialize, D: serde::de::DeserializeOwned>(
...
-    /// for all of its modules with an immutable reference to this config
-    /// struct. Each of those modules may choose to call this function
-    /// to extract a module config. Note that this config is loaded from
-    /// disk and can be edited by humans, so the serialization on the module
-    /// config should be tolerant to missing properties, setting sane defaults.
-    /// A module may choose to warn about missing properties and should warn about extraneous properties.
-    pub fn get_module_config<M: ModConfig>(
+    /// Call this in your module constructor once for every parameter for
+    /// which you would like to receive runtime updates. This will immediately
+    /// invoke the callback with the current value to ensure this is atomic.
```
ThetaSinner, line 205:
```suggestion
    /// invoke the callback with the current value to ensure this is idempotent.
```

?

crates/api/src/config.rs
```diff
@@ -84,100 +253,91 @@ mod test {
...
-        // ensure we can load a weird config from disk
-        let config: Config = serde_json::from_str(
-            r#"{
-          "modBAD": { "foo": "bar" },
-          "mod1": { "p_b": "test-p_b" },
-          "mod2": { "p_c": 42, "p_d": "test-p_d", "extra": "foo" }
-        }"#,
+        println!("{}", serde_json::to_string_pretty(&c).unwrap());
+
+        let resp: ModConfig = c.get_module_config().unwrap();
```
ThetaSinner, line 317: Not quite sure I'm following this part. This is actually transcoding from the internal format to the `ModConfig` struct here?

If so, I wonder if we want to try to be a bit more efficient than calling `get_module_config` always in modules that can have their config change. We always want to use the latest config but don't want to do the deserialize every time we need config.

ThetaSinner (commented):
This is good, I like where it's going, just a few questions. Partly to make sure I'm following the code :)

crates/api/src/config.rs
neonphog: This is purely about providing a warning if a config is updated that doesn't have an update callback registered. So we mark this when we switch to runtime, and thereafter the warning traces are emitted.

crates/api/src/config.rs
neonphog: No, I meant atomic. Because another thread could be setting the value at the time you're registering the update cb. If it immediately calls it with the current value when registering the update callback you have the option to *only* pay attention to values sent through the callback.

crates/api/src/config.rs
neonphog: The intention is in general to use `get_module_config` the normal intuitive way, most modules will just get it at the beginning and never bother with updates.

If your module has a config that wants to change at runtime, it should register an update callback which will receive the value directly.

crates/api/src/config.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/58/commits/eb1137054b4208e7ba66e65f3d20bd4e8acd9c0a

crates/api/src/config.rs
ThetaSinner: Got it, cool okay!

crates/api/src/config.rs
ThetaSinner: I see! Right yeah I hadn't thought about it like that. Cool, yeah this makes sense then

crates/api/src/config.rs
ThetaSinner: Cool okay, so get the config once and put it somewhere then your update functions can update your inner config. Great, happy with that

ThetaSinner (approved):
Thank you for the answers to my many questions :) Looks good!

PR #59, opened 2024-12-18
Dht diff

ThetaSinner:
Top level DHT model and diff comparison to figure out which ops might be missing from the other.

I know there's a lot of code addition on this PR. I think the best place to start is probably the module documentation, then following that into the entry functions like `snapshot_minimal`. That should be good for a high-level understanding I hope. Then the test harness `crates/dht/tests/harness.rs` is a good place to understand how the sync is done. That code will be more or less what gossip needs to do to use the code under test.

If it helps, there is +1574 just from tests. So only half of this is actually new production code.

Following this PR I think there's a need to go back over the code in this crate and ensure consistent naming and documentation. I think the new documentation is consistent but I don't think it's entirely consistent with what's in here from previous PRs.


crates/dht/src/arc_set.rs
```diff
@@ -0,0 +1,278 @@
...
+//! DHTs can be intersected to find the common sectors, using [ArcSet::intersection].
+
+use kitsune2_api::{DhtArc, K2Error, K2Result};
+use std::collections::HashSet;
+
+/// Represents a set of [DhtArc]s.
+#[derive(Debug)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct ArcSet {
+    inner: HashSet<u32>,
```
neonphog, line 15: hehe : )

crates/dht/src/dht/snapshot.rs
```diff
@@ -0,0 +1,763 @@
...
+                    return SnapshotDiff::DiscMismatch;
+                }
+
+                // This is more common, it can happen if we're close to a UNIT_TIME boundary
+                // and there is a small clock difference or just one node calculated this snapshot
+                // before the other did. Still, we'll have to wait until we next compare to
+                // compare our DHT state.
+                if our_ring_top_hashes.len() != other_ring_top_hashes.len() {
+                    return SnapshotDiff::CannotCompare;
+                }
```
neonphog, line 109: I assume, if these are off by one because one of the nodes moved forward, we cannot just ignore that extra, because the remaining rings may be re-sized so the hashes won't match?

What is the effect of returning a CannotCompare do you imagine? Will the node's just retry after a bit when they should again be in the same time-state?

ThetaSinner: Exactly that, if the times are off by enough for the rings to change then we can't work with indices safely. I believe the representation of partial time slices is unique so if you pin the end of the full time slices and check the number of partial slices, I think you can be sure you're comparing the same time slices.

Yes I think it has to terminate the sync process early. So we'll still do gossip of agents and "what's new since we last gossiped" and then this part would have to fail if the model doesn't match.

If it gets updated mid round that would kill the round so at least for rounds we've initiated, I think we should avoid doing that during the round. Right before we initiate would be a good time. Then rounds initiated by others might then fail but the model only needs updating every UNIT_TIME which is 15 minutes, so even if we only update every 5 minutes, there has to be a pretty significant clock skew for a mismatch to happen I think

crates/dht/src/arc_set.rs
ThetaSinner: Yes it's a little odd, but it's using sector id values so it's size limited to the number of sectors we chose which is 512 and having set operations makes life a little easier... I could probably simplify the constructor thinking about it actually. I'll leave that alone for now

neonphog: All good. It's perfect for a first implementation. It totally may never need to be optimized further.

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,425 @@
...
+    /// provided back to this function in the `our_previous_snapshot` parameter. In all other cases,
+    /// the caller should provide `None` for `our_previous_snapshot`.
+    ///
+    /// Note also that there are two possible routes through the comparison process. The first is
+    /// when the historical disc mismatches, the second is when the recent rings mismatch. The
+    /// historical disc mismatch is prioritised, so if a mismatch is detected there then the sync
+    /// process will resolve that. Otherwise, the recent rings mismatch will be resolved. That means
+    /// that it may take up to two rounds of sync to resolve all mismatches. Of course, both the
+    /// disc and the rings must be considered a moving target so it cannot be assumed that 2 rounds
+    /// are actually enough to resolve all mismatches.
```
neonphog, line 165: Is the idea that, when we complete a disc mismatch round, we go into a second round with no delay to see if we need to resolve any ring mismatches?

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,425 @@
...
+    ///
+    /// The `arc_set` parameter is used to determine which arcs are relevant to the DHT model. This
+    /// should be the [ArcSet::intersection] of the arc sets of the two DHT models to be compared.
```
neonphog, line 168: Do we need a short-circuit here or somewhere else to abort even gossiping if there is no overlap to the sets? Or do we just assume that was checked before even getting to this point?

crates/dht/src/dht/snapshot.rs
```diff
@@ -0,0 +1,763 @@
...
+
+use kitsune2_api::Timestamp;
+use std::collections::{HashMap, HashSet};
+
+/// A snapshot of the DHT state at a given point in time.
+///
+/// This is largely opaque to the user of the [crate::dht::Dht] model. It is intended to be sent
+/// between nodes to compare their DHT states and compared with [DhtSnapshot::compare].
+#[derive(Debug, Eq, PartialEq)]
+pub enum DhtSnapshot {
```
neonphog, line 15: `Disc -> Past, Ring -> Recent`

Consider this a very light suggestion. I like that we already have terminology similar to this in legacy Kitsune. But, I'm actually fine with the Disc/Ring names as long as we document them well and are consistent about them.

Other things I considered:
- `Disc -> Old, Ring -> New` - Working with data, I like to use the prefix `old_`, and wouldn't want to see `old_old_`.
- `Disc -> Historical? History?, Ring -> Recent` - Past is shorter.

```rust
pub enum DhtSnapshot {
    Minimal {
        past_top_hash: bytes::Bytes,
        past_boundary: Timestamp,
        recent_top_hashes: Vec<bytes::Bytes>,
    },
    PastSectors {
        past_sector_top_hashes: HashMap<u32, bytes::Bytes>,
        past_boundary: Timestamp,
    },
    PastSectorDetails {
        past_sector_hashes: HashMap<u32, HashMap<u64, bytes::Bytes>>,
        past_boundary: Timestamp,
    },
    RecentSectorDetails {
        recent_sector_hashes: HashMap<u32, HashMap<u32, bytes::Bytes>>,
        past_boundary: Timestamp,
    },
}
```

crates/dht/src/dht/tests.rs
```diff
@@ -0,0 +1,542 @@
+mod harness;
```
neonphog, line 1: Maybe add a couple tests for the `CannotCompare` cases?

neonphog: Nevermind, I see those in the snapshot tests.

neonphog (dismissed):
Couple questions including the potential name change, but the code here looks good. Nice work :+1:

crates/dht/src/dht/snapshot.rs
neonphog:
```
         |-----Past-----|Recent-|
u32::MAX +----+----+----+--+--+-+-------
         |    |    |    |  |  | | Sector
L        +----+----+----+--+--+-+-------
o        |    |    |    |  |  | |
c        +----+----+----+--+--+-+
a        |    |    |    |  |  | |
t        +----+----+----+--+--+-+
i        |    |    |    |  |  | |
o      0 +----+----+----+--+--+-+
n        0         Time         Now
```

jost-s (commented):
More tomorrow

crates/dht/src/dht.rs
ThetaSinner: Not unless we make a modification to not check the `disc` here by keeping our minimal snapshot around and using it to create a ring diff or something like that because until the ops have been fetched, we'll like get another ring diff that's the same as we just got if we go straight back into another sync. I was just thinking that would have to happen on the next gossip round.

Doing it how I was thinking means that we could get stuck syncing historical data if more keeps showing up and not get to recent data... Until that data becomes historical. I'm relying on some "reasonable" behavior here where historical data showing up isn't a constantly occurring thing.

crates/dht/src/dht.rs
ThetaSinner: Oh this is a good point. In the previous Kitsune we took all our local agents and merged them then checked the overlap with the remote agent to gossip with. If the overlap was empty then they wouldn't be selected as a gossip target.

Now we're actually decomposing this we should make an effort to defend and document our interface to make sure it's getting used right. I've added an error if this is called with an empty arc, plus tests and documented the error :)

crates/dht/src/dht/snapshot.rs
ThetaSinner: I'd be happy with Historical/Recent. As you say, that's similar to terminology we already have and they make a nice pair.

The bit I'd be losing slightly and have to come up with some other wording for is that the Historical starts out singular because we compress that right down to one hash. The Recent starts out pluralized because it's a more predictable size and we want to diff it in fewer steps. I've already had to do that with adding `_sectors` and then `_sector_details`, so that's nothing new. Just needs a bit of thought.

Another option, because ring is not a geometry term and that's part of the confusion here because it's too similar to disc. The correct term would be [annulus](https://en.wikipedia.org/wiki/Annulus_(mathematics))... But I think there's some potential for misspellings and slips of the tongue trying to get the plural "annuli" right :)

ThetaSinner: I'll wait for some more votes for options here before I start a big rename, it'll be a bit of a project :)

crates/dht/src/combine.rs
```diff
@@ -0,0 +1,47 @@
...
+pub fn combine_hashes(into: &mut bytes::BytesMut, other: bytes::Bytes) {
+    // Properly initialise the target from the source if the target is empty.
+    // Otherwise, the loop below would run 0 times.
+    if into.is_empty() && !other.is_empty() {
+        into.extend_from_slice(&other);
+        return;
+    }
+
+    if into.len() != other.len() {
+        tracing::debug!("Combining hashes of different lengths ({} != {}). This is undefined behaviour.", into.len(), other.len());
```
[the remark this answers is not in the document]
ThetaSinner: Good point yes. I think I was doing this in tests intentionally and didn't want to see the warning but I should just fix that and this is actually quite bad if it happens. It means we haven't recalculated hashes of incoming data or have a problematic host implementation or some other bug

crates/dht/src/dht/snapshot.rs
matthme: > ring is not a geometry term

It is in German...it's called "[Kreisring](https://de.wikipedia.org/wiki/Kreisring)". Maybe just just need to add somewhere in the inline docs that it's the thing that's supposed to be thrown into volcanoes rather than the thing that keeps Elvis alive if you put it on a record player ;).

ThetaSinner: That's a much better name, a "circle ring"!

Technically a top hash over the set of rings would represent a one ring... :D

matthme: I think I'm derailing too far into non-technical terminology but it could also be considered as "flat donut" vs. a "frisbee". Or a 2d-torus?

mattyg: I'm fine with disc/ring, or past/recent. Another option would that is maybe more consistent would be full/partial. The comments make it clear enough that I don't feel strongly. I assuming these terms are contained within this module anyway, right? Or will they leak out?

crates/dht/src/combine.rs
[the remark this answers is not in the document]
jost-s: Actually what about panicking? It would be an internal bug after all.

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,447 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there are no arcs to snapshot. If there is no overlap between the arc
+    /// sets of two DHT models then there is no point in comparing them because it will always
+    /// yield an empty diff. The [ArcSet::covered_sector_count] should be checked before calling
```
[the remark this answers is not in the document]
ThetaSinner: I guess that's the caller's choice. I'd prefer to avoid an error that I can only convert to a string to tell it apart from other errors. Or even with strong types, if I know an input won't work for a function, I'd prefer to filter it out.

Gossip should be choosing other nodes with agent arcs that overlap with ours and the node accepting a gossip round should be checking what the initiator thought was the arc overlap so there's really no reason for an empty ArcSet to get this far. I added the error just to be doubly sure we find out about mistakes elsewhere because it'll be across a crate boundary :)

crates/dht/src/dht/snapshot.rs
ThetaSinner: They're visible as enum variants but nothing needs to operate on them, just check for a "*Details" enum variant and cache that but otherwise the enum doesn't need to be inspected. The fields will have to be sent over gossip so the gossip protocol will use the sane names... it leaks :)

crates/dht/src/dht/tests.rs
```diff
@@ -0,0 +1,588 @@
...
+    // Try a sync
+    let outcome = dht1.sync_with(&mut dht2).await.unwrap();
+    assert!(matches!(outcome, SyncWithOutcome::SyncedDisc));
+
+    // We shouldn't learn about any ops
+    assert!(dht1.discovered_ops.is_empty());
+
+    // The other agent should have learned about the op
+    assert_eq!(1, dht2.discovered_ops.len());
+    assert_eq!(1, dht2.discovered_ops[&dht1.agent_id].len());
```
jost-s, line 134: Directly assert that discovered ops is vec![op_id] from above?

crates/dht/src/dht/tests.rs
```diff
@@ -0,0 +1,588 @@
...
+
+    // Move any discovered ops between the two DHTs
+    dht1.apply_op_sync(&mut dht2).await.unwrap();
```
jost-s, line 137: Assert that before this operation dht1 is not in sync with dht2?

crates/dht/src/dht/tests.rs
jost-s, line 1: Love this test suite!

crates/dht/src/dht/snapshot.rs
jost-s, line 1: Love this too

jost-s (commented):
Great looking code! I made some suggestions.

crates/dht/src/dht/snapshot.rs
```diff
@@ -0,0 +1,763 @@
...
+                // This shouldn't be common, but we sync forwards through time, so if we
+                // find a historical mismatch then focus on fixing that first.
+                if our_disc_top_hash != other_disc_top_hash {
+                    return SnapshotDiff::DiscMismatch;
+                }
+
+                // This is more common, it can happen if we're close to a UNIT_TIME boundary
+                // and there is a small clock difference or just one node calculated this snapshot
+                // before the other did. Still, we'll have to wait until we next compare to
+                // compare our DHT state.
```
jost-s, line 106:
```suggestion
                // before the other did. Still, we'll have to wait until we next compare to
                // our DHT state.
```

crates/dht/src/dht/tests/harness.rs
```diff
@@ -0,0 +1,429 @@
...
+impl DhtSyncHarness {
+    pub(crate) async fn new(current_time: Timestamp, arc: DhtArc) -> Self {
+        let store = Arc::new(Kitsune2MemoryOpStore::default());
+        let dht = Dht::try_from_store(current_time, store.clone())
+            .await
+            .unwrap();
+
+        let mut bytes = [0; 32];
+        rand::thread_rng().fill_bytes(&mut bytes);
+        let agent_id = AgentId::from(bytes::Bytes::copy_from_slice(&bytes));
```
jost-s, line 40: There's a `random_agent_id` fn in the `core_fetch` tests which I invite you to avail of here.

crates/dht/src/dht/tests.rs
```diff
@@ -0,0 +1,588 @@
...
+    // Try to sync the DHTs
+    let outcome = dht1.sync_with(&mut dht2).await.unwrap();
+    assert!(matches!(outcome, SyncWithOutcome::SyncedRings));
+
+    // Sync the ops
+    dht1.apply_op_sync(&mut dht2).await.unwrap();
+
+    // Now the DHTs should be in sync
+    assert!(dht1.is_in_sync_with(&dht2).await.unwrap());
+}
```
jost-s, line 588: I think something like an initial gossip session that syncs down to the level of hashes would be good to add. Also to get a feel for how that'll work. What do you think?
```suggestion
}

#[tokio::test]
async fn disc_sync() {
    let current_time = Timestamp::now();
    let store_1 = Arc::new(Kitsune2MemoryOpStore::default());
    let mut dht1 = Dht::try_from_store(current_time, store_1.clone())
        .await
        .unwrap();
    let store_2 = Arc::new(Kitsune2MemoryOpStore::default());
    let dht2 = Dht::try_from_store(current_time, store_2.clone())
        .await
        .unwrap();

    let arc_set_1 = ArcSet::new(SECTOR_SIZE, vec![DhtArc::FULL]).unwrap();
    let arc_set_2 = ArcSet::new(SECTOR_SIZE, vec![DhtArc::FULL]).unwrap();

    let op_id_1 = OpId::from(bytes::Bytes::from_static(&[0]));
    let op_1 = Kitsune2MemoryOp::new(op_id_1.clone(), UNIX_TIMESTAMP, vec![]);
    store_1
        .process_incoming_ops(vec![op_1.clone().try_into().unwrap()])
        .await
        .unwrap();
    dht1.inform_ops_stored(store_1.clone(), vec![op_1.into()])
        .await
        .unwrap();

    let dht_minimal_1 = dht1
        .snapshot_minimal(&arc_set_1, store_1.clone())
        .await
        .unwrap();

    let next_action_1 = dht2
        .handle_snapshot(&dht_minimal_1, None, &arc_set_2, store_2.clone())
        .await
        .unwrap();

    let snapshot_disc_sectors_2 = match next_action_1 {
        DhtSnapshotNextAction::NewSnapshot(snapshot) => snapshot,
        _ => panic!("ohono"),
    };

    let next_action_2 = dht1
        .handle_snapshot(
            &snapshot_disc_sectors_2,
            Some(dht_minimal_1),
            &arc_set_1,
            store_1.clone(),
        )
        .await
        .unwrap();

    let snapshot_disc_sector_details_1 = match next_action_2 {
        DhtSnapshotNextAction::NewSnapshot(snapshot) => snapshot,
        _ => panic!("ohono"),
    };

    let next_action_1 = dht2
        .handle_snapshot(
            &snapshot_disc_sector_details_1,
            Some(snapshot_disc_sectors_2),
            &arc_set_2,
            store_2.clone(),
        )
        .await
        .unwrap();

    let snapshot_and_hash_list_2 = match next_action_1 {
        DhtSnapshotNextAction::NewSnapshotAndHashList(snapshot, hashes) => {
            (snapshot, hashes)
        }
        _ => panic!("ohono"),
    };

    let next_action_2 = dht1
        .handle_snapshot(
            &snapshot_and_hash_list_2.0,
            Some(snapshot_disc_sector_details_1),
            &arc_set_1,
            store_1.clone(),
        )
        .await
        .unwrap();

    let hashes = match next_action_2 {
        DhtSnapshotNextAction::HashList(hashes) => hashes,
        _ => panic!("ohono"),
    };

    assert_eq!(hashes, vec![op_id_1]);
}
```

jost-s (commented):
Some more comments

crates/dht/src/combine.rs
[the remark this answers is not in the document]
ThetaSinner: That's an option if we are really careful about checking incoming hashes. The only data getting pushed into this model should already have been stored by the host and had its hash recalculated and checked. If it's being combined here then it really ought to be valid. Okay, I'll go for that

crates/dht/src/hash.rs
```diff
@@ -190,6 +197,238 @@ impl PartitionedHashes {
...
+    /// Along with the disc top hash, the end timestamp of the last full time slice is returned.
+    /// This should be used when comparing disc top hash of one DHT model with that of another node
+    /// to ensure that both nodes are using a common reference point.
+    pub(crate) async fn disc_top_hash(
+        &self,
+        arc_set: &ArcSet,
+        store: DynOpStore,
+    ) -> K2Result<(bytes::Bytes, Timestamp)> {
+        let mut combined = bytes::BytesMut::new();
+        for (sector_id, sector) in self.partitioned_hashes.iter().enumerate() {
```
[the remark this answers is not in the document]
ThetaSinner: Makes sense, done across the 3 places I've used `_id` that are referring to an index

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,447 @@
...
+
+    /// Handle a snapshot from another DHT model.
+    ///
+    /// This is a two-step process. First the type of the incoming snapshot is checked and a
+    /// snapshot of the same type is computed. Secondly, the two snapshots are compared to determine
+    /// what action should be taken next.
+    ///
+    /// The state flow is as follows:
+    /// - If the two snapshots are identical, the function will return [DhtSnapshotNextAction::Identical].
+    /// - If the two snapshots cannot be compared, the function will return [DhtSnapshotNextAction::CannotCompare].
```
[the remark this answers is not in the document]
ThetaSinner: That happens when the DHT models have been updated at different times by us and the remote. If our clocks are reasonably close together (within a minute or two) and we're updating when the model says it's due then we should be getting this rarely. But because we're relying on indices, if the models are out of step then they're difficult to compare. We could check historical time slices but beyond that we're stuck so I'm expecting a `CannotCompare` to terminate the current gossip round.

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,447 @@
...
+    ///   This can happen if the time slices of the two DHTs are not aligned or one side is following
+    ///   a different flow to what we're expecting.
+    /// - If the snapshots are different, the function will return [DhtSnapshotNextAction::NewSnapshot]
+    ///   with a more detailed snapshot of the DHT model.
+    /// - When the most detailed snapshot type is reached, the function will return [DhtSnapshotNextAction::NewSnapshotAndHashList]
+    /// - The new snapshot from [DhtSnapshotNextAction::NewSnapshotAndHashList] should be sent to the
+    ///   other party so that they can compare it with their own snapshot and determine which op hashes
```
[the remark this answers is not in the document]
ThetaSinner: Oh but even that wouldn't happen in this code because we deliberately don't have access to the `DhtSnapshotNextAction` from the other peer. I think comparing hashes and deciding what to do with them belongs in gossip. This just details with finding where the differences are and storing the DHT data in a structure. The logic for coordinating the updates and working out what to send over the network I'd prefer to keep in the gossip module

crates/dht/src/dht/tests.rs
ThetaSinner: https://github.com/holochain/kitsune2/pull/59/commits/f0d08bc0e054917b3d7e4e015db1aefc5d0043eb

crates/dht/src/dht/tests.rs
ThetaSinner: Good call, done https://github.com/holochain/kitsune2/pull/59/commits/4c67a761f07461ca9003e6ff320004e873585027

crates/dht/src/arc_set.rs
```diff
@@ -0,0 +1,316 @@
...
+}
+
+impl ArcSet {
+    /// Create a new arc set from a list of arcs.
+    ///
+    /// The resulting arc set represents the union of the input arcs.
+    pub fn new(size: u32, arcs: Vec<DhtArc>) -> K2Result<Self> {
```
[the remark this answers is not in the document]
ThetaSinner: Done https://github.com/holochain/kitsune2/pull/59/commits/b4fa680bb53d40b114e796a9235b461d3089aa15

Though I think I probably want to tweak this API when it comes to integrating it. I don't really want gossip having to provide this value. Ideally it'd be internal to the DHT module while the value is hard-coded. I think hiding this constructor and providing a public one without the size parameter is probably the way to go

crates/dht/src/dht.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// This is the final step in the comparison process. The yielded op hashes should be fetched
+    /// from the other party. No further snapshots are required for this comparison.
+    HashList(Vec<OpId>),
+}
+
+impl Dht {
+    /// Create a new DHT from an op store.
+    ///
+    /// Creates the inner [PartitionedHashes] using the store.
+    pub async fn try_from_store(
```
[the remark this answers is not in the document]
ThetaSinner: I've done the same thing with the other types. The intention was to be clear with what functions need that state and which don't. So it's clear when using this module when the global op state is relevant and when you're just working with in-memory state. I can see that makes less sense with this type though because it's always across the full data set so most operations require the store.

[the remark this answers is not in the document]
ThetaSinner: Done https://github.com/holochain/kitsune2/pull/59/commits/01f402cc7ae936008de8f23c21d1ab49b1765774

crates/dht/src/dht/tests.rs
ThetaSinner: I agree but I think this logic is being tested, it's just not as clear because it's happening in the harness. The `hashes` you're finding are stored in `discovered_ops`in the harness. So for example the test `two_stage_sync_with_symmetry` is doing what your suggestion is doing. It's checking that we can find ops in the disc and in the rings. It's also checking that those ops are identified exactly the same way regardless of which side initiated the gossip.

Anything I'm missing there?

jost-s: No, that's fine with me.

jost-s: I wanted to do the exercise of writing this to see how ergonomic and intelligible the API is. All good from my end.

jost-s (approved):
Very happy with this code!

PR #60, opened 2024-12-20
Back off improvement

jost-s:
Alternative to #56 

closes #56


ThetaSinner (approved):
Nice! I find that logic much easier to read even though the library is only hiding a little bit. I can forget about the maths and details and just focus on the logic

jost-s:
@ThetaSinner Moved back off to a separate module. The fetch module was getting too long.

ThetaSinner (approved):

PR #61, opened 2024-12-20
Bootstrap and Re-Sign Agent Infos Before they Expire

neonphog:


crates/core/src/factories/core_space.rs
```diff
@@ -5,6 +5,55 @@ use std::sync::{Arc, Mutex, Weak};
...
+    pub struct CoreSpaceConfig {
+        /// The interval in millis at which we check for about to expire
+        /// local agent infos.
+        /// Default: 60s.
+        pub re_sign_freq_ms: u32,
+
+        /// The time in millis before an agent info expires after which we will
+        /// re sign them.
+        /// Default: 5m.
+        pub re_sign_expire_time_ms: u32,
```
jost-s, line 22: What is this for? If an agent expires within this period of time and the check task runs, they'll send their info to the bootstrap again. If an agent expires after this period, they won't send the info. Why?

neonphog: No, it's the other way around. This is making sure we wait *at least* until five minutes before expiry before re-signing a new one.

jost-s: Right yes, I wrote it wrongly. What's the purpose of this? Or configuring this?

crates/core/src/factories/core_space.rs
```diff
@@ -5,6 +5,55 @@ use std::sync::{Arc, Mutex, Weak};
...
 
+/// CoreSpace configuration types.
+pub mod config {
+    /// Configuration parameters for [CoreSpaceFactory](super::CoreSpaceFactory).
+    #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct CoreSpaceConfig {
+        /// The interval in millis at which we check for about to expire
+        /// local agent infos.
+        /// Default: 60s.
```
ThetaSinner, line 16:
```suggestion
        ///
        /// Default: 60s.
```

crates/core/src/factories/core_space.rs
```diff
@@ -5,6 +5,55 @@ use std::sync::{Arc, Mutex, Weak};
...
+        pub re_sign_freq_ms: u32,
+
+        /// The time in millis before an agent info expires after which we will
```
ThetaSinner, line 19:
```suggestion
        /// The time in millis before an agent info expires, after which we will
```

crates/core/src/factories/core_space.rs
```diff
@@ -5,6 +5,55 @@ use std::sync::{Arc, Mutex, Weak};
...
+        /// re sign them.
+        /// Default: 5m.
```
ThetaSinner, line 21:
```suggestion
        ///
        /// Default: 5m.
```

crates/core/src/factories/core_space.rs
```diff
@@ -312,5 +399,44 @@ impl Space for CoreSpace {
...
+            .collect::<Vec<_>>();
+
+        for agent in agents {
+            // is this agent going to expire?
+            let should_re_sign = if let Ok(Some(info)) =
+                peer_store.get(agent.agent().clone()).await
+            {
+                info.expires_at <= cutoff
+            } else {
+                true
```
ThetaSinner, line 430: We re-sign on either `Err(_)` or `Ok(None)`, is that okay? I'd expect the error to get logged here at least?

crates/core/src/factories/core_space/test.rs
```diff
@@ -214,3 +214,94 @@ async fn space_notify_send_recv() {
...
+    .with_default_config()
+    .unwrap();
+
+    builder
+        .config
+        .set_module_config(&super::CoreSpaceModConfig {
+            core_space: super::CoreSpaceConfig {
+                // check every 5 millis if we need to re-sign
+                re_sign_freq_ms: 5,
+                // set it so we have to re-sign ALL agent infos always
```
ThetaSinner, line 286: This is setting the expire value to be equal to the period of time the agent info is valid for so that work always needs doing? Could the comment describe that, that the value is chosen relative to the agent info validity period?

ThetaSinner (commented):
Looking good! A couple of minor comments :)

crates/core/src/factories/core_space.rs
neonphog: Honestly? Because every time I just hard-code something people ask it to be in the config before approving.

ThetaSinner: Sorryyyyyy :) I think because it interacts with the check interval and the agent info expiry, it's probably sensible to have this in configuration but can see why it would be find to hard-code it too

crates/core/src/factories/core_space.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/61/commits/9afa6cc40e1072d76614481aa2eeb728a5def2b8

crates/core/src/factories/core_space.rs
jost-s: Those damn people!

jost-s (approved):

crates/core/src/factories/core_space/test.rs
neonphog: Let me know if this is better: https://github.com/holochain/kitsune2/pull/61/commits/93f7e03e60d661c77de075e94b763564312541ea

ThetaSinner: Awesome, yes thank you!

ThetaSinner (approved):

PR #63, opened 2024-12-31
Fetch response queue

### LINKED ISSUE #30: [k2] fetch response queue module
The fetch queue module is being created as part of https://github.com/holochain/holochain/issues/4479 to make requests for op content from remote peers.

Those remote peers must queue these requests (or send back error messages) so that they can respond with the op content in a measured sequential manner.

jost-s:
Adds a simple response queue to the fetch module which processes fetch requests one by one by loading requested ops from an op store and sending the available ops to the requester.

depends on #60 

resolves #30


crates/api/proto/fetch.proto
```diff
@@ -2,8 +2,16 @@ syntax = "proto3";
...
 // A list of op ids.
 message OpIds {
   // Op ids.
   repeated bytes data = 1;
 }
+
+// A list of ops.
+message Ops {
+  // Ops.
+  repeated kitsune2.op_store.Op data = 1;
```
jost-s, line 16: Re-use op definition.

crates/api/src/fetch.rs
```diff
@@ -34,17 +48,35 @@ pub fn serialize_op_ids(value: Vec<OpId>) -> bytes::Bytes {
 /// Deserialize list of op ids.
 pub fn deserialize_op_ids(value: bytes::Bytes) -> K2Result<Vec<OpId>> {
     let op_ids = OpIds::decode(value).map_err(K2Error::other)?;
-    let vec = Vec::from(op_ids);
-    Ok(vec)
+    Ok(op_ids.into())
+}
+
+/// Serialize list of ops for sending over the wire.
+pub fn serialize_ops(value: Vec<MetaOp>) -> bytes::Bytes {
```
jost-s, line 55: Very similar to the previous op id one. I could write this as a generic function which saves us code here but requires then types when calling the function.

crates/core/src/factories/core_fetch/back_off.rs
```diff
@@ -157,191 +147,4 @@ mod test {
 
         assert!(!back_off_list.is_agent_on_back_off(&agent_id));
     }
-
```
jost-s, line 160: These tests are just moved to the `request_queue.rs` file.

crates/core/src/factories/core_fetch/test.rs
jost-s, line 1: This file contains now shared util functions. Tests for the parts of the fetch module are split out into their own test files.

crates/core/src/factories/core_fetch/test/request_queue.rs
jost-s, line 1: No new test cases here, just updated API.

crates/api/proto/op_store.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package kitsune2.op_store;
+
+// An op.
+message Op {
+    // Op id.
+    bytes op_id = 1;
```
ThetaSinner, line 8: Want to question the value of sending the `op_id` with the `Op`. When we're communicating a diff, we have to work in terms of `op_id`s but when we have the `op_data`, we shouldn't trust the `op_id` that's been provided, we should calculate it from the `op_data`. This links to what we were discussing on the DHT diff PR where hash length mismatches would cause an error, we need to be properly checking these as they arrive.

crates/api/src/op_store.rs
```diff
@@ -70,6 +90,12 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
         end: Timestamp,
     ) -> BoxFuture<'_, K2Result<Vec<OpId>>>;
 
+    /// Read a list of ops from the store.
+    fn read_ops(
```
ThetaSinner, line 94: The DHT diff PR defines `retrieve_ops`, we'll have to resolve these two

crates/core/src/factories/core_fetch.rs
```diff
@@ -36,24 +38,38 @@
...
 //!   fetch request.
 //!
+//! #### Responses
+//!
+//! Similarly to fetch requests, a channel serves as a queue for responses to fetch requests. The queue
+//! has the following properties:
+//! - Simple queue which processes items in the order of the incoming requests.
+//! - Requests consist of a list of requested op ids and the requesting agent id.
+//! - Attempts to look up the op in the data store and send response are done once.
+//! - Requests for data that the remote doesn't hold should be logged.
```
ThetaSinner, line 48: This looks like it needs a second look? Is this "requests for data we don't have are logged"?

crates/core/src/factories/core_fetch.rs
```diff
@@ -36,24 +38,38 @@
...
+//! - If none of the requested ops could be read from the store, no response is sent.
+//! - If sending or reception fails, it's the caller's responsibility to request again.
```
ThetaSinner, line 50:
```suggestion
//! - If sending or receiving fails, it's the caller's responsibility to request again.
```

crates/core/src/factories/core_fetch.rs
```diff
@@ -335,6 +391,54 @@ impl CoreFetch {
...
+            response_rx.lock().await.recv().await
+        {
+            tracing::debug!(?op_ids, ?agent_id, "incoming request");
+            let peer = match CoreFetch::get_peer_url_from_store(
+                &agent_id,
+                peer_store.clone(),
+            )
+            .await
+            {
+                None => continue,
```
ThetaSinner, line 411: A comment here might be nice, I think it's pretty clear but something like the one below for the `is_empty` check would be good

crates/core/src/factories/core_fetch/back_off.rs
```diff
@@ -8,9 +8,9 @@ use kitsune2_api::AgentId;
 
 #[derive(Debug)]
 pub struct BackOffList {
-    state: HashMap<AgentId, BackOff>,
-    first_back_off_interval: Duration,
-    last_back_off_interval: Duration,
+    pub state: HashMap<AgentId, BackOff>,
```
ThetaSinner, line 11: Is this really `pub` or `pub(crate)`?

crates/core/src/factories/core_fetch/test/response_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl MockTransport {
+    fn new() -> Arc<MockTransport> {
+        Arc::new(Self {
+            responses_sent: Arc::new(Mutex::new(Vec::new())),
+        })
+    }
+}
+
+impl Transport for MockTransport {
```
ThetaSinner, line 33: Would mockall be appropriate here? Given we just want to work with one method on the type

crates/core/src/factories/core_fetch/test/response_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+        peer_store.clone(),
+        op_store,
+        mock_transport.clone(),
+    );
+
+    // Receive 2 op requests.
+    let requested_ops_1 = vec![op_id_1.clone(), op_id_2.clone()];
+    let requested_ops_2 = vec![op_id_3.clone(), op_id_4.clone()];
+    futures::future::join_all([
+        fetch.respond_with_ops(requested_ops_1, agent_id_1.clone()),
```
ThetaSinner, line 161: I wonder if `request_op_data` might be a clearer name for this? I can't quite get this into my head

crates/core/src/factories/core_fetch/test/response_queue.rs
```diff
@@ -0,0 +1,228 @@
...
+    // Only op 3 is in op store.
+    assert!(mock_transport
+        .responses_sent
+        .lock()
+        .unwrap()
+        .contains(&(vec![op_3], agent_url_2)));
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn no_response_sent_when_no_ops_found() {
```
ThetaSinner, line 191: Possible to also test agent info not in peer store?

ThetaSinner: and also it'd be great to test error, and then queuing the request again for success. Just to demonstrate that there's no state preventing that working?

ThetaSinner (commented):
Looking good, I like the approach and most of my comments are about naming and testing.

crates/api/src/op_store.rs
jost-s: resolved

crates/core/src/factories/core_fetch/back_off.rs
jost-s: I think it makes no difference because the module `back_off` is not `pub`.

crates/core/src/factories/core_fetch/test/response_queue.rs
jost-s: Yes, for sure, I introduced mockall in the other PR and wanted to adapt the existing tests in another PR. But I can do it here already too.

crates/core/src/factories/core_fetch/test/response_queue.rs
jost-s: I don't find the name very comprehensible either. But it's not a request, it's a response to a request which pulls ops from the store and sends them to the requesting agent.

crates/api/proto/op_store.proto
jost-s: That's true. I'll refactor.

jost-s: Wait, we don't have a hashing for that yet tho. Should I add some temporary hashing for the id?

ThetaSinner: Yes please. I had to do something similar with receiving ops into the store vs notifying the DHT model that ops have arrived on the host. Maybe it needs to be a proper test utility as part of demonstrating the host. I can't remember exactly where I did this because it's been a couple of weeks, but I think I skipped actually calculating hashes.

crates/core/src/factories/core_fetch/back_off.rs
ThetaSinner: I guess so yes, I just prefer to be as restrictive as possible :)

crates/core/src/factories/core_fetch/test/response_queue.rs
ThetaSinner: No problem if mockall is coming elsewhere. It can be added here later if it's a pain to mix that between PRs

crates/core/src/factories/core_fetch/test/response_queue.rs
ThetaSinner: I see what you're saying but in my head it kind of is a request. We're choosing to queue the work in the implementation and this doesn't actually respond in the function return but it is still a remote peer requesting us to do work... Naming is hard :) 

Maybe something like `handle_op_data_request`?

jost-s: Yes, that's good and fits the `handle_incoming_ops` in the next PR. I'll change it to `handle_op_request`. I'd change `add_ops` to `request_ops` too.

crates/core/src/factories/core_fetch/back_off.rs
jost-s: Okay, restrictive works for me.

crates/api/proto/op_store.proto
jost-s: I'll have to compute hashes if the ops that are requested by id and then sent without id need to be compared. I can add that as a test utility.

crates/core/src/factories/core_fetch/test/response_queue.rs
jost-s: I added it now after all.

crates/core/src/factories/core_fetch/test/response_queue.rs
jost-s: Sure, tested those two aspects https://github.com/holochain/kitsune2/pull/63/commits/a3ee21fd8b6547405a5cbdf5ae4ca34282e84b3e

(Look at the commit while I haven't rewritten git history yet ;-)

crates/core/src/factories/core_fetch/back_off.rs
jost-s: Restricted the hell out of it.

crates/api/proto/op_store.proto
jost-s: I added it for this test only so far. The hash function will sooner or later be added globally. We can refactor this then to use it.

crates/core/src/factories/core_fetch.rs
jost-s: Yes, looked twice now.

crates/core/src/factories/core_fetch.rs
jost-s: Applied

crates/api/src/fetch.rs
```diff
@@ -34,17 +42,29 @@ pub fn serialize_op_ids(value: Vec<OpId>) -> bytes::Bytes {
...
+    fn request_ops(
         &self,
-        op_list: Vec<OpId>,
-        source: AgentId,
+        ops: Vec<OpId>,
+        from: AgentId,
+    ) -> BoxFut<'_, K2Result<()>>;
+
+    /// Send held ops as a response to a requesting peer.
+    fn handle_op_request(
```
ThetaSinner, line 64: Nice, I like the naming now. It's clear these are a pair

crates/core/src/factories/core_fetch/test/request_queue.rs
```diff
@@ -0,0 +1,599 @@
...
+    .unwrap();
+
+    assert!(fetch
+        .state
+        .lock()
+        .unwrap()
+        .requests
+        .iter()
+        .all(|(_, agent_id)| *agent_id != agent_id_1),);
+}
```
ThetaSinner, line 599: These tests don't seem to be using the mockall'd transport, is that intentional?

crates/core/src/factories/core_fetch/test/response_queue.rs
```diff
@@ -0,0 +1,318 @@
...
+        ))
+        .unwrap(),
+    }
+}
+
+fn make_mock_transport(
+    responses_sent: Arc<Mutex<ResponsesSent>>,
+) -> Arc<MockTransport> {
+    let mut mock_transport = MockTransport::new();
+    mock_transport.expect_send_module().returning({
```
ThetaSinner, line 52: This is nice, just mocking what you need to mock for the test

ThetaSinner (approved):
Looking great, just needs a rebase :)

Cargo.toml
```diff
@@ -35,6 +35,8 @@ clap = "4.5.21"
 ctrlc = { version = "3.4.5", features = ["termination"] }
 # bootstrap_srv for signature verification.
 ed25519-dalek = "2.1.1"
+# automock traits for tests
+mockall = "0.13"
```
neonphog, line 39: move down to dev dependencies?

crates/api/Cargo.toml
```diff
@@ -14,6 +14,7 @@ edition = "2021"
 base64 = { workspace = true }
 bytes = { workspace = true }
 futures = { workspace = true }
+mockall = { workspace = true }
```
neonphog, line 17: move down to dev dependencies?

crates/api/src/fetch.rs
```diff
@@ -34,17 +42,29 @@ pub fn serialize_op_ids(value: Vec<OpId>) -> bytes::Bytes {
...
     let op_ids = OpIds::decode(value).map_err(K2Error::other)?;
-    let vec = Vec::from(op_ids);
-    Ok(vec)
+    Ok(op_ids.into())
+}
+
+/// Serialize list of ops for sending over the wire.
+pub fn serialize_ops(value: Vec<MetaOp>) -> bytes::Bytes {
+    let ops = Ops::from(value).encode_to_vec();
+    bytes::Bytes::copy_from_slice(&ops)
```
neonphog, line 51: This writes the data, and then copies it again. See: https://github.com/holochain/kitsune2/blob/9cd2ec666f66ebc90d71b445abd612002884e034/crates/api/src/protocol.rs#L15-L24

crates/api/src/transport.rs
```diff
@@ -139,6 +141,7 @@ pub trait TxImp: 'static + Send + Sync + std::fmt::Debug {
 pub type DynTxImp = Arc<dyn TxImp>;
 
 /// A high-level wrapper around a low-level [DynTxImp] transport implementation.
+#[automock]
```
neonphog, line 144:
```suggestion
#[cfg_attr(any(test, feature = "test_utils"), automock)]
```

?

crates/core/Cargo.toml
```diff
@@ -29,5 +29,7 @@ axum = { workspace = true, default-features = false, features = [
   "tokio",
 ] }
 ed25519-dalek = { workspace = true, features = ["rand_core"] }
+kitsune2_memory = { workspace = true }
 rand = { workspace = true }
+sha2 = "0.10.8"
```
neonphog, line 34: move to workspace

crates/api/Cargo.toml
neonphog: Alas, if this is needing to be exported it cannot go in dev-dependencies : ( -- but at the very least it should be optional and only enabled with a test_utils feature, right?

crates/api/src/fetch.rs
neonphog: This should not be in the external api.

1. In fetch module constructor--register with transport module to receive module messages.
2. When we receive a request, send that request to the response queue.
3. The request will come with a URL, so you don't need to look up by agent id... just send it to where we got the request from.

neonphog: Of course, the concrete implementation can have this function be private that you can call for testing to bypass actually receiving things in the transport.

neonphog (changes requested):
Some suggestions re efficiency and dependencies, but mostly the module api: https://github.com/holochain/kitsune2/pull/63#discussion_r1905770150

Otherwise, this looks awesome!

crates/api/src/fetch.rs
jost-s: Cool, I'll do that. Question about that protocol code, should that not panic when encoding fails?

neonphog: Indeed. Probably the only way it could error is memory allocation. Up to you : )

crates/api/src/fetch.rs
jost-s: `recv_module_msg` is sync, the response queue sender is async - use a tokio task for that?

neonphog: The pattern we've started to establish (but should evaluate if appropriate on a case-by-case basis) is to use a bounded channel, then [try_send](https://docs.rs/async-channel/2.3.1/async_channel/struct.Sender.html#method.try_send) and if the send fails, it means the system is overloaded, so log the problem and drop the request.

It seems like that is appropriate here.

jost-s: Great.

crates/api/src/transport.rs
jost-s: Yes, applied. `all(test, feature = "test_utils")` doesn't work. Do you know why? I would think that automock only works in a test when `test_utils` is enabled.

crates/core/src/factories/core_fetch/test/response_queue.rs
jost-s: I've taken out the test for agent info in peer store, because the response is now sent directly to the incoming url.

crates/api/proto/fetch.proto
jost-s: done https://github.com/holochain/kitsune2/pull/63/commits/48fdacbc3d74d86cfafefe7ce4575fda7592021f

crates/core/src/factories/core_fetch/test/request_queue.rs
jost-s: done here https://github.com/holochain/kitsune2/pull/63/commits/6a3221aa87f0f42aee660813ad5588c4c942a7db

Cargo.toml
jost-s: all the mockall feature gating https://github.com/holochain/kitsune2/pull/63/commits/d123d56c507e42f05e2088fa03d4da9c6b452e86

crates/api/src/fetch.rs
jost-s: done https://github.com/holochain/kitsune2/pull/63/commits/4659a4af7162553e50f00b809afa615f6e3010f0
and for op_id serialization https://github.com/holochain/kitsune2/pull/63/commits/3d7fb68e70738c8c1e8a5f53d9621a7bbbad5c76

crates/core/src/factories/core_fetch.rs
```diff
@@ -156,23 +207,26 @@ struct State {
 #[derive(Debug)]
 struct CoreFetch {
     state: Arc<Mutex<State>>,
-    fetch_queue_tx: Sender<FetchRequest>,
-    fetch_tasks: Vec<JoinHandle<()>>,
+    request_tx: Sender<FetchRequest>,
+    tasks: Vec<JoinHandle<()>>,
+    #[cfg(test)]
+    response_handler: DynTxModuleHandler,
```
jost-s, line 213: I'm not so happy with this, but somehow I need to send messages to the response queue. This seemed the simplest.

crates/core/src/factories/core_fetch.rs
```diff
@@ -146,6 +165,38 @@ impl FetchFactory for CoreFetchFactory {
...
+struct FetchResponseHandler {
+    response_tx: Sender<FetchResponse>,
+}
+
+impl TxModuleHandler for FetchResponseHandler {
+    fn recv_module_msg(
+        &self,
+        peer: Url,
+        _space: SpaceId,
+        _module: String,
```
jost-s, line 180: I don't know what to do with space and module here. For every space there'll be a fetch module instance, and this is a handler registered specifically for the fetch module, so what are these parameters for?

crates/api/src/transport.rs
ThetaSinner: Do we want to use `test-utils` rather than `test_utils` this time? Or no preference?

crates/core/src/factories/core_fetch.rs
neonphog: Just in case you want to use the same callback for multiple spaces / modules. You can safely ignore them.

crates/core/src/factories/core_fetch.rs
```diff
@@ -146,6 +165,38 @@ impl FetchFactory for CoreFetchFactory {
...
+    }
+}
+
+impl TxBaseHandler for FetchResponseHandler {
+    fn peer_connect(&self, _peer: Url) -> K2Result<()> {
+        Ok(())
+    }
+    fn new_listening_address(&self, _this_url: Url) {}
+    fn peer_disconnect(&self, _peer: Url, _reason: Option<String>) {}
+}
```
neonphog, line 199: pretty sure these all have default impls

```suggestion
impl TxBaseHandler for FetchResponseHandler {}
```

crates/core/src/factories/core_fetch.rs
neonphog: meh, I'm fine with this : )

crates/api/src/transport.rs
neonphog: I wish rust just never allowed dashes anywhere in filenames or features or anything... then everything could be underscores and we wouldn't have to worry about stuff like this : ) -- that being said, given the way things are already, I don't have a preference. Totally cool to switch this or not : )

neonphog (dismissed):
Sweet

crates/core/src/factories/core_fetch.rs
jost-s: That hadn't occurred to me.

crates/api/src/transport.rs
jost-s: I have no preference whatsoever, just for consistency. @ThetaSinner Are you suggesting using dashes in features?

jost-s: And would be great to know the answer to this https://github.com/holochain/kitsune2/pull/63#discussion_r1906250155.

ThetaSinner (dismissed):

crates/api/src/transport.rs
ThetaSinner: > Yes, applied. all(test, feature = "test_utils") doesn't work. Do you know why? I would think that automock only works in a test when test_utils is enabled.

I'm not sure about this. it's not documented so I'd expect this to behave like any other attribute and be possible to conditionally enable. What happens?

ThetaSinner: Oh but shouldn't it be an "any"? So that it's enabled either on "test" or from another crate if "test_utils" is enabled?

jost-s: When I use `all` instead of `any` and then run `make test`, the mock imports cannot be found in the tests.

jost-s: Using the automock'd struct inside of `kitsune2_api` fails if run without feature `test_utils`.

jost-s: Here it gets wicked: With `all`, the test in `kitsune2_api` succeeds when run with `--features test-utils`. But the test in `kitsune2_core` only works with `any`.

jost-s: Mystery lifted: `#[cfg(test)]` is only active for unit tests of the crate where the tests are run. So `all(test, feature = "mockall")` will work for `kitsune2_api` if there were unit tests _and_ the feature is enabled when running the test. It will never work for tests outside of the crate, because `cfg(test)` is then inactive. That's why `any` must be used.

jost-s:
All tests enabled again an no longer flaky, it seems. Please approve again @neonphog.

neonphog (approved):

PR #66, opened 2025-01-06
DHT Docs--Alternate Svg Options

neonphog:
Took my own stab at these.

I wanted them to have the same number of time slices back to the origin, a little more separation between the segments, and labels on both diagrams to more easily relate them.


ThetaSinner (approved):
These are so much better, I am pretty disappointed in my own artwork now :D 

I really like the cutaway on the circle one! Thank you for these

PR #67, opened 2025-01-07
Move the in-memory op store to the core crate

ThetaSinner:


neonphog (approved):
:+1:

ThetaSinner:
This one did conflict with the docs PR, but only on imports

neonphog (approved):

PR #68, opened 2025-01-10
Gossip init

ThetaSinner:
Trying to keep this contained. The gossip implementation is 99% missing and I've just added what I think the first two messages should look like. Otherwise this is all just plumbing to be filled in


crates/api/src/gossip.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use crate::peer_store::DynPeerStore;
+use crate::transport::{DynTransport, DynTxModuleHandler};
+use crate::{builder, config, BoxFut, DynOpStore, K2Result, SpaceId};
+use std::sync::Arc;
+
+/// Represents the ability to sync DHT data with other agents through background communication.
+pub trait Gossip: 'static + Send + Sync + std::fmt::Debug {
+    /// Get the tx module handler for this gossip instance.
+    fn tx_module_handler(&self) -> DynTxModuleHandler;
```
ThetaSinner, line 11: This I'm not quite sure about. I want to be able to have the gossip implementation also act as a module message receiver. That is think is reasonable but I'm wondering if there was a different way of achieving that already envisioned

crates/core/Cargo.toml
```diff
@@ -11,11 +11,12 @@ categories = ["network-programming"]
 edition = "2021"
 
 [dependencies]
+kitsune2_api = { workspace = true }
```
ThetaSinner, line 14: These are easier to find if they're kept in a separate block. We've done that in most places I think

crates/api/src/lib.rs
```diff
@@ -59,7 +59,10 @@ pub use timestamp::*;
 
 pub mod fetch;
 
-pub mod op_store;
```
ThetaSinner, line 62: Prefer not having both "pub mod" and "pub use". I only need to be able to import it one way. If we have conflicts between modules then we'd have to resolve that with "pub use" anyway

crates/gossip/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "kitsune2_gossip"
```
neonphog, line 2: Does this need a separate crate? It feels like this particular gossip implementation is exactly tied to the code in the dht crate, so perhaps they could be one and the same?

crates/gossip/proto/gossip.proto
```diff
@@ -0,0 +1,49 @@
+syntax = "proto3";
+
+package kitsune2.gossip;
+
+// A Kitsune2 gossip protocol message.
+message K2GossipProto {
+  oneof gossip_message {
```
neonphog, line 7: oneof has some fiddly [backward compatibility issues](https://protobuf.dev/programming-guides/proto3/#backward).

Another way of handling this would be to have an `enum` that types the message, then a further encoded bytes field to be parsed based on the enum. This would mean more complexity in our code, though, so I'm okay with the fiddly-ness so long as we're aware of it.

crates/api/src/gossip.rs
neonphog: I was thinking either the factory::create or the constructor for the internal gossip module instance type could do the registering of the module callback (since they are passed the transport instance) so that wouldn't have to be exposed outside of the module.

crates/core/src/factories.rs
```diff
@@ -18,8 +18,12 @@ pub use core_bootstrap::CoreBootstrapFactory;
...
 pub use core_fetch::CoreFetchFactory;
 
+mod core_gossip;
+pub use core_gossip::*;
+
 mod mem_transport;
 pub use mem_transport::*;
 
 mod mem_op_store;
+
```
neonphog, line 28:
```suggestion
```

crates/gossip/src/gossip.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+struct GossipResponse(bytes::Bytes, Url);
+
+/// The gossip implementation.
+///
+/// This type acts as both an implementation of the [Gossip] trait and a [TxModuleHandler].
+#[derive(Debug, Clone)]
+pub struct K2Gossip {
```
neonphog, line 56:
```suggestion
struct K2Gossip {
```

crates/gossip/src/protocol.rs
```diff
@@ -0,0 +1,22 @@
...
+    K2GossipProto::decode(value).map_err(K2Error::other)
+}
+
+/// Serialize a gossip message
+pub fn serialize_gossip_message(
+    value: K2GossipProto,
+) -> K2Result<bytes::Bytes> {
+    let mut buf = Vec::new();
+    value.encode(&mut buf).map_err(K2Error::other)?;
+    Ok(bytes::Bytes::from(buf))
```
neonphog, line 21: I think at the moment this compiles down to a zero copy since internally bytes is using a Vec... but in the docs it warns not to depend on this behavior because they might switch the internal repr someday. To future proof: https://github.com/holochain/kitsune2/blob/093aceb02791c6ffd206190146bd7803abc2fec6/crates/api/src/protocol.rs#L15-L24

crates/gossip/proto/gossip.proto
```diff
@@ -0,0 +1,49 @@
...
+
+    // A gossip acceptance protocol message.
+    K2GossipAcceptProto accept = 2;
+  }
+}
+
+// A Kitsune2 gossip initiation protocol message.
+message K2GossipInitiateProto {
+  // The agents from the initiator who are participating in this gossip round.
+  repeated bytes participating_agents = 1;
```
neonphog, line 19: Are these full agent infos? Perhaps add some docs about the contents here.

crates/gossip/proto/gossip.proto
```diff
@@ -0,0 +1,49 @@
...
+
+  // Request ops that are new since the given timestamp.
+  int64 new_since = 3;
+
+  // Ops that we have stored since the timestamp provided by the initiator in `new_since`.
+  repeated bytes new_ops = 4;
+
+  // Provide a new bookmark for the initiator. Any new ops will have been returned in `new_ops`
+  // and the initiator should use this new timestamp in their `new_since` next time they gossip with us.
+  int64 updated_new_since = 5;
```
neonphog, line 44: As written here, it feels like we might have to just keep sending accepts back and forth until we're caught up on time, since the accept contains the requested ops and requests ops in turn. Should we rename this to "SyncProto" or something?

```mermaid
sequenceDiagram
    Alice->>Bob: Init(new_since = 4)
    Bob->>Alice: Accept(new_since = 17)
    Alice->>Bob: Accept(new_since = 42)
    Bob->>Alice: Accept(new_since = 99)
```

neonphog (changes requested):
Review complete. Other than removing the exposure of the transport handler that you pointed out, all my other comments could be addressed later as you're doing the real implementation.

It'll be interesting if the gossip module trait ends up not having any functions in it, because it's just manipulating other modules passed in : )

crates/api/src/gossip.rs
ThetaSinner: That makes a lot more sense, done!

crates/gossip/Cargo.toml
ThetaSinner: I'm thinking the gossip needs to do more than the DHT module should be responsible for. I'd be fine with merging the gossip implementation into the top level crate because it's probably not useful anywhere other than that. For now I'd like to leave it in its own crate and work out where to put it later

crates/gossip/proto/gossip.proto
ThetaSinner: Hmm yeah okay. So I guess if we're careful with an enum then we'd be a bit more able to implement backwards compatibility ourselves but aren't we going to have similar problems?

I'd suggest that if we really want to make major changes then we can test them against an older Kitsune2. If we have to make breaking changes then we define a new `oneof` field and have the code behave differently depending on which one is present until we can deprecate the old field.

crates/gossip/src/protocol.rs
ThetaSinner: Good call, done

crates/gossip/proto/gossip.proto
ThetaSinner: They are not, they are just the ids. I believe this is similar to what we currently do in gossip, send the ids, check against the store, request what was missing.

Updated the doc comments to make this clearer

crates/gossip/proto/gossip.proto
neonphog: Yep, I'm fine with that approach.

crates/gossip/proto/gossip.proto
ThetaSinner: I wasn't planning to do that. I'd like for gossip to terminate in a predictable number of steps. There is a missing field here from what was discussed around incremental update, which is a size limit. So you should be able to request "from this time stamp, up to X MB". You get that back and you do your DHT diff check to see what your historical sync state is. But you don't keep looping. The next message types should cover "no DHT diff, here are your new ops", "disc diff" or "ring diff".

When the size limit is hit, you can gossip with somebody else to discover more ops but otherwise, you'll have to continue from the updated bookmark timestamp on your next gossip round. If there's a lot of data, it will take some time to sync.

Gossip pulling all the hashes at once for a large data set won't actually help because it'll just overflow the fetch queue and we'll end up having to fetch ops through the DHT diff mechanism anyway. Which is more work.

As for how to tune this right, I'm not sure what that'll look like. Maybe if we are fetching quickly and hitting size limits then we should be initiating gossip with those peers again sooner to keep the pace of the sync up. Equally, Kitsune 1 was tuned for large networks, so we could assume that there are enough peers to sync from that we'll be able to keep our assigned bandwidth maxed out just by fetching a little bit from everyone.

ThetaSinner: Or the much shorter answer, I don't know yet, but if this needs renaming, then it can be renamed :)

neonphog: Hehe, yep totally fair. :+1:

neonphog (approved):

jost-s (approved):

crates/core/src/factories/core_gossip.rs
```diff
@@ -0,0 +1,55 @@
...
+use std::sync::Arc;
+
+#[cfg(test)]
+mod test;
+
+/// Factory for creating core gossip instances.
+///
+/// This factory returns stub gossip instances that do nothing.
+#[derive(Debug)]
+pub struct CoreGossipFactory;
```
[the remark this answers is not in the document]
ThetaSinner: Certainly, done

crates/gossip/proto/gossip.proto
```diff
@@ -0,0 +1,49 @@
...
+
+  // Request ops that are new since the given timestamp.
+  int64 new_since = 2;
+
+  // TODO Send initial DHT snapshot.
+}
+
+// A Kitsune2 gossip acceptance protocol message.
+message K2GossipAcceptProto {
+  // The agent ids of the from the acceptor who are participating in this gossip round.
```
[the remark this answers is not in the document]
ThetaSinner: Updated to something similar but it is the agents currently in the space rather than the agent's in the peer store. The way things are set up, that's the same but I do want to be explicit about it here. The comment should match the one in the initiate message so I updated both

jost-s (approved):

neonphog (approved):

PR #70, opened 2025-01-13
Add handler for request/response to fetch module

jost-s:
Fetch sends requests for ops. There'll be incoming requests and incoming responses to these requests to be handled. This PR separates existing handler code into a separate file and adds the response encoding/decoding and handling, plus tests just for the handler.

Delegation of incoming responses to the corresponding task will be implemented in another PR.


crates/api/proto/fetch.proto
```diff
@@ -4,14 +4,33 @@ package kitsune2.fetch;
...
-  repeated kitsune2.op_store.Op op_list = 1;
+  repeated kitsune2.op_store.Op ops = 1;
+}
+
+// Fetch message.
+message K2FetchProto {
+  // Fetch message type.
+  enum FetchMessageType {
+    // Default value.
+    UNSPECIFIED = 0;
```
jost-s, line 24: This is recommended in [the language guide](https://protobuf.dev/programming-guides/proto3/#enum-default)

crates/api/src/fetch.rs
```diff
@@ -92,40 +97,47 @@ pub type DynFetchFactory = Arc<dyn FetchFactory>;
...
+        };
+        println!("op id {}", op_1.op_id);
+        let ops_enc = Response::from(vec![op_1]).encode_to_vec();
+        let ops_dec = Request::decode(ops_enc.as_slice()).unwrap();
+        let op_ids_dec_vec = ops_dec
+            .op_ids
+            .into_iter()
+            .map(Into::into)
+            .collect::<Vec<OpId>>();
+        println!("op ids dec {op_ids_dec_vec:?}");
```
jost-s, line 136: This is an example of where decoding to the wrong type doesn't fail and yields an empty vector. I'll delete this before merging the PR.

crates/core/src/factories/core_fetch/test.rs
```diff
@@ -31,4 +32,31 @@ pub(crate) mod utils {
...
+                data,
+            ))
+            .unwrap(),
+        }
+    }
+
+    pub fn meta_op_from_op(op: Op) -> MetaOp {
+        let mem_op = Kitsune2MemoryOp::try_from(op).unwrap();
+        MetaOp::try_from(mem_op).unwrap()
+    }
```
jost-s, line 61: These have just been moved here for availability.

crates/api/src/fetch.rs
```diff
@@ -1,59 +1,72 @@
...
         }
     }
 }
 
 /// Serialize list of op ids for sending over the wire.
-pub fn serialize_op_ids(value: Vec<OpId>) -> bytes::Bytes {
-    let mut out = bytes::BytesMut::new();
-    OpIds::from(value)
+pub fn serialize_request(value: Vec<OpId>) -> bytes::Bytes {
+    let mut data = BytesMut::new();
```
ThetaSinner, line 39: I'd be tempted to do the inner serialize in a separate function. It takes a bit of thought to decode the double encode that's happening in here

crates/api/src/fetch.rs
ThetaSinner: I think this is expected behavior though right? It's why things like this exist, to help coordinate schemas across applications https://buf.build/product/bsr

At least that's why I've used the Confluent schema registry with Avro when trying to use binary encoding over Kafka.

crates/api/src/op_store.rs
```diff
@@ -61,6 +63,7 @@ impl PartialOrd for StoredOp {
 }
 
 /// The API that a kitsune2 host must implement to provide data persistence for kitsune2.
+#[cfg_attr(any(test, feature = "mockall"), automock)]
 pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
```
ThetaSinner, line 67: Question! Now that I've spent a bit more time getting to grips with using the core module. I'm wondering whether we want to `automock` things that have in-memory implementations? Like the op store and transport modules do.

I think the answer should be yes if we're testing errors or something else that the memory implementations won't do for us. So I think this is valid to put here but I am wondering if the tests in this PR need it still?

crates/core/src/factories/core_fetch/message_handler.rs
```diff
@@ -0,0 +1,212 @@
...
+    K2Error, K2Result, SpaceId, Url,
+};
+use prost::Message;
+use tokio::sync::mpsc::Sender;
+
+use super::{IncomingFetchRequest, IncomingResponse};
+
+#[derive(Debug)]
+pub(super) struct FetchMessageHandler {
+    pub(super) response_tx: Sender<IncomingFetchRequest>,
```
ThetaSinner, line 15: This naming confuses me a little why are requests going into a response channel?

crates/core/src/factories/core_fetch/test/request_queue.rs
```diff
@@ -32,10 +35,20 @@ fn make_mock_transport(
             assert_eq!(space, SPACE_ID);
             assert_eq!(module, crate::factories::core_fetch::MOD_NAME);
             Box::pin({
-                let value = requests_sent.clone();
+                let requests_sent = requests_sent.clone();
```
ThetaSinner, line 38: This is where I'm wondering if we've not got enough in core to be able to test with the in-memory transport and just observe the results, rather than the inner messages?

ThetaSinner (commented):
This is looking good. A couple of questions about testing and naming but nothing that would stop me approving after a quick discussion :)

crates/api/src/fetch.rs
jost-s: What is a schema used for here?

It is probably nothing we prevent though. If the op is just a byte array, and a response is a vector of such ops, and an op id is a byte array, and the request is a vector of such op ids, then deserializing ops into op ids results in the op ids being the entire ops. As there's no length to op ids, they may well be byte arrays of the length of the whole op.

crates/core/src/factories/core_fetch/test/request_queue.rs
jost-s: Yes, likely so. An in-memory transport could just keep a hash map of registered handlers and do the forwarding. What do you mean by observe the results? How should the transport process inner messages to achieve the results?

crates/api/src/fetch.rs
jost-s: Cool, done.

crates/core/src/factories/core_fetch/test/request_queue.rs
ThetaSinner: I thought at some point there were tests here checking that the right outgoing messages were sent in response but maybe that never worked through the transport :) That's not a problem if the tests aren't depending on getting hold of the messages and can just stub the receiver of the message to check that it arrived

There already is an in-memory transport, the coretransport type does that

crates/api/src/op_store.rs
jost-s: This PR does not need the op store mock. It was split off from the process incoming response PR, where the mock is used to fail at writing ops in a test.

crates/core/src/factories/core_fetch/test/request_queue.rs
jost-s: The following file checks outgoing responses. Is the problem there?

crates/core/src/factories/core_fetch/message_handler.rs
jost-s: It's also been quite confusing to me. I did a renaming sweep and am quite happy with it now. Let me know if you can understand it.

crates/api/src/fetch.rs
ThetaSinner: In the cases I know about, you use a serializer that knows about the schema types and the schema registry. It's all slightly magical but the outcome is just a "type" and a "version" that tell you exactly what schema, schema version and message kind are in use. Literally just some extra bytes on the start of the message to avoid exactly what you're running into.

crates/core/src/factories/core_fetch/test/request_queue.rs
ThetaSinner: Probably not, I was struggling to find what I was looking for because the diff is quite noisy. It was more of a general question. This looks fine as is :)

ThetaSinner (approved):
Nice, much clearer with the renaming :)

crates/core/src/factories/core_fetch/message_handler.rs
ThetaSinner: Yeah, much easier to follow. Thank you!

neonphog (approved):
woot

PR #71, opened 2025-01-14
refactor(op-store)!: pass ops as bytes to op store

jost-s:
Refactor the op store trait to take opaque bytes instead of `MetaOp`s when processing incoming ops. I think that's already it. Retrieving ops from the store should still be `MetaOp`s.


crates/api/src/op_store.rs
```diff
@@ -68,7 +68,7 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
     /// if it is able to process them.
     fn process_incoming_ops(
         &self,
-        op_list: Vec<MetaOp>,
+        op_list: Vec<Op>,
```
ThetaSinner, line 71: I probably wouldn't choose to use the wire types in the API. Could we use `bytes::Bytes` here?

ThetaSinner (approved):
Looks good otherwise! 

Are all the `From`/`TryFrom` conversions in `mem_op_store.rs` still used? There's a conversion between mem ops and meta ops that also used serialization that might not be useful any more?

jost-s:
> Are all the `From`/`TryFrom` conversions in `mem_op_store.rs` still used? There's a conversion between mem ops and meta ops that also used serialization that might not be useful any more?

They weren't, deleted the Kitsune2MemoryOp conversions.

crates/api/src/op_store.rs
jost-s: I follow your inclination here =)
No, makes perfect sense to not use the wire representation.

ThetaSinner (approved):
Nice, looking good!

jost-s:
Apologies, I addressed another windows test flake. Windows getting slower by the day obviously.

ThetaSinner (approved):

PR #72, opened 2025-01-14
Go go gossip

ThetaSinner:
Adds:
  - Gossip round tracking
  - Sync of *unknown* agents. I'm expecting agents to broadcast their updated agent infos to known peers, so we should only need to share agents that just aren't known at all during gossip I think.
  - Incremental sync of "what's new" ops. We can't actually sync op data yet, just discover op ids, until we have the fetch queue.
  - Module config for K2Gossip
  - PeerMetaStore implementation for K2Gossip

Changes:
  - Update the wire protocol for gossip
  - Various tweaks to the API and other modules, which I've left comments for


crates/api/src/agent.rs
```diff
@@ -326,13 +326,14 @@ impl AgentInfoSigned {
...
-            })
+            .map(|v| Self::from_parts(verifier, v.agent_info, v.signature))
             .collect())
     }
 
-    fn inner_decode_one<V: Verifier>(
+    /// Create an instance from its parts.
+    ///
+    /// Useful when receiving an agent info over the wire.
+    pub fn from_parts<V: Verifier>(
```
ThetaSinner, line 336: Exposed this so that I can reconstruct these after sending them over the network. Happy to be told there's a better way

crates/api/src/gossip.rs
```diff
@@ -18,13 +22,17 @@ pub trait GossipFactory: 'static + Send + Sync + std::fmt::Debug {
     fn default_config(&self, config: &mut config::Config) -> K2Result<()>;
 
     /// Construct a gossip instance.
+    #[allow(clippy::too_many_arguments)]
```
ThetaSinner, line 25: Needs all the things 👍🏻

crates/api/src/op_store.rs
```diff
@@ -91,6 +91,22 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
...
+    ///
+    /// The `limit_bytes` applies to the size of the op data, not the size of the op ids.
+    ///
+    /// # Returns
+    ///
+    /// As many op ids as can be returned within the `limit_bytes` limit.
+    ///
+    /// If the limit is applied, then the timestamp of the last op id returned is returned.
+    /// Otherwise, the timestamp for when this operation started is returned.
+    fn retrieve_op_ids_bounded(
```
ThetaSinner, line 104: This is new and is intended to be used for either incremental get or with the DHT diffing when that's integrated. This will need modifying to respect the common arc set at some point which means moving that from the DHT model to the API. Think that's better left for another PR while we're doing full arc only for now

crates/api/src/peer_meta_store.rs
```diff
@@ -1,34 +1,34 @@
-use crate::{builder, config, AgentId, BoxFut, K2Result, SpaceId, Timestamp};
+use crate::{builder, config, BoxFut, K2Result, SpaceId, Timestamp, Url};
 use futures::future::BoxFuture;
 use std::sync::Arc;
 
 /// A store for peer metadata.
 ///
-/// This is expected to be backed by a key-value store that keys by space, agent_id and key.
+/// This is expected to be backed by a key-value store that keys by space, peer URL and key.
```
ThetaSinner, line 7: This was my mistake implementing this on the first pass. We want to track interactions with a peer/node rather than each agent on that node. Here, it's not very interesting what the set of agents on that machine are doing, really just their common arc set.

What is a pain is that this is a session value. So every time the peer restarts or reconnects we are starting fresh with information about them... I think that's fine for gossip, and probably for fetch too. I'm not sure whether it's a good enough replacement for the information we were capturing in the peer metrics in K1. I'm thinking we'll know when we need something like that!

crates/core/src/factories/mem_op_store.rs
```diff
@@ -234,6 +234,55 @@ impl OpStore for Kitsune2MemoryOpStore {
         })
     }
 
+    fn retrieve_op_ids_bounded(
```
ThetaSinner, line 237: This isn't particularly efficient, I'm hoping it's clear though and for tests I'm not expecting us to be dealing with more than a few hundred ops, maybe a few thousand.

crates/dht/src/arc_set.rs
```diff
@@ -114,6 +114,22 @@ impl ArcSet {
         self.inner.len()
     }
 
+    /// Create an [ArcSet] from a list of sectors indices.
+    ///
+    /// This is intended to be used for constructing an [ArcSet] from a network message.
+    pub fn from_raw(inner: impl Iterator<Item = u32>) -> Self {
```
ThetaSinner, line 120: Same deal here with sending over the network, I want to be able to reconstruct these and am happy to be pointed at a better method.

One option would just be to serialize the type and that avoids exposing the implementation, for a little extra brain power required to understand all the levels of nested serialization

crates/gossip/proto/gossip.proto
```diff
@@ -4,46 +4,133 @@ package kitsune2.gossip;
 
 // A Kitsune2 gossip protocol message.
 message K2GossipMessage {
-  oneof gossip_message {
+  enum GossipMessageType {
```
ThetaSinner, line 7: Switched to preferred pattern for enums - it's actually minimal impact on the code and probably easier to evolve so all good there :)

crates/gossip/proto/gossip.proto
```diff
@@ -4,46 +4,133 @@ package kitsune2.gossip;
...
   // The agent ids of the agents from the initiator who are in the in the peer store for the space where gossip is running.
-  repeated bytes participating_agents = 1;
+  repeated bytes participating_agents = 10;
+
+  // The DHT sectors covered by the union of the agents in the participating_agents list.
+  ArcSetMessage arc_set = 11;
 
   // Request ops that are new since the given timestamp.
-  int64 new_since = 2;
+  int64 new_since = 20;
```
ThetaSinner, line 59: Updating field numbers to leave some room for grouping fields later

crates/gossip/proto/gossip.proto
```diff
@@ -4,46 +4,133 @@ package kitsune2.gossip;
...
+// This message is a final message when used in a gossip round.
+message K2GossipAgentsMessage {
+  bytes session_id = 1;
 
-  // TODO Send DHT snapshot diff response.
+  // The agent infos for the agents that were sent back in the missing_agents list of the previous message.
+  repeated AgentInfoMessage provided_agents = 10;
 }
 
 // TODO Further gossip protocol messages.
```
ThetaSinner, line 136: Note, still no DHT diff messages here

crates/gossip/src/gossip.rs
ThetaSinner, line 1: This module is still REALLY rough and I'm nowhere near happy with the testing. I can't get much further without the fetch queue though and this review is big enough.

Please comment as required but I do plan to make improvements to this code and to split this file up... and add many more tests :)

crates/gossip/src/gossip.rs
```diff
@@ -279,32 +854,98 @@ mod test {
...
+
+        harness_2
+            .gossip
+            .initiate_gossip(agent_1.agent().clone())
+            .await
+            .unwrap();
+
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
+        // TODO fetch ops and assert that the op stores get updated.
```
ThetaSinner, line 949: This is a special test where you run it and look for "Discovered op ids:" in the output :)

This test is working for me and op ids are being exchanged but without the fetch queue there's no way to retrieve the op data yet.

crates/gossip/src/state.rs
```diff
@@ -0,0 +1,194 @@
+use crate::protocol::{
+    K2GossipAcceptMessage, K2GossipAgentsMessage, K2GossipNoDiffMessage,
+};
+use kitsune2_api::{AgentId, K2Error, K2Result, Url};
+use rand::RngCore;
+
+/// The state of a gossip round.
+#[derive(Debug)]
+pub(crate) struct GossipRoundState {
```
ThetaSinner, line 9: This is a similar thing to what we had in K1. It's just about keeping track of the gossip round so that we can apply a timeout and check that the other peer is sending us messages in the sequence that we're expecting, based on our state as we go.

crates/gossip/src/state.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// The time at which this round was initiated.
+    ///
+    /// This is used to apply a timeout to the round.
+    #[allow(dead_code)]
+    started_at: std::time::Instant,
+
+    /// The session id of this round.
+    ///
+    /// Must be randomly chosen and unique for each initiated round.
+    pub session_id: bytes::Bytes,
```
ThetaSinner, line 22: Here to prevent message replay etc. Helps us avoid getting into unexpected states by receiving messages that were created before a timeout or anything like that

crates/api/src/agent.rs
neonphog: I was imagining you'd use the `encode` below to send a single string for agent info across the wire, then use the above decode to decode it.

crates/api/src/op_store.rs
```diff
@@ -91,6 +91,22 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
...
 
+    /// Retrieve a size-bounded list of op ids that have been stored since `start`.
+    ///
+    /// The `limit_bytes` applies to the size of the op data, not the size of the op ids.
+    ///
+    /// # Returns
+    ///
+    /// As many op ids as can be returned within the `limit_bytes` limit.
+    ///
+    /// If the limit is applied, then the timestamp of the last op id returned is returned.
```
jost-s, line 102:
```suggestion
    /// If the limit is applied, then the timestamp of the last op id is returned.
```

crates/api/src/peer_meta_store.rs
neonphog: Indeed, the legacy peer metrics never really worked as intended. We'll need to design that fresh when we get to sharding I think.

crates/api/src/op_store.rs
```diff
@@ -91,6 +91,22 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
...
+    /// Otherwise, the timestamp for when this operation started is returned.
```
jost-s, line 103: What is this operation start time used for?

crates/api/src/gossip.rs
```diff
@@ -18,13 +22,17 @@ pub trait GossipFactory: 'static + Send + Sync + std::fmt::Debug {
...
     fn create(
         &self,
         builder: Arc<builder::Builder>,
-        space: SpaceId,
+        space_id: SpaceId,
+        space: DynSpace,
```
neonphog, line 30: I'm not sure this can work. The space is going to be constructing the gossip instance, and while we could potentially use [Arc::new_cyclic](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.new_cyclic) to keep a weak reference to ourselves to pass into the gossip module, I'd worry about that type of circularity causing mutex deadlocks.

crates/api/src/space.rs
```diff
@@ -57,6 +58,9 @@ pub trait Space: 'static + Send + Sync + std::fmt::Debug {
     /// be made to publish this tombstone to peers in the space as well.
     fn local_agent_leave(&self, local_agent: id::AgentId) -> BoxFut<'_, ()>;
 
+    /// Get a list of all local agents currently joined to this space.
+    fn get_local_agents(&self) -> BoxFut<'_, K2Result<Vec<DynLocalAgent>>>;
```
neonphog, line 62: Re my previous comment, it makes sense gossip would need this... but we may need to make a separate data structure that the space can own but that can also be passed into gossip without causing circular references.

crates/core/src/factories/core_space.rs
```diff
@@ -361,6 +362,15 @@ impl Space for CoreSpace {
         })
     }
 
+    fn get_local_agents(&self) -> BoxFut<'_, K2Result<Vec<DynLocalAgent>>> {
+        let inner = self.inner.clone();
+
+        Box::pin(async move {
+            let inner = inner.lock().unwrap();
```
jost-s, line 369:
```suggestion
        Box::pin(async move {
            let inner = self.inner.lock().unwrap();
```

crates/dht/src/arc_set.rs
```diff
@@ -114,6 +114,22 @@ impl ArcSet {
         self.inner.len()
     }
 
+    /// Create an [ArcSet] from a list of sectors indices.
```
jost-s, line 117:
```suggestion
    /// Create an [ArcSet] from a list of sector indices.
```

neonphog (commented):
Couple comments and a once-over to try to get the structure of changes. I'll give this a more thorough review tomorrow.

crates/gossip/proto/gossip.proto
```diff
@@ -4,46 +4,133 @@ package kitsune2.gossip;
...
 // A Kitsune2 gossip initiation protocol message.
+//
+// Acceptable responses:
+// - `K2GossipAcceptMessage`
 message K2GossipInitiateMessage {
+  bytes session_id = 1;
+
   // The agent ids of the agents from the initiator who are in the in the peer store for the space where gossip is running.
-  repeated bytes participating_agents = 1;
+  repeated bytes participating_agents = 10;
```
jost-s, line 53: Why this jump to 10 and 20 indices?

crates/gossip/src/config.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct K2GossipConfig {
+    /// The maximum number of bytes of op data to request in a single gossip round.
+    ///
+    /// This applies to both "new ops" which is the incremental sync of newly created data, and
+    /// to a DHT mismatch in existing data. The new ops are synced first, so the remaining capacity
+    /// is used for the mismatch sync.
+    ///
+    /// The maximum size of an op is host dependant, but assuming a 1MB limit, this would allow
```
jost-s, line 15: Easier to read with a space between value and unit.
```suggestion
    /// The maximum size of an op is host dependant, but assuming a 1 MB limit, this would allow
```

crates/gossip/src/gossip.rs
```diff
@@ -26,132 +35,634 @@ impl K2GossipFactory {
...
+    // TODO dead code until the initiate task is created
+    #[allow(dead_code)]
+    pub(crate) async fn initiate_gossip(
+        &self,
+        target: AgentId,
+    ) -> K2Result<bool> {
+        let state = self.initiated_round_state.clone();
+        let mut initiated_lock = state.lock().await;
+        if initiated_lock.is_some() {
+            tracing::info!("initiate_gossip: already initiated");
```
jost-s, line 190: I imagine this is not supposed to happen if everything works correctly. Should these "already doing this" messages be at `warn` level?

crates/gossip/src/peer_meta_store.rs
```diff
@@ -0,0 +1,174 @@
...
+        let inner = MemPeerMetaStoreFactory::create()
+            .create(builder)
+            .await
+            .unwrap();
+        K2PeerMetaStore::new(inner, space)
+    }
+
+    #[tokio::test]
+    async fn create() {
+        test_store().await;
```
jost-s, line 140: This seems to be sufficiently tested implicitly by the other tests.

crates/gossip/src/protocol.rs
```diff
@@ -1,26 +1,143 @@
...
+                ))
+            })?;
+
+            Ok((k2_gossip_message::GossipMessageType::Agents, out.freeze()))
+        }
+    }
+}
+
+/// Encode agent ids as bytes
+pub(crate) fn encode_agent_ids(
```
jost-s, line 119: We could add this or `From<IntoIterator>` for ids to the `api/id` module. They're used in multiple places.

crates/gossip/src/protocol.rs
```diff
@@ -1,26 +1,143 @@
...
 ) -> K2Result<bytes::Bytes> {
     let mut out = bytes::BytesMut::new();
 
-    value.encode(&mut out).map_err(|e| {
+    let (msg_type, inner) = serialize_inner_gossip_message(value)?;
+
+    K2GossipMessage {
+        msg_type: msg_type as i32,
+        data: inner,
+    }
```
jost-s, line 60:
```suggestion
    let (msg_type, data) = serialize_inner_gossip_message(value)?;

    K2GossipMessage {
        msg_type: msg_type as i32,
        data,
    }
```

crates/gossip/src/state.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub stage: RoundStage,
+}
+
+impl GossipRoundState {
+    /// Create a new gossip round state.
+    pub(crate) fn new(
+        session_with_peer: Url,
+        our_agents: Vec<AgentId>,
+    ) -> Self {
+        let mut session_id = bytes::BytesMut::with_capacity(96);
```
jost-s, line 36: Why 96?

jost-s (dismissed):
Mostly comments and questions. LGTM

crates/api/src/op_store.rs
ThetaSinner: This is to do with the incremental get of ops by stored_at. The start time is the bookmark that the other peer provided to us to fetch ops from that point in time. I can improve naming here maybe to make that clearer?

jost-s: I thought so, but it doesn't immediately make sense, because one peer could be requesting ops from a week ago with a limit, then get half of them with the last op's timestamp, and get the other half with the timestamp of when the operation started which is a couple of seconds ago maybe.
I see that it is probably an operation over _all_ ops that exist since the start_time until now, so the above case would only happen if there are no more ops after the second batch of them. Then it makes sense and could be aided by a short comment.

crates/api/src/gossip.rs
```diff
@@ -18,13 +22,17 @@ pub trait GossipFactory: 'static + Send + Sync + std::fmt::Debug {
...
         peer_store: DynPeerStore,
+        peer_meta_store: DynPeerMetaStore,
         op_store: DynOpStore,
         transport: DynTransport,
+        agent_verifier: DynVerifier,
```
neonphog, line 35: This is available in the builder.

crates/dht/src/arc_set.rs
neonphog: If we're pretty sure we're going to stick with the 512 indefinitely, what do you think about packing these into the individual bits of 64 bytes rather than the potentially 4096 bytes of 512 u32s?

crates/gossip/src/config.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// for at least 100 ops to be requested in a single round.
+    ///
+    /// Default: 100MB
+    pub max_gossip_op_bytes: u32,
+
+    /// The interval in seconds between gossip rounds.
+    ///
+    /// Default: 300 (5m)
+    pub interval_s: u64,
```
neonphog, line 24:
```suggestion
    pub interval_s: u32,
```

The config is translated through serde_json, so big ints are unreliable. But also, this duration shouldn't ever be that huge.

crates/gossip/src/gossip.rs
neonphog: woah... I didn't know you could comment on a whole file without a line number. It's cool that it shows up here even when the contents are collapsed.

crates/gossip/src/gossip.rs
neonphog: End users won't be able to do anything with that though... I'd say it should go the other way and be a debug.

crates/gossip/src/gossip.rs
```diff
@@ -26,132 +35,634 @@ impl K2GossipFactory {
...
 
-        let Some(msg) = msg.gossip_message else {
-            return Err(K2Error::other("no gossip message"));
-        };
+        let this = self.clone();
+        tokio::task::spawn(async move {
+            if let Err(e) = this.respond_to_msg(from_peer, msg).await {
+                tracing::error!("could not respond to gossip message: {:?}", e);
+            }
+        });
```
neonphog, line 268: Like an unbounded channel, unlimited tasks here could fill up our memory if the app is running slow. As you say, this module is rough, so doesn't have to be now, but at some point if gossip is overloading the system we need a relief valve to shut it down.

crates/gossip/src/state.rs
```diff
@@ -0,0 +1,194 @@
...
+        rand::thread_rng().fill_bytes(&mut session_id);
```
neonphog, line 37: I don't think this is working. The `with_capacity` will still have a `len()` of zero, so the slice here will be a zero-length slice that doesn't get any random bytes in it.

crates/gossip/proto/gossip.proto
```diff
@@ -4,46 +4,133 @@ package kitsune2.gossip;
...
-  // TODO Send initial DHT snapshot.
+  // The maximum number of bytes of new ops to respond with.
+  uint32 max_new_bytes = 21;
 }
 
 // A Kitsune2 gossip acceptance protocol message.
+//
+// Acceptable responses:
+// - `K2GossipNoDiffMessage`
 message K2GossipAcceptMessage {
```
neonphog, line 69: I'm still unsure about the intention here. On the one hand, this accept message has `new_since` and `max_new_bytes` fields, but on the other, the only acceptable response type is `NoDiffMessage`. If there are more message types coming later that will be valid, totally ignore this and I can be patient! : )

neonphog (changes requested):
Some changes requested, but looking good!

crates/api/src/agent.rs
ThetaSinner: Done -> https://github.com/holochain/kitsune2/pull/72/commits/52be4176fd0491287dc47c50f0fd45076907078d

crates/dht/src/arc_set.rs
ThetaSinner: Done -> https://github.com/holochain/kitsune2/pull/72/commits/8a54415ed535282a99c6f4381b7dce2e19e768dd

crates/api/src/op_store.rs
ThetaSinner: I don't follow. The timestamp bookmark is maintained per peer by each peer. So it's not going to move unpredictably. It's not a perfect mechanism because people will drop their bookmarks while offline and then we have to rely on the DHT module to get us back in sync.

Can we maybe discuss your example so I can figure out what it would be useful to add to this documentation?

crates/dht/src/arc_set.rs
ThetaSinner: Code replaced per comment above about compressing the arc sets

crates/gossip/proto/gossip.proto
ThetaSinner: To leave room for new fields and to allow grouping fields together here. This is how I'm used to working with protobuf, doesn't have to stay

crates/gossip/src/gossip.rs
ThetaSinner: Agree, it's not actionable by an end user. It's something we might track and block if somebody keeps doing it, in some fancy future version with a stronger security stance.

Changed to `debug!`

crates/gossip/src/peer_meta_store.rs
ThetaSinner: Yep, written while I was getting the module set up, not needed

crates/gossip/src/protocol.rs
ThetaSinner: Where else is sending agent ids over the wire?

crates/gossip/src/state.rs
ThetaSinner: Chosen to be 12 bytes, I thought that was enough to be reasonably unique. Should I change it to 16?

crates/gossip/src/gossip.rs
ThetaSinner: We could limit the number of concurrent incoming rounds and just return a busy message? That along with the number of messages that can be sent within a single round would limit the number of tasks that could be open

crates/gossip/src/state.rs
ThetaSinner: Always get this wrong with bytes... Fixed and added a unit test

crates/api/src/gossip.rs
ThetaSinner: Relying to this comment and the one below. Yes I had some trouble with circular references and that's why I'm holding a weak reference to it. I think splitting the local agents into their own little store would be a solution but I'm wondering how likely deadlocks are because of this? The only thing I was really concerned about was de-allocation working properly when both sides were using full arcs

crates/gossip/src/state.rs
neonphog: The capacity is in bytes, not bits. you have 96 here, not 12

crates/api/src/gossip.rs
neonphog: Okay, cool. I think a cloned local agent data type would be safer and easier to maintain, but I'm also fine with this and testing, especially as far as merging this pr goes. We can always update as a followon if needed.

neonphog (approved):

crates/api/src/op_store.rs
jost-s: Please FIXME

crates/api/src/op_store.rs
jost-s: To me it would be useful to have a comment such as
"The returned timestamp is kept in memory as a bookmark of up until when an agent has requested ops from a peer. If only a portion of all ops from `start` until now have been exchanged, the timestamp of the last exchanged op is returned. Otherwise all ops have already been exchanged and the time of the request serves as the bookmark."

The bit that I didn't understand at first was that the retrieval of ops is across _all_ ops until now.

crates/dht/src/arc_set.rs
jost-s: Which code is replaced? My suggestion was about the comment "a list of sectors indices".

crates/gossip/proto/gossip.proto
jost-s: Fine by me.

crates/gossip/src/config.rs
jost-s: So no?

crates/gossip/src/peer_meta_store.rs
jost-s: Do you want to keep it anyway?

crates/gossip/src/protocol.rs
jost-s: Not agent ids, but just ids, e.g. op ids. There's a `From<Bytes> for Id` and a `From<Id> for AgentId`.

crates/api/src/op_store.rs
ThetaSinner: This has been fixed

crates/api/src/op_store.rs
ThetaSinner: Alright, I see what you're getting at. I should be explaining that this is operating on the "store_at" timestamp so that this is a "paging" operation. I won't explain how gossip uses the function here, but I do want to make it clear how this function is meant to operate

crates/dht/src/arc_set.rs
ThetaSinner: This was an `into_raw` and a `from_raw`, but is now an `encode` and `decode` that has different documentation

crates/gossip/src/config.rs
ThetaSinner: I applied this change in a batch locally, sorry that doesn't mark it as resolved here. I can commit as batch on the UI if that's easier. I'm just aware that won't work with conventional commits at some point so I'm trying to work locally

crates/gossip/src/peer_meta_store.rs
ThetaSinner: No, I think removing it is the right call, it's gone

crates/gossip/src/protocol.rs
ThetaSinner: Oh I get you! Yeah that's a nice idea

ThetaSinner: Done here -> https://github.com/holochain/kitsune2/pull/72/commits/e61dfbc3645efb8ebe2fcde43bf289bc8f56af2e

crates/api/src/op_store.rs
ThetaSinner: Updated here -> https://github.com/holochain/kitsune2/pull/72/commits/e61dfbc3645efb8ebe2fcde43bf289bc8f56af2e

crates/gossip/src/config.rs
jost-s: Ah okay, good to know, then just "carry on" =)

jost-s (approved):
Looks great

PR #73, opened 2025-01-15
refactor(core): add configurable delay to fetch request queue

jost-s:
Adds a configurable delay to fetch request queue (as it says above), which passes before re-inserting a just sent request back into the outgoing request queue.

### DIFF
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -73,6 +73,7 @@ use message_handler::FetchMessageHandler;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 use tokio::{
     sync::mpsc::{channel, Receiver, Sender},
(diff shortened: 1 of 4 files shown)


crates/core/src/factories/core_fetch.rs
```diff
@@ -97,6 +98,9 @@ pub mod config {
     pub struct CoreFetchConfig {
         /// How many parallel op fetch requests can be made at once. Default: 2.  
         pub parallel_request_count: u8,
+        /// Delay before re-inserting ops to request back into the outgoing request queue.
+        /// Default: 100 ms.
+        pub re_insert_outgoing_request_delay: Duration,
```
ThetaSinner, line 103: Don't we need to be careful that these can deserialize from YAML and use `_ms` or is that not a concern here?

ThetaSinner (approved):
Looks good, one question about config but that can be worked out later if needed :)

crates/core/src/factories/core_fetch.rs
jost-s: I do not know. I remember there was something yamly. Durations can be deserialized from YAML, but it's not so user-friendly.
I'll just convert all to ms.

jost-s:
> Looks good, one question about config but that can be worked out later if needed :)

How about now?

PR #74, opened 2025-01-16
Op store split metadata

ThetaSinner:
The op store should ingest only op data. Hashes should be computed when data arrives so that we aren't trusting somebody else to do our hashing. The rest of this PR is consequences from trying to achieve that.

Within the op store, we want to be able to store more fields than the op contains. Like `stored_at` which I'm using on the gossip PR for incremental sync. That means we need to keep the op data model and the model the op store uses internally as separate types.

The big fiddle here was that I couldn't use the hash function that was added with the fetch queue inside the mem op store without destroying the DHT tests. Those need predictable IDs to get the ops into different parts of the DHT to test the location logic. I could spend time trying to find op data that places the ops in the right locations but predictable IDs seemed so much simpler for testing. So I'm just using the first 32 bytes of the op data as the id. If there isn't enough op data, then pad with 0.

### DIFF
--- a/crates/api/src/fetch.rs
+++ b/crates/api/src/fetch.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     builder, config, peer_store::DynPeerStore, transport::DynTransport,
-    AgentId, BoxFut, DynOpStore, K2Result, MetaOp, OpId, SpaceId,
+    AgentId, BoxFut, DynOpStore, K2Result, OpId, SpaceId,
 };
 use bytes::{Bytes, BytesMut};
 use k2_fetch_message::FetchMessageType;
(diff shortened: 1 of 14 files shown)


crates/api/src/fetch.rs
```diff
@@ -57,17 +57,17 @@ pub fn serialize_request_message(value: Vec<OpId>) -> Bytes {
...
     Response::from(value)
         .encode(&mut out)
         .expect("failed to encode response");
     out.freeze()
 }
 
-/// Serialize list of ops to fetch resopnse message.
-pub fn serialize_response_message(value: Vec<MetaOp>) -> Bytes {
-    let mut out = bytes::BytesMut::new();
+/// Serialize list of ops to fetch response message.
```
jost-s, line 68: I just love my resopnes!

jost-s (approved):
LGTM

That hashing works okay, I think. I was thinking of putting the whole MemOpStore under `#[cfg(test))`, but we need a replacement for it. Were it test code, we could use some actual hashing algorithm.

ThetaSinner:
Yeah I'm thinking of it as test code. We could and probably should have a more realistic host implementation for integration tests. Once we have tx5 integrated and the top level crate created, I'm happy to add an op store that behaves in a more realistic way

PR #75, opened 2025-01-20
Tx5Transport Module

neonphog:
I've ignored the `peer_connect_disconnect` test due to the following issue: https://github.com/holochain/tx5/issues/119. I think that issue may be a low-ish priority, since we don't yet have any logic that depends on the disconnect events/callbacks.

### DIFF
--- a/crates/transport_tx5/Cargo.toml
+++ b/crates/transport_tx5/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "kitsune2_transport_tx5"
+version = "0.0.1-alpha"
+description = "The Tx5-based transport implementation for kitsune2"
+license = "Apache-2.0"
+homepage = "https://github.com/holochain/kitsune2"
+documentation = "https://docs.rs/kitsune2_transport_tx5"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "holo", "p2p", "dht", "networking"]
+categories = ["network-programming"]
+edition = "2021"
+
+[dependencies]
+bytes = { workspace = true }
+kitsune2_api = { workspace = true }
+tokio = { workspace = true, features = ["rt", "sync", "time"] }
+tracing = { workspace = true }
+tx5 = { version = "0.2.0-beta", default-features = false, features = ["backend-libdatachannel"] }
--- a/crates/transport_tx5/Cargo.toml
+++ b/crates/transport_tx5/Cargo.toml
@@ -13,6 +13,7 @@ edition = "2021"
 [dependencies]
 bytes = { workspace = true }
 kitsune2_api = { workspace = true }
+serde = { workspace = true }
 tokio = { workspace = true, features = ["rt", "sync", "time"] }
 tracing = { workspace = true }
 tx5 = { version = "0.2.0-beta", default-features = false, features = [
--- a/crates/transport_tx5/Cargo.toml
+++ b/crates/transport_tx5/Cargo.toml
@@ -22,5 +22,6 @@ tx5 = { version = "0.2.0-beta", default-features = false, features = [
 
 [dev-dependencies]
 kitsune2_core = { workspace = true }
+kitsune2_test_utils = { workspace = true }
 sbd-server = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
--- a/crates/transport_tx5/Cargo.toml
+++ b/crates/transport_tx5/Cargo.toml
@@ -19,3 +19,8 @@ tracing = { workspace = true }
 tx5 = { version = "0.2.0-beta", default-features = false, features = [
   "backend-libdatachannel",
 ] }
+
+[dev-dependencies]
+kitsune2_core = { workspace = true }
+sbd-server = { workspace = true }
+tokio = { workspace = true, features = ["full"] }
--- a/crates/transport_tx5/Cargo.toml
+++ b/crates/transport_tx5/Cargo.toml
@@ -15,4 +15,6 @@ bytes = { workspace = true }
 kitsune2_api = { workspace = true }
 tokio = { workspace = true, features = ["rt", "sync", "time"] }
 tracing = { workspace = true }
-tx5 = { version = "0.2.0-beta", default-features = false, features = ["backend-libdatachannel"] }
+tx5 = { version = "0.2.0-beta", default-features = false, features = [
+  "backend-libdatachannel",
+] }
--- a/crates/transport_tx5/src/lib.rs
+++ b/crates/transport_tx5/src/lib.rs
@@ -1,4 +1,5 @@
-//! The core stub transport implementation provided by Kitsune2.
+#![deny(missing_docs)]
+//! kitsune2 tx5 transport module.
 
 use kitsune2_api::{config::*, transport::*, *};
 use std::sync::Arc;
@@ -54,9 +55,7 @@ pub mod config {
 
 use config::*;
 
-/// The core stub transport implementation provided by Kitsune2.
-/// This is NOT a production module. It is for testing only.
-/// It will only establish "connections" within the same process.
+/// Provides a Kitsune2 transport module based on the Tx5 crate.
 #[derive(Debug)]
 pub struct Tx5TransportFactory {}
 
@@ -120,6 +119,7 @@ impl Tx5Transport {
             signal_allow_plain_text: true,
             preflight: Some((
                 Arc::new(move |peer_url| {
+                    // gather any preflight data, and send to remote
                     let handler = preflight_send_handler.clone();
                     let peer_url = peer_url.to_k();
                     Box::pin(async move {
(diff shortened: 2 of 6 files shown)


crates/transport_tx5/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+                    .await?;
+            Ok(DefaultTransport::create(&handler, imp))
+        })
+    }
+}
+
+#[derive(Debug)]
+struct Tx5Transport {
+    ep: Arc<tx5::Endpoint>,
+    pre_task: tokio::task::JoinHandle<()>,
```
ThetaSinner, line 95: Get the `AbortHandle` from the join handle when there's no intention to `.join` the tasks? Minor thing, really just semantics

crates/transport_tx5/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+impl Tx5Transport {
+    pub async fn create(
+        config: Tx5TransportConfig,
+        handler: Arc<TxImpHnd>,
+    ) -> K2Result<DynTxImp> {
+        let (pre_send, pre_recv) = tokio::sync::mpsc::channel::<PreCheck>(1024);
+
+        let preflight_send_handler = handler.clone();
+        let tx5_config = Arc::new(tx5::Config {
+            signal_allow_plain_text: true,
```
ThetaSinner, line 119: Is this just permitted for testing and it will always use TLS with `wss` SBD urls?

crates/transport_tx5/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+            if let Ok(peer) = tx5::PeerUrl::parse(&peer) {
+                self.ep.close(&peer);
+            }
+        })
+    }
+
+    fn send(&self, peer: Url, data: bytes::Bytes) -> BoxFut<'_, K2Result<()>> {
+        Box::pin(async move {
+            let peer = peer.to_peer_url()?;
+            // this would be more efficient if we retool tx5 to use bytes
```
ThetaSinner, line 216: Should this be a TODO or a task?

crates/transport_tx5/src/test.rs
```diff
@@ -0,0 +1,407 @@
...
+use std::sync::Mutex;
+
+// We don't need or want to test all of tx5 in here... that should be done
+// in the tx5 repo. Here we should only test the kitsune2 integration of tx5.
+//
+// Specifically:
+//
+// - That new_listening_address is called if the sbd server is restarted
+// - That peer connect / disconnect are invoked appropriately.
+// - That messages can be send / received.
```
ThetaSinner, line 11:
```suggestion
// - That messages can be sent / received.
```

crates/transport_tx5/src/test.rs
```diff
@@ -0,0 +1,407 @@
...
+                },
+            })
+            .unwrap();
+
+        this.builder = Arc::new(builder);
+
+        this
+    }
+
+    pub async fn restart(&mut self) {
```
ThetaSinner, line 51: On first start this will use port `0` and on subsequent restarts we need to re-use the port that we assigned the first time so that the transport config doesn't need updating for peers to reconnect?

A quick doc comment about what this is doing might be useful. If something does happen to claim the port while we're restarting (not that likely) then we could see the occasional flake from this and it'd be nice to have a doc comment to remind us what it does.

crates/transport_tx5/src/test.rs
```diff
@@ -0,0 +1,407 @@
...
+        peer: Url,
+        space: SpaceId,
+        module: String,
+        data: bytes::Bytes,
+    ) -> K2Result<()> {
+        (self.module)(peer, space, module, data)
+    }
+}
+
+const S1: SpaceId = SpaceId(id::Id(bytes::Bytes::from_static(b"space1")));
```
ThetaSinner, line 189: There is now a test space in the test_utils module that could be used here

crates/transport_tx5/src/test.rs
```diff
@@ -0,0 +1,407 @@
...
+
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+        }
+    })
+    .await
+    .unwrap();
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "disconnects currently broken in tx5"]
```
ThetaSinner, line 225: Something that needs a follow up issue?

ThetaSinner (commented):
A handful of minor things. This is looking good and the testing looks ace :)

crates/transport_tx5/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+    #[serde(rename_all = "camelCase")]
+    pub struct Tx5TransportConfig {
+        /// The url of the sbd signal server. E.g. `wss://sbd.kitsu.ne`.
+        pub server_url: String,
+    }
+
+    impl Default for Tx5TransportConfig {
+        fn default() -> Self {
+            Self {
+                server_url: "<wss://your.sbd.url>".into(),
```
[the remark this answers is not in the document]
neonphog: My hope was to be able to generate a default conductor config with these defaults in there. I feel like we're going to want an example like this in there when we do that. But I'm open to other suggestions.

crates/transport_tx5/src/lib.rs
neonphog: As you say exactly. I can carry this forward into the config, since we'll eventually want that anyways.

crates/transport_tx5/src/lib.rs
neonphog: This is part of the refactor tx5 option. We'd decided not to do it now. Do you think we should create an issue for it?

crates/transport_tx5/src/test.rs
neonphog: [already made : )](https://github.com/holochain/tx5/issues/119)

crates/transport_tx5/src/lib.rs
```diff
@@ -0,0 +1,311 @@
+#![deny(missing_docs)]
+//! kitsune2 tx5 transport module.
+
+use kitsune2_api::{config::*, transport::*, *};
+use std::sync::Arc;
+
+trait PeerUrlExt {
+    fn to_k(&self) -> K2Result<Url>;
```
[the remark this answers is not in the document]
neonphog: Done: https://github.com/holochain/kitsune2/pull/75/commits/0db5652107f1096b610db1b68c9ad72cd6b2c7fe

crates/transport_tx5/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/75/commits/947666b6b597c8b964531fb9115a696add002006

crates/transport_tx5/src/lib.rs
ThetaSinner: I think it'd be good to track it even if we aren't going to do it right now

crates/transport_tx5/src/lib.rs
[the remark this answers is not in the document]
ThetaSinner: That's fair, we can enforce this higher up. I'm wondering whether Kitsune2 could reject a configuration that doesn't provide a value for this

crates/transport_tx5/src/lib.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/75/commits/e281c2bc6e603afff7f03884232bb495c94e6bb6

crates/transport_tx5/src/test.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/75/commits/3fcf51a5a7a02739badd0e83a54021941b7bd412

crates/transport_tx5/src/test.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/75/commits/a52181013a9e7eeab19fde3c297632256b24c4a0

crates/transport_tx5/src/lib.rs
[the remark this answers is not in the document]
ThetaSinner: Hmm, the config callback we have is just dealing with setting defaults right? So it would have to be the `create`s on the factories that would need to check this. That would work and yes Holochain gets the feedback. 

I do quite like the idea of a validate option so that we have a clear place to look for config checking rather than hunting through the `create` code when instances are created. Oh and that also means we should have some easy to call code when we get to updating config. Yeah I think I like that option more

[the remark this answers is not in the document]
ThetaSinner: Probably not a job for this PR then

ThetaSinner (dismissed):
💥 :)

jost-s (approved):

crates/transport_tx5/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "kitsune2_transport_tx5"
+version = "0.0.1-alpha"
+description = "The Tx5-based transport implementation for kitsune2"
+license = "Apache-2.0"
+homepage = "https://github.com/holochain/kitsune2"
+documentation = "https://docs.rs/kitsune2_transport_tx5"
+authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
+keywords = ["holochain", "holo", "p2p", "dht", "networking"]
```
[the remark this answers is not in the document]
neonphog: Indeed, we should probably make a more concerted audit of the Cargo.tomls before we publish for the first time.

crates/transport_tx5/src/lib.rs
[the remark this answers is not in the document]
neonphog: https://github.com/holochain/kitsune2/issues/83

PR #76, opened 2025-01-21
Integrate fetch with gossip

ThetaSinner:
Most of this is renaming because the gossip isn't working with `AgentId`s, it's working with `Url`s.

When we gossip, we aren't gossiping with a single agent. I could list all peers and map a peer url back to an agent id... but then that agent might move and just because we can use the agent id to look up a peer url, that doesn't mean it's the right one.
On the other hand, when we store a peer url, there's no guarantee that the URL stays valid.

I think it's slightly better to fetch from a specific node by its `Url` and have to drop requests if that peer isn't handing us back ops, than it is to track by agent id which may or may not provide a Url that has the ops we originally discovered.

### DIFF
--- a/crates/api/src/fetch.rs
+++ b/crates/api/src/fetch.rs
@@ -1,8 +1,8 @@
 //! Kitsune2 fetch types.
 
 use crate::{
-    builder, config, peer_store::DynPeerStore, transport::DynTransport,
-    AgentId, BoxFut, DynOpStore, K2Result, OpId, SpaceId,
+    builder, config, transport::DynTransport, BoxFut, DynOpStore, K2Result,
+    OpId, SpaceId, Url,
 };
 use bytes::{Bytes, BytesMut};
 use k2_fetch_message::FetchMessageType;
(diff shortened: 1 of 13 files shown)


jost-s:
Fetch requests come out of a gossip session which is based on a url to a peer. I agree the url seems more reliable than an agent id.
Glancing at the PR it seems you're equating "agent" with "node" and therefore change all the occurrences of the word "agent" to "peer", right?

ThetaSinner:
That's a good question, another case of me not being quite clear on terminology :)

I have been using the name `peer_url` with the `Url` type because a `Url` could be anything and I want to remind myself and readers of my code what the `Url` is for.
The `Url` doesn't translate directly to an agent, technically it translates to the address of 1 or more agents. So I thought it best to rename references to agent to peer.
Whether that was best is an open question and I can do something else!

jost-s (approved):
Clear, pretty and easy on the eyes. You're just making this simpler and fetch is working with gossip - how could I not approve of that?!

neonphog (approved):
Woot

PR #77, opened 2025-01-21
Split local agent store out of the space

ThetaSinner:
This is to resolve the comment thread from here -> https://github.com/holochain/kitsune2/pull/72#discussion_r1919313679

We don't want the two-way dependency between the space and the gossip instance it creates. This breaks the dependency by splitting the local agent store into its own entity that they can both share. That was the only reason for the gossip module to depend on the space.

Note that the gossip tests still need the space to get at the "local agent join" logic, which I believe does belong in the space. But the gossip implementation only depends on the local agent store.


crates/api/src/transport.rs
```diff
@@ -31,8 +31,11 @@ impl TxImpHnd {
 
     /// Call this when you receive or bind a new address at which
     /// this local node can be reached by peers
-    pub fn new_listening_address(&self, this_url: Url) {
-        self.handler.new_listening_address(this_url);
+    pub fn new_listening_address(
+        &self,
+        this_url: Url,
+    ) -> BoxFut<'static, K2Result<()>> {
+        self.handler.new_listening_address(this_url)
```
neonphog, line 38: The pattern for the handlers (for peer connect and recv data) is to return an error when the connection should be closed based on the data received (preflight checks in connect, and decoding or validation problems with recv). It doesn't make sense to me to have the implementer return an error into the library here (it's not going to shut down the whole endpoint because the client didn't like the address that was bound). What are your thoughts?

neonphog (commented):
I *think* I vote for the new_listening_address to return `BoxFut<'static, ()>` instead of the result. But I want to hear your thoughts first.

Otherwise, this looks awesome!

jost-s (approved):
I _think_ I agree with David.

jost-s (approved):

PR #80, opened 2025-01-22
Add gossip tasks

ThetaSinner:
Adds initiate and timeout tasks for gossip.

The initiate task is responsible for selecting peers to gossip with. The logic there is a little complex but hopefully reasonably well described by tests.

The timeout task is super simple. If a round times out, its state gets dropped. That frees Kitsune up to do something else and if we receive further messages for that round then we don't find state for it and we'll reject it.

### DIFF
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -38,6 +38,17 @@ pub struct K2GossipConfig {
     ///
     /// Default: 300,000 (5m)
     pub min_initiate_interval_ms: u32,
+
+    /// The timeout for a gossip round.
+    ///
+    /// This will be loosely enforced on both sides. Kitsune periodically checks for timed out
+    /// rounds and will terminate them if they have been running for longer than this timeout.
+    ///
+    /// There is no point setting this lower than 5s because that is how often Kitsune checks for
+    /// timed out rounds.
+    ///
+    /// Default: 60,000 (1m)
+    pub gossip_timeout_ms: u32,
 }
 
 impl Default for K2GossipConfig {
@@ -46,6 +57,7 @@ impl Default for K2GossipConfig {
             max_gossip_op_bytes: 100 * 1024 * 1024,
             initiate_interval_ms: 120_000,
             min_initiate_interval_ms: 300_000,
+            gossip_timeout_ms: 60_000,
         }
     }
 }
@@ -60,6 +72,11 @@ impl K2GossipConfig {
     /// initiated by a given peer.
     pub(crate) fn min_initiate_interval(&self) -> std::time::Duration {
         std::time::Duration::from_millis(self.min_initiate_interval_ms as u64)
+    }
+
+    /// The timeout for a gossip round.
+    pub(crate) fn gossip_timeout(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.gossip_timeout_ms as u64)
     }
 }
 
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -18,25 +18,48 @@ pub struct K2GossipConfig {
     /// Default: 100MB
     pub max_gossip_op_bytes: u32,
 
-    /// The interval in seconds between gossip rounds.
+    /// The interval in seconds between initiating gossip rounds.
+    ///
+    /// This controls how often Kitsune will attempt to find a peer to gossip with.
+    ///
+    /// This can be set as low as you'd like, but you will still be limited by
+    /// [K2GossipConfig::min_initiate_interval_ms]. So a low value for this will result in Kitsune
+    /// doing its gossip initiation in a burst. Then, when it has run out of peers, it will idle
+    /// for a while.
+    ///
+    /// Default: 120,000 (2m)
+    pub initiate_interval_ms: u32,
+
+    /// The minimum amount of time that must be allowed to pass before a gossip round can be
+    /// initiated by a given peer.
+    ///
+    /// This is a rate-limiting mechanism to be enforced against incoming gossip and therefore must
+    /// be respected when initiating too.
     ///
     /// Default: 300,000 (5m)
-    pub interval_ms: u32,
+    pub min_initiate_interval_ms: u32,
 }
 
 impl Default for K2GossipConfig {
     fn default() -> Self {
         Self {
             max_gossip_op_bytes: 100 * 1024 * 1024,
-            interval_ms: 300_000,
+            initiate_interval_ms: 120_000,
+            min_initiate_interval_ms: 300_000,
         }
     }
 }
 
 impl K2GossipConfig {
-    /// The interval between gossip rounds.
-    pub(crate) fn interval(&self) -> std::time::Duration {
-        std::time::Duration::from_millis(self.interval_ms as u64)
+    /// The interval between initiating gossip rounds.
+    pub(crate) fn initiate_interval(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.initiate_interval_ms as u64)
+    }
+
+    /// The minimum amount of time that must be allowed to pass before a gossip round can be
+    /// initiated by a given peer.
+    pub(crate) fn min_initiate_interval(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.min_initiate_interval_ms as u64)
     }
 }
 
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -48,7 +48,7 @@ pub struct K2GossipConfig {
     /// timed out rounds.
     ///
     /// Default: 60,000 (1m)
-    pub gossip_timeout_ms: u32,
+    pub round_timeout_ms: u32,
 }
 
 impl Default for K2GossipConfig {
@@ -57,7 +57,7 @@ impl Default for K2GossipConfig {
             max_gossip_op_bytes: 100 * 1024 * 1024,
             initiate_interval_ms: 120_000,
             min_initiate_interval_ms: 300_000,
-            gossip_timeout_ms: 60_000,
+            round_timeout_ms: 60_000,
         }
     }
 }
@@ -75,8 +75,8 @@ impl K2GossipConfig {
     }
 
     /// The timeout for a gossip round.
-    pub(crate) fn gossip_timeout(&self) -> std::time::Duration {
-        std::time::Duration::from_millis(self.gossip_timeout_ms as u64)
+    pub(crate) fn round_timeout(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.round_timeout_ms as u64)
     }
 }
 
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -20,15 +20,15 @@ pub struct K2GossipConfig {
 
     /// The interval in seconds between gossip rounds.
     ///
-    /// Default: 300 (5m)
-    pub interval_s: u32,
+    /// Default: 300,000 (5m)
+    pub interval_ms: u32,
 }
 
 impl Default for K2GossipConfig {
     fn default() -> Self {
         Self {
             max_gossip_op_bytes: 100 * 1024 * 1024,
-            interval_s: 300,
+            interval_ms: 300_000,
         }
     }
 }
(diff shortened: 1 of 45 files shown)


crates/test_utils/src/agent.rs
```diff
@@ -144,7 +156,7 @@ impl AgentBuilder {
         });
         let is_tombstone = self.is_tombstone.unwrap_or(false);
         let url = self.url.unwrap_or(None);
-        let storage_arc = self.storage_arc.unwrap_or(DhtArc::FULL);
+        let storage_arc = self.storage_arc.unwrap_or_default();
```
ThetaSinner, line 159: I want to be able to set this in tests and the default is `Empty`. Didn't break any existing tests so I'm assuming this value wasn't too important to anyone else yet

crates/gossip/src/timeout.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    accepted_round_states.lock().await.retain(|_, state| {
+        match state.started_at.elapsed() > round_timeout {
+            true => {
+                tracing::warn!("Accepted round timed out: {:?}", state);
+                false
+            }
+            false => true,
+        }
```
neonphog, line 58: Is an if a little easier to read here?

```suggestion
        if state.started_at.elapsed() > round_timeout {
            tracing::warn!("Accepted round timed out: {:?}", state);
            false
        } else {
            true
        }
```

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+) -> AbortHandle {
+    tracing::info!("Starting initiate task");
+
+    let initiate_interval = config.initiate_interval();
+    let min_initiate_interval = config.min_initiate_interval();
+    tokio::task::spawn(async move {
+        loop {
+            tokio::time::sleep(initiate_interval).await;
+
+            match select_next_target(
```
neonphog, line 25: We're not checking if we have an active initiate round here. Is that because we assume it worst-case will have timed out during the initiate_interval period?

neonphog (dismissed):
woot, this looks awesome!

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+                harness.local_agent_store.clone(),
+                harness.peer_meta_store.clone(),
+            )
+            .await
+            .unwrap();
+
+            seen.insert(url.unwrap());
+        }
+
+        assert_eq!(2, seen.len());
```
jost-s, line 546:
```suggestion
        assert_eq!(2, seen.len());
        assert!(seen.contains(agent_1_url));
        assert!(seen.contains(agent_2_url));
```

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+    }
+
+    // Filter local agents out of the list of all agents
+    remove_local_agents(&mut all_agents, &local_agent_ids);
+
+    // There are no agents with an overlapping arc to gossip with. We should cast the net
+    // wider and gossip with agents that might still be growing their arc and have some ops
+    // or agent infos that we are missing
+    if all_agents.is_empty() {
+        all_agents = peer_store.get_all().await?.into_iter().collect();
```
jost-s, line 97:
```suggestion
    if all_agents.is_empty() {
        // There are no agents with an overlapping arc to gossip with. We should cast the net
        // wider and gossip with agents that might still be growing their arc and have some ops
        // or agent infos that we are missing
        all_agents = peer_store.get_all().await?.into_iter().collect();
```

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+            .unwrap();
+
+            seen.insert(url.unwrap());
+        }
+
+        assert_eq!(2, seen.len());
+    }
+
+    #[tokio::test]
+    async fn prioritises_overlapping_arc() {
```
jost-s, line 550:
```suggestion
    async fn selects_non_overlapping_when_no_overlapping_peers_available() {
```

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+        all_agents.extend(by_local_arc_agents);
+    }
+
+    // Filter local agents out of the list of all agents
+    remove_local_agents(&mut all_agents, &local_agent_ids);
+
+    // There are no agents with an overlapping arc to gossip with. We should cast the net
+    // wider and gossip with agents that might still be growing their arc and have some ops
+    // or agent infos that we are missing
+    if all_agents.is_empty() {
```
jost-s, line 96: How about if this depends on timestamp too? Right now if there are peers with an overlapping arc of my target arc and I'm caught up with them, I'm not gossiping with peers outside of my arc. Instead I could gossip with them if I'm caught up with overlapping peers.

jost-s (changes requested):
Looks fantastic! Nice test suite. There's one question about who to gossip with that is meant as an exploration rather than a suggestion.

crates/gossip/src/initiate.rs
jost-s: I was about to suggest to add another peer that does have the overlapping arc, but then realized that the non-overlapping peers will never be selected then, regardless of timestamp.

crates/gossip/src/config.rs
```diff
@@ -18,25 +18,65 @@ pub struct K2GossipConfig {
...
+
+    /// The minimum amount of time that must be allowed to pass before a gossip round can be
+    /// initiated by a given peer.
+    pub(crate) fn min_initiate_interval(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.min_initiate_interval_ms as u64)
+    }
+
+    /// The timeout for a gossip round.
+    pub(crate) fn round_timeout(&self) -> std::time::Duration {
+        std::time::Duration::from_millis(self.round_timeout_ms as u64)
```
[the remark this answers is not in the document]
ThetaSinner: I'm not sure. Maybe? I'm assuming that the config being public might actually be useful in the production factory? The other config modules have made both their struct and their fields `pub`

crates/gossip/src/timeout.rs
```diff
@@ -0,0 +1,155 @@
...
+
+async fn remove_timed_out_rounds(
+    round_timeout: Duration,
+    initiated_round_state: Arc<Mutex<Option<GossipRoundState>>>,
+    accepted_round_states: Arc<Mutex<HashMap<Url, GossipRoundState>>>,
+) {
+    {
+        let mut initiated_state = initiated_round_state.lock().await;
+        match initiated_state.as_ref() {
+            Some(state) if state.started_at.elapsed() > round_timeout => {
```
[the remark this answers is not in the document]
ThetaSinner: Prefer something else? It is a little funky

crates/gossip/src/gossip.rs
```diff
@@ -180,16 +183,17 @@ impl K2Gossip {
             Arc::new(gossip.clone()),
         );
 
+        spawn_initiate_task(gossip.config.clone(), gossip.clone());
+        spawn_timeout_task(gossip.config.clone(), gossip.clone());
```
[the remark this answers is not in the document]
ThetaSinner: Yes, it absolutely is! Oversight on my part

crates/gossip/src/timeout.rs
```diff
@@ -0,0 +1,155 @@
...
+use crate::K2GossipConfig;
+use kitsune2_api::Url;
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::Mutex;
+use tokio::task::AbortHandle;
+
+/// Spawns a task that checks for timed-out gossip rounds and drops their state.
+pub(crate) fn spawn_timeout_task(
```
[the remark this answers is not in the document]
ThetaSinner: Yeah this might be a good point for the refactor of this module. I'm thinking of breaking the K2Gossip implementation over several files. Given this takes it as an input, it may as well be `self`

crates/gossip/src/initiate.rs
```diff
@@ -0,0 +1,646 @@
...
+            )
+            .await;
+
+        harness
+            .new_remote_agent(
+                Some(Url::from_str("ws://test:80/2").unwrap()),
+                Some(DhtArc::FULL),
+                Some(DhtArc::FULL),
+            )
+            .await;
```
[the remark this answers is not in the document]
ThetaSinner: The first agent isn't supposed to have the same URL. It'll be filtered out by agent id but I have corrected the test data

crates/gossip/src/initiate.rs
ThetaSinner: Yeah that was the original intention of this test which is why it's confusingly named :)

crates/gossip/src/initiate.rs
ThetaSinner: Unless there is an `else` with the `if`, I would generally comment on the outside rather than the inside

crates/gossip/src/initiate.rs
ThetaSinner: Yes that is worth doing -> https://github.com/holochain/kitsune2/pull/80/commits/d8b25041aa642c03fa0b389114fcce0a6541c9b6

I was generally concerned about what happens if you get stuck in a set of peers that you know you have an overlapping arc with and don't discover more peers. You could end up with an isolated group that are only gossiping with each other. I'm not sure this is a complete solution to that problem because in a bigger network, one of the overlapping arc peers will likely have come off timeout... So you aren't likely to hit the case in this code.

Something to think about later. I think for now we can rely on some peers restarting sometimes and having to rediscover agents on the network and then that newly started peer should gossip agent infos to the isolated group and get everyone talking again... I think that's a reasonable description of how this would behave currently.

crates/gossip/src/initiate.rs
ThetaSinner: Renamed back, in light of the change below, this is now relevant again

crates/gossip/src/gossip.rs
```diff
@@ -91,29 +93,32 @@ impl Drop for DropAbortHandle {
...
+    pub(crate) peer_store: DynPeerStore,
+    pub(crate) local_agent_store: DynLocalAgentStore,
+    pub(crate) peer_meta_store: Arc<K2PeerMetaStore>,
     op_store: DynOpStore,
     fetch: DynFetch,
     agent_verifier: DynVerifier,
     response_tx: Sender<GossipResponse>,
     _response_task: Arc<DropAbortHandle>,
+    _initiate_task: Arc<Option<DropAbortHandle>>,
+    _timeout_task: Arc<Option<DropAbortHandle>>,
```
jost-s, line 121: This task spawning business leads to strange fields, same in the fetch module.

crates/gossip/src/timeout.rs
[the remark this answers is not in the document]
jost-s: No, I didn't know this was possible.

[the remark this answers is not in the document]
jost-s: Sometimes I'm not being sarcastic ;-)

jost-s (approved):

crates/gossip/src/config.rs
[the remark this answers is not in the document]
jost-s: I mean in the sense of removing the getters and just making the fields accessible.

jost-s:
Awesome!

crates/gossip/src/gossip.rs
ThetaSinner: It does, yes. They're really just there to link the life cycles together but aren't useful in their own right... If I think of a nicer way to express the same thing I will go for it, but I have no ideas at the moment :)

crates/gossip/src/config.rs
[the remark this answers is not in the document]
ThetaSinner: Oh, I see. The fields to be serializable as arbitrary fields in something like JSON or YAML. These are returning `Duration`s because they're easier to work with. So these are "getters that do something" rather than just being there to hide the fields :)

ThetaSinner:
It wasn't, there's a tx5 test failure on here. I'm assuming that's a flake

crates/gossip/src/config.rs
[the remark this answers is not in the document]
jost-s: Ah, right, I had missed that.

PR #81, opened 2025-01-23
fix(core): increase timeouts in fetch tests to prevent flakiness

jost-s:
I hope tests will be steady now on Windows.

### DIFF
--- a/crates/test_utils/src/lib.rs
+++ b/crates/test_utils/src/lib.rs
@@ -3,6 +3,10 @@
 
 use rand::RngCore;
 
+pub mod agent;
+pub mod id;
+pub mod space;
+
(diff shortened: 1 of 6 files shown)


crates/test_utils/src/lib.rs
```diff
@@ -25,6 +29,26 @@ pub fn random_bytes(length: u16) -> Vec<u8> {
     bytes
 }
 
-pub mod agent;
-pub mod id;
-pub mod space;
+/// Repeat a code block after a pause until a timeout has elapsed.
+/// The default timeout is 100 ms.
+#[macro_export]
+macro_rules! iter_check {
```
jost-s, line 35: Moved this from space tests.

ThetaSinner (approved):

PR #85, opened 2025-01-28
feat(kitsune2): add top-level crate kitsune2

### LINKED ISSUE #79: [k2] Add default production builder
Create a new crate `kitsune2` with production builder. The production builder uses production-ready core modules.

AC:
- Add an integration test that instantiates Kitsune and tests successful gossiping between 2 peers.

jost-s:
Adds a top-level crate called "kitsune2" with a production default builder. There's a full integration test using the builder.

The test is flaky with higher number of ops. ~~and I'm asking here first before going about the debugging myself. When the number of ops to be exchanged during gossip is increased to about 100 each, the following error happens almost always:~~

Fixing the error will be part of another branch.

```sh
2025-01-28T18:37:32.101834Z DEBUG NETAUDIT: pub_key=0HDL-1H_0bmwwx7XZUWY6k39pRM9OSHLWwIs_P2c6xE byte_count=195 err=Kind(BrokenPipe) m="tx5-connection" a="send_framed_error"
2025-01-28T18:37:32.101862Z  WARN kitsune2_core::factories::core_fetch: could not send ops to requesting peer: tx5 send error (src: broken pipe) op_ids=[qLkchkQ6c3wdv8e8z2edk6OAjSQFbHIEwMr7vXYiEfM] peer=ws://127.0.0.1:38851/0HDL-1H_0bmwwx7XZUWY6k39pRM9OSHLWwIs_P2c6xE
```

Famous error that we spotted in Holochain too, I believe.

resolved #79

### DIFF
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -14,9 +14,9 @@ use kitsune2_api::fetch::DynFetch;
 use kitsune2_api::peer_store::DynPeerStore;
 use kitsune2_api::transport::{DynTransport, TxBaseHandler, TxModuleHandler};
 use kitsune2_api::{
-    DynGossip, DynGossipFactory, DynLocalAgentStore, DynOpStore,
+    BoxFut, DynGossip, DynGossipFactory, DynLocalAgentStore, DynOpStore,
     DynPeerMetaStore, Gossip, GossipFactory, K2Error, K2Result, SpaceId,
-    Timestamp, Url, UNIX_TIMESTAMP,
+    StoredOp, Timestamp, Url, UNIX_TIMESTAMP,
 };
 use kitsune2_dht::{Dht, DhtApi};
 use std::collections::HashMap;
@@ -63,11 +63,12 @@ impl GossipFactory for K2GossipFactory {
         fetch: DynFetch,
     ) -> kitsune2_api::BoxFut<'static, K2Result<DynGossip>> {
         Box::pin(async move {
-            let config: K2GossipConfig = builder.config.get_module_config()?;
+            let config =
+                builder.config.get_module_config::<K2GossipModConfig>()?;
 
             let gossip: DynGossip = Arc::new(
                 K2Gossip::create(
-                    config,
+                    config.k2_gossip,
                     space_id,
                     peer_store,
                     local_agent_store,
@@ -273,7 +274,7 @@ impl K2Gossip {
             max_new_bytes: self.config.max_gossip_op_bytes,
         };
 
-        // Right before we send the initiate message, check whether the target has already
+        // Before we send the initiate message, check whether the target has already
         // initiated with us. If they have, we can just skip initiating.
         if self
             .accepted_round_states
@@ -323,7 +324,16 @@ impl K2Gossip {
     }
 }
 
-impl Gossip for K2Gossip {}
+impl Gossip for K2Gossip {
+    fn inform_ops_stored(
+        &self,
+        ops: Vec<StoredOp>,
+    ) -> BoxFut<'_, K2Result<()>> {
+        Box::pin(
+            async move { self.dht.write().await.inform_ops_stored(ops).await },
+        )
+    }
+}
 
 impl TxBaseHandler for K2Gossip {}
 impl TxModuleHandler for K2Gossip {
@@ -382,14 +392,17 @@ mod test {
     use kitsune2_core::{default_test_builder, Ed25519LocalAgent};
     use kitsune2_dht::UNIT_TIME;
     use kitsune2_test_utils::enable_tracing;
+    use kitsune2_test_utils::noop_bootstrap::NoopBootstrapFactory;
+    use kitsune2_test_utils::space::TEST_SPACE_ID;
     use std::time::Duration;
 
     #[derive(Debug, Clone)]
     struct GossipTestHarness {
-        gossip: K2Gossip,
+        _gossip: DynGossip,
         peer_store: DynPeerStore,
         op_store: DynOpStore,
         space: DynSpace,
+        peer_meta_store: Arc<K2PeerMetaStore>,
     }
 
     impl GossipTestHarness {
@@ -500,11 +513,30 @@ mod test {
     }
 
     impl TestGossipFactory {
-        pub async fn create(space: SpaceId) -> TestGossipFactory {
-            let mut builder = default_test_builder();
+        pub async fn create(
+            space: SpaceId,
+            bootstrap: bool,
+        ) -> TestGossipFactory {
+            let mut builder =
+                default_test_builder().with_default_config().unwrap();
             // Replace the core builder with a real gossip factory
             builder.gossip = K2GossipFactory::create();
-            let builder = Arc::new(builder.with_default_config().unwrap());
+
+            if !bootstrap {
+                builder.bootstrap = Arc::new(NoopBootstrapFactory);
+            }
+
+            builder
+                .config
+                .set_module_config(&K2GossipModConfig {
+                    k2_gossip: K2GossipConfig {
+                        initiate_interval_ms: 10,
+                        min_initiate_interval_ms: 10,
+                        ..Default::default()
+                    },
+                })
+                .unwrap();
+            let builder = Arc::new(builder);
 
             TestGossipFactory {
                 space_id: space,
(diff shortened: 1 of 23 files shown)


crates/kitsune2/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+keywords = ["holochain", "p2p", "dht", "networking"]
+categories = ["network-programming"]
+edition = "2021"
+
+[dependencies]
+bytes = { workspace = true }
+kitsune2_api = { workspace = true }
+kitsune2_core = { workspace = true }
+kitsune2_gossip = { workspace = true }
+kitsune2_transport_tx5 = { path = "../transport_tx5", version = "0.0.1-alpha" }
```
neonphog, line 18: put this in the workspace so any automated tools only have to update there

crates/kitsune2/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+/// Construct a default production builder for Kitsune2.
+///
+/// - `verifier` - The default verifier is [Ed25519Verifier].
+/// - `kitsune` - The default top-level kitsune module is [factories::CoreKitsuneFactory].
+/// - `space` - The default space module is [factories::CoreSpaceFactory].
+/// - `peer_store` - The default peer store is [factories::MemPeerStoreFactory].
+/// - `bootstrap` - The default bootstrap is [factories::CoreBootstrapFactory].
+/// - `fetch` - The default fetch module is [factories::CoreFetchFactory].
+/// - `transport` - The default transport is [Tx5TransportFactory].
+/// - `op_store` - The default op store is [MemOpStoreFactory].
```
neonphog, line 18:
```suggestion
/// - `op_store` - The default op store is [MemOpStoreFactory].
///                Note: you will likely want to implement your own op_store.
```

crates/transport_tx5/Cargo.toml
```diff
@@ -16,8 +16,8 @@ kitsune2_api = { workspace = true }
 serde = { workspace = true }
 tokio = { workspace = true, features = ["rt", "sync", "time"] }
 tracing = { workspace = true }
-tx5 = { version = "0.2.0-beta", default-features = false, features = [
-  "backend-libdatachannel",
+tx5 = { workspace = true, default-features = false, features = [
+    "backend-libdatachannel",
```
neonphog, line 20: haha, woops

crates/kitsune2/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+            url: std::sync::Mutex::new(
+                Url::from_str("ws://placeholder.url:1").unwrap(),
+            ),
+            notify_received_receiver: tokio::sync::Mutex::new(None),
+        });
+        let kitsune_1 = kitsune_builder_1
+            .kitsune
+            .create(kitsune_builder_1.clone(), kitsune_handler_1.clone())
+            .await
+            .unwrap();
```
neonphog, line 164: Instead of using the factory directly, use the [build function on the builder](https://github.com/holochain/kitsune2/blob/2428cf512c63e2b23efdb9566f0ff5d9fe34205d/crates/api/src/builder.rs#L117), so we get the extra validation and config markers.

crates/kitsune2/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+            url: std::sync::Mutex::new(
+                Url::from_str("ws://placeholder.url:2").unwrap(),
+            ),
+            notify_received_receiver: tokio::sync::Mutex::new(None),
+        });
+        let kitsune_2 = kitsune_builder_2
+            .kitsune
+            .create(kitsune_builder_2.clone(), kitsune_handler_2.clone())
+            .await
+            .unwrap();
```
neonphog, line 200: ditto build instead of factory

crates/kitsune2/src/lib.rs
```diff
@@ -0,0 +1,318 @@
...
+        let space_2 = kitsune_2.space(TEST_SPACE_ID).await.unwrap();
+
+        let agent_url_1 = kitsune_handler_1.url.lock().unwrap().clone();
+        let local_agent_1 = Arc::new(TestLocalAgent::default());
+        let agent_1 = AgentBuilder {
+            url: Some(Some(agent_url_1.clone())),
+            storage_arc: Some(DhtArc::FULL),
+            ..Default::default()
+        }
+        .build(local_agent_1.clone() as DynLocalAgent);
```
neonphog, line 210: As this is an integration test, should we rely on the bootstrap server module rather than doing this manually? (and the peer store injection below) ?

crates/gossip/src/gossip.rs
```diff
@@ -250,6 +251,18 @@ impl K2Gossip {
             target_peer_url
         );
 
+        // Before we send the initiate message, check whether the target has already
```
ThetaSinner, line 254: I was thinking that checking this as late as possible would minimize the race between two peers trying to initiate with each other. Was this causing a problem?

crates/gossip/src/gossip.rs
```diff
@@ -698,17 +642,8 @@ mod test {
             .await
             .unwrap();
 
-        harness_1
```
ThetaSinner, line 701: This should be necessary shouldn't it? I don't think anything else would do the update of the DHT model and if it's not updated then the DHT diff will just return "no diff" and not sync anything

crates/gossip/src/gossip.rs
```diff
@@ -717,42 +652,6 @@ mod test {
             .await
             .unwrap();
 
-        harness_2
```
ThetaSinner, line 720: Same here, I think this is a key part of the test. If it passes without it then something is wrong with the test

crates/gossip/src/gossip.rs
```diff
@@ -717,42 +652,6 @@ mod test {
...
-            .inform_ops_stored(vec![op_2.clone().into()])
-            .await
-            .unwrap();
-
-        // Inform the harnesses that we've already synced up to now. This will force the
-        // ops we just created to be treated as historical disc ops.
-        harness_1
-            .gossip
-            .peer_meta_store
-            .set_new_ops_bookmark(
```
ThetaSinner, line 734: Oh this is why the block above isn't needed. This test has just been reduced to a "what's new" check

ThetaSinner (changes requested):
Mostly looking good. I'm pretty sure the Gossip integration tests aren't covering what they were before. I think they probably need some more adjustment to work with the automatic initiation but I do want to be sure they're still testing what they were supposed to

crates/gossip/src/gossip.rs
```diff
@@ -63,11 +63,12 @@ impl GossipFactory for K2GossipFactory {
...
     ) -> kitsune2_api::BoxFut<'static, K2Result<DynGossip>> {
         Box::pin(async move {
-            let config: K2GossipConfig = builder.config.get_module_config()?;
+            let config =
+                builder.config.get_module_config::<K2GossipModConfig>()?;
 
             let gossip: DynGossip = Arc::new(
                 K2Gossip::create(
-                    config,
+                    config.k2_gossip,
```
jost-s, line 71: This caused a gossip build error.

crates/gossip/src/gossip.rs
jost-s: I forgot to comment here. It wasn't causing a problem, no. Just seemed negligible the operation to look up local agents.

crates/gossip/src/gossip.rs
```diff
@@ -614,25 +583,6 @@ mod test {
         let secret_agent_1 = harness_1.join_local_agent(DhtArc::FULL).await;
         let secret_agent_2 = harness_2.join_local_agent(DhtArc::FULL).await;
 
-        assert!(harness_1
```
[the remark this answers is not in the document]
jost-s: Yes, the checks no longer work, the peers are available immediately.

crates/gossip/src/gossip.rs
jost-s: It's not accessible and the test passed without it. Right, then something must be wrong.

ThetaSinner: Let me check this out and run the tests. It should be possible to see from the logs which sync method it's choosing. I'd like to be able to assert but what it did isn't recorded anywhere yet

crates/gossip/src/gossip.rs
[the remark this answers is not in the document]
ThetaSinner: That's not good, then the test is invalid

crates/gossip/src/gossip.rs
jost-s: Undone

crates/kitsune2/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+/// - `space` - The default space module is [factories::CoreSpaceFactory].
+/// - `peer_store` - The default peer store is [factories::MemPeerStoreFactory].
+/// - `bootstrap` - The default bootstrap is [factories::CoreBootstrapFactory].
+/// - `fetch` - The default fetch module is [factories::CoreFetchFactory].
+/// - `transport` - The default transport is [Tx5TransportFactory].
+/// - `op_store` - The default op store is [MemOpStoreFactory].
+///                Note: you will likely want to implement your own op_store.
+/// - `peer_meta_store` - The default peer meta store is [factories::MemPeerMetaStoreFactory].
+/// - `gossip` - The default gossip module is [K2GossipFactory].
+/// - `local_agent_store` - The default local agent store is [factories::CoreLocalAgentStoreFactory].
```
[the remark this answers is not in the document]
jost-s: You mean rename the type? There's no clear convention. The peer store is called MemPeerStore, but other in-memory structures are just called Core-.

crates/gossip/src/gossip.rs
jost-s: Okay, go ahead. I'm happy to do it as well. Sorry I castrated the tests. I was mostly focused on the transport issue.

ThetaSinner: Haha! No worries, they're not the best tests because it's not entirely clear what they do without an assertion at the end that they've used a particular mechanism to achieve something. I will look for ways to improve that but I've provided some suggested solutions now

jost-s:
I've removed the kitsune2 test from this PR. It wasn't stable. The rest is good to go in.

crates/kitsune2/src/lib.rs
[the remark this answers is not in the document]
neonphog: My intention, which I don't think I succeeded at, is that the things named Core are bits you'd likely never want to swap out, like CoreKitsune, CoreSpace, and CoreBootstrap. Arguably MemPeerStore could/should be Core also. Holochain will be swapping out this local agent store so it can use lair. But I'm totally up for a different convention if this isn't working : )

neonphog (approved):
awesome! :+1:

ThetaSinner (approved):

PR #88, opened 2025-01-29
test(core): fix test requests_are_dropped_when_max_back_off_expired in core_fetch

jost-s:
Before the test didn't make it through 200 iterations. Now 1000 iterations won't break it =)


ThetaSinner (approved):
🎉

PR #92, opened 2025-01-31
New ops respect common arc set

ThetaSinner:
This is another hopefully straight forward change. Requesting "what's new" should respect the common arc set during gossip in the same way that the DHT diff does. We wouldn't notice this right now with full arc but I think it's best to have this in place.

### DIFF
--- a/crates/core/src/factories/mem_op_store/test.rs
+++ b/crates/core/src/factories/mem_op_store/test.rs
@@ -4,7 +4,7 @@ use kitsune2_api::{DhtArc, DynOpStore, SpaceId, Timestamp};
 use std::sync::Arc;
 
 fn op_at_location(loc: u32) -> MemoryOp {
-    let mut out = vec![];
+    let mut out = Vec::with_capacity(32);
     out.extend_from_slice(&loc.to_le_bytes());
     out.resize(32, 0);
 
--- a/crates/core/src/factories/mem_op_store/test.rs
+++ b/crates/core/src/factories/mem_op_store/test.rs
@@ -1,6 +1,6 @@
 use crate::factories::mem_op_store::Kitsune2MemoryOpStore;
 use crate::factories::MemoryOp;
-use kitsune2_api::{DhtArc, DynOpStore, SpaceId, Timestamp};
+use kitsune2_api::{DhtArc, DynOpStore, Timestamp};
 use kitsune2_test_utils::space::TEST_SPACE_ID;
 use std::sync::Arc;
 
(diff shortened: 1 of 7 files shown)


crates/core/src/factories/mem_op_store/test.rs
```diff
@@ -1,8 +1,16 @@
 use crate::factories::mem_op_store::Kitsune2MemoryOpStore;
 use crate::factories::MemoryOp;
-use kitsune2_api::{DynOpStore, SpaceId, Timestamp};
+use kitsune2_api::{DhtArc, DynOpStore, SpaceId, Timestamp};
 use std::sync::Arc;
 
+fn op_at_location(loc: u32) -> MemoryOp {
+    let mut out = vec![];
+    out.extend_from_slice(&loc.to_le_bytes());
+    out.resize(32, 0);
```
neonphog, line 9:
```suggestion
    let mut out = Vec::with_capacity(32);
    out.extend_from_slice(&loc.to_le_bytes());
    out.resize(32, 0);
```

neonphog: not that it really matters for a test : )

neonphog (dismissed):

crates/core/src/factories/mem_op_store/test.rs
ThetaSinner: Can't hurt to be responsible :)

crates/dht/src/arc_set.rs
```diff
@@ -153,6 +153,51 @@ impl ArcSet {
...
+        }
+
+        let mut arcs = vec![];
+
+        let mut sectors = self.inner.iter().copied().collect::<Vec<_>>();
+        sectors.sort();
+
+        let mut start = 0;
+        while start < sectors.len() {
+            let mut end = start;
```
jost-s, line 180:
```suggestion
            // Determine consecutive sectors.
            let mut end = start;
```

jost-s (approved):

crates/core/src/factories/mem_op_store/test.rs
```diff
@@ -23,3 +31,120 @@ async fn process_and_retrieve_op() {
     let out: MemoryOp = serde_json::from_slice(&retrieved[0].op_data).unwrap();
     assert_eq!(op, out);
 }
+
+#[tokio::test]
+async fn retrieve_op_ids_bounded_empty_arc() {
+    let op_store: DynOpStore = Arc::new(Kitsune2MemoryOpStore::new(
+        SpaceId::from(bytes::Bytes::from_static(b"test")),
```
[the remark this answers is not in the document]
ThetaSinner: Applied locally to fix both tests

jost-s (approved):

PR #93, opened 2025-02-03
Proposed test modifications for gossip

ThetaSinner:


crates/api/src/space.rs
```diff
@@ -91,6 +94,19 @@ pub trait Space: 'static + Send + Sync + std::fmt::Debug {
...
+    ///
+    /// This should be called once the ops have been:
+    /// - Checked to be correctly structured according to the Kitsune2 host implementation.
+    /// - Persisted to the op store.
+    ///
+    /// Until this is called, gossip will not include these ops in its DHT model, and they will
+    /// therefore not be synced with other peers. They may be synced with other peers through the
+    /// "new ops" mechanism, depending on the op store implementation, but that only makes them
+    /// available briefly.
+    fn inform_ops_stored(&self, ops: Vec<StoredOp>)
```
ThetaSinner, line 108: This will need to be available for the host to call anyway, I just hadn't added it yet. So this solves the problem of the operation not being available on the K2Gossip when it's behind a DynGossip

crates/gossip/src/gossip.rs
```diff
@@ -502,11 +513,19 @@ mod test {
...
+        pub async fn create(
+            space: SpaceId,
+            bootstrap: bool,
+        ) -> TestGossipFactory {
             let mut builder =
                 default_test_builder().with_default_config().unwrap();
             // Replace the core builder with a real gossip factory
             builder.gossip = K2GossipFactory::create();
+
+            if !bootstrap {
```
ThetaSinner, line 525: Gives us a way to disable bootstrap so that gossip doesn't start until we manually populate one of the peer stores

crates/gossip/src/gossip.rs
```diff
@@ -552,11 +571,17 @@ mod test {
                 .await
                 .unwrap();
 
+            let peer_meta_store = K2PeerMetaStore::new(
```
ThetaSinner, line 574: This is a stateless wrapper, so it's safe to construct a "shadow" one here for now

crates/gossip/src/gossip.rs
```diff
@@ -251,6 +251,18 @@ impl K2Gossip {
...
+        if self
+            .accepted_round_states
+            .read()
+            .await
+            .contains_key(&target_peer_url)
+        {
+            tracing::info!("initiate_gossip: already accepted");
+            return Ok(false);
+        }
+
```
jost-s, line 265: I've reverted that =D

jost-s (approved):
Lovely

crates/gossip/src/gossip.rs
ThetaSinner: Haha, that's why it's showing up on this diff. I was confused by that because my rebase went through without a conflict. Cool, let me revert here!

PR #94, opened 2025-02-03
Limit data sent after dht differences

ThetaSinner:
Applies the same logic that we have for "new ops" to the op ids returned by the DHT diff process. I believe this is necessary regardless of whether it's a complete solution to the "new agent joining an existing network" problem.

By leaving this unlimited, we'd just return the entire data set in one batch during the first gossip round with new peers. That's too much load on the peer you happen to hit. As I'm writing, I'm wondering if we should enforce a maximum value for this that peers can request...

By limiting how much data we get in one go, we aren't doing anything too complicated like trying to decide which sectors we want data from or which time slices, when both are sparse and we'd get unpredictable results that way. Like this, we do our best to sync. Then we have to hope that between one gossip round and the next, the other peer will fetch enough to reduce the number of sectors/slices that produce a diff on the next round. If it's not that quick, then over time that is the effect.

As soon as we stop having a diff for a time slice or a sector, we'll be able to progress to new areas of the DHT in a natural way. There will be some duplicate requests of op ids but that's a reasonable thing to have happen I think. it's part of learning about the network.

This change really needs tests that exercise the 3 different paths
- Limit on "new ops"
- Limit on "disc diff" when part of the limit was consumed by "new ops"
- Limit on "ring diff" when part of the limit was consumed by "new ops"

But adding those tests requires modifying the testing in the same ways as #85

### DIFF
--- a/crates/api/src/op_store.rs
+++ b/crates/api/src/op_store.rs
@@ -77,12 +77,19 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
     ///
     /// This must be the timestamp of the op, not the time that we saw the op or chose to store it.
     /// The returned ops must be ordered by timestamp, ascending.
+    ///
+    /// # Returns
+    ///
+    /// - As many op ids as can be returned within the `limit_bytes` limit, within the arc and time
+    ///   bounds.
+    /// - The total size of the op data that is pointed to by the returned op ids.
     fn retrieve_op_hashes_in_time_slice(
         &self,
         arc: DhtArc,
         start: Timestamp,
         end: Timestamp,
-    ) -> BoxFuture<'_, K2Result<Vec<OpId>>>;
+        limit_bytes: Option<u32>,
+    ) -> BoxFuture<'_, K2Result<(Vec<OpId>, u32)>>;
 
     /// Retrieve a list of ops by their op ids.
     ///
@@ -112,13 +119,13 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
     ///
     /// - As many op ids as can be returned within the `limit_bytes` limit.
     /// - The total size of the op data that is pointed to by the returned op ids.
-    /// - A new timestamp to be used for the next query..
+    /// - A new timestamp to be used for the next query.
     fn retrieve_op_ids_bounded(
         &self,
         arc: DhtArc,
         start: Timestamp,
-        limit_bytes: usize,
-    ) -> BoxFuture<'_, K2Result<(Vec<OpId>, usize, Timestamp)>>;
+        limit_bytes: u32,
+    ) -> BoxFuture<'_, K2Result<(Vec<OpId>, u32, Timestamp)>>;
 
     /// Store the combined hash of a time slice.
     fn store_slice_hash(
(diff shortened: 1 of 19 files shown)


crates/api/src/op_store.rs
```diff
@@ -77,12 +77,19 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
...
+    /// - As many op ids as can be returned within the `limit_bytes` limit, within the arc and time
+    ///   bounds.
+    /// - The total size of the op data that is pointed to by the returned op ids.
     fn retrieve_op_hashes_in_time_slice(
         &self,
         arc: DhtArc,
         start: Timestamp,
         end: Timestamp,
-    ) -> BoxFuture<'_, K2Result<Vec<OpId>>>;
+        limit_bytes: Option<u32>,
```
ThetaSinner, line 91: Optional limit because this method is used in places where we don't want a limit.

crates/api/src/op_store.rs
```diff
@@ -112,13 +119,13 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
...
     /// - The total size of the op data that is pointed to by the returned op ids.
-    /// - A new timestamp to be used for the next query..
+    /// - A new timestamp to be used for the next query.
     fn retrieve_op_ids_bounded(
         &self,
         arc: DhtArc,
         start: Timestamp,
-        limit_bytes: usize,
-    ) -> BoxFuture<'_, K2Result<(Vec<OpId>, usize, Timestamp)>>;
+        limit_bytes: u32,
```
ThetaSinner, line 127: The limit config is u32 and I can't see us needing > u32::MAX bytes in one go :)

crates/gossip/proto/gen/kitsune2.gossip.rs
```diff
@@ -142,9 +142,11 @@ pub struct K2GossipInitiateMessage {
     /// Request ops that are new since the given timestamp.
     #[prost(int64, tag = "20")]
     pub new_since: i64,
-    /// The maximum number of bytes of new ops to respond with.
+    /// Gossip exchanges just op ids. This value is a hint to the other party about how many bytes of op data
+    /// those op ids should point to. The other party can't check this until they have fetched and checked the
+    /// op data, so you cannot rely on this being respected during gossip.
     #[prost(uint32, tag = "21")]
-    pub max_new_bytes: u32,
+    pub max_op_data_bytes: u32,
```
ThetaSinner, line 149: Rename because this limit is now shared between "new ops" and "dht diff ops"

crates/gossip/src/initiate.rs
```diff
@@ -576,7 +576,7 @@ mod tests {
             .await;
 
         let mut seen = HashSet::new();
-        for _ in 0..10 {
+        for _ in 0..100 {
```
ThetaSinner, line 579: Got a flake on this locally. It's waiting for a random number to select both agents so higher is less likely to flake

neonphog (approved):
woot, looks good.

>  As I'm writing, I'm wondering if we should enforce a maximum value for this that peers can request

Agree, just a local config, and we can `std::cmp::min()` the requested size and our max config.

PR #96, opened 2025-02-04
Op data limit tests

ThetaSinner:
Tests the previous PR -> https://github.com/holochain/kitsune2/pull/94

Some minor improvements needed in the mem peer store to support this change, I'll comment on those separately.

I think this raises an important problem though. I still want these tests in but I think I should change how I've applied limits. It shouldn't apply in the middle of a time slice because if a time slice grows larger than the limit, we'll never be able to sync it. So I think I should roll back the code that applies limits in the op store and does a soft application of the limit in the DHT code that calls the op store. That way we might return more op ids than the other side wanted but we won't get stuck on a time slice. It's an edge case really, for when there's more data around, but I think it's easier to fix now.

### DIFF
--- a/crates/core/src/factories/mem_op_store.rs
+++ b/crates/core/src/factories/mem_op_store.rs
@@ -189,9 +189,17 @@ impl OpStore for Kitsune2MemoryOpStore {
                 K2Error::other_src("Failed to deserialize op data, are you using `Kitsune2MemoryOp`s?", e)
             })?;
 
-            let op_ids =
-                ops_to_add.iter().map(|(op_id, _)| op_id.clone()).collect();
-            self.write().await.op_list.extend(ops_to_add);
+            let mut op_ids = Vec::with_capacity(ops_to_add.len());
+            let mut lock = self.write().await;
+            for (op_id, record) in ops_to_add {
+                if let std::collections::hash_map::Entry::Vacant(entry) =
+                    lock.op_list.entry(op_id.clone())
+                {
+                    entry.insert(record);
+                    op_ids.push(op_id);
+                }
+            }
+
             Ok(op_ids)
         })
     }
--- a/crates/core/src/factories/mem_op_store.rs
+++ b/crates/core/src/factories/mem_op_store.rs
@@ -215,16 +215,21 @@ impl OpStore for Kitsune2MemoryOpStore {
             let self_lock = self.read().await;
 
             let mut used_bytes = 0;
+            let mut candidate_ops = self_lock
+                .op_list
+                .iter()
+                .filter(|(_, op)| {
+                    let loc = op.op_id.loc();
+                    op.created_at >= start
+                        && op.created_at < end
+                        && arc.contains(loc)
+                })
+                .collect::<Vec<_>>();
+            candidate_ops.sort_by_key(|a| a.1.created_at);
+
             Ok((
-                self_lock
-                    .op_list
+                candidate_ops
                     .iter()
-                    .filter(|(_, op)| {
-                        let loc = op.op_id.loc();
-                        op.created_at >= start
-                            && op.created_at < end
-                            && arc.contains(loc)
-                    })
                     .take_while(|(_, op)| {
                         let data_len = op.op_data.len() as u32;
                         match limit_bytes {
@@ -243,7 +248,7 @@ impl OpStore for Kitsune2MemoryOpStore {
                             }
                         }
                     })
-                    .map(|(op_id, _)| op_id.clone())
+                    .map(|(op_id, _)| (*op_id).clone())
                     .collect(),
                 used_bytes,
             ))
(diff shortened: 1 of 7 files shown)


crates/core/src/factories/mem_op_store.rs
```diff
@@ -189,9 +189,17 @@ impl OpStore for Kitsune2MemoryOpStore {
                 K2Error::other_src("Failed to deserialize op data, are you using `Kitsune2MemoryOp`s?", e)
             })?;
 
-            let op_ids =
-                ops_to_add.iter().map(|(op_id, _)| op_id.clone()).collect();
-            self.write().await.op_list.extend(ops_to_add);
+            let mut op_ids = Vec::with_capacity(ops_to_add.len());
```
ThetaSinner, line 192: We shouldn't overwrite when inserting into the op store. I could swear I already fixed this but clearly not.

crates/core/src/factories/mem_op_store.rs
```diff
@@ -207,16 +215,21 @@ impl OpStore for Kitsune2MemoryOpStore {
...
+                .op_list
+                .iter()
+                .filter(|(_, op)| {
+                    let loc = op.op_id.loc();
+                    op.created_at >= start
+                        && op.created_at < end
+                        && arc.contains(loc)
+                })
+                .collect::<Vec<_>>();
+            candidate_ops.sort_by_key(|a| a.1.created_at);
```
ThetaSinner, line 228: We should return these in order. I actually think this matters to the DHT implementation if the hash map order isn't consistent so this would have been a significant bug if the real op store was implemented this way. For now, it was just about which ops we see in tests which appears a bit random if it's the hashmap order. Good to change this for both cases I believe.

crates/dht/src/dht/tests.rs
```diff
@@ -316,7 +316,7 @@ async fn ring_sync_with_matching_disc() {
     // Put recent data in the first ring of both DHTs
     dht1.inject_ops(vec![MemoryOp::new(
         dht1.dht.partition.full_slice_end_timestamp(),
-        vec![7; 4],
+        vec![17; 4],
```
ThetaSinner, line 319: Duplicate op id, the test was passing before the op store was more strict and it was valid that it was passing but broke with these changes. Better to just use unique op ids

crates/gossip/src/respond/initiate.rs
```diff
@@ -98,6 +98,12 @@ impl K2Gossip {
 
         // Update the peer's max op data bytes to reflect the amount of data we're sending ids for.
         // The remaining limit will be used for the DHT diff as required.
+        tracing::debug!(
```
ThetaSinner, line 101: Would quite like a human friendly display of byte sizes here but didn't think we wanted to introduce a library. Thoughts?

crates/gossip/src/gossip.rs
```diff
@@ -813,4 +823,253 @@ mod test {
...
+            .process_incoming_ops(
+                ops.clone().into_iter().map(Into::into).collect(),
+            )
+            .await
+            .unwrap();
+
+        let bookmark = Timestamp::now();
+
+        for i in 0u8..5 {
+            let op = MemoryOp::new(Timestamp::now(), vec![100 + i; 128]);
```
jost-s, line 931: might this be flaky? That's not the syntax, I guess, but
```suggestion
            let op = MemoryOp::new(bookmark + 1, vec![100 + i; 128]);
```

crates/gossip/src/gossip.rs
```diff
@@ -813,4 +823,253 @@ mod test {
...
+            .space
+            .inform_ops_stored(
+                ops.clone().into_iter().map(Into::into).collect(),
+            )
+            .await
+            .unwrap();
+
+        let harness_2 = factory.new_instance().await;
+
+        // Set a bookmark so that the first 5 ops are behind the bookmark and need a disc sync and
```
jost-s, line 952: That implies that new ops take preference over historical ops. Could add that as a comment for people who've not grown up with a gossip algorithm.

jost-s (approved):

crates/gossip/src/gossip.rs
ThetaSinner: I'm struggling greatly with holding what these tests are doing in my head and I think we need some metrics sooner rather than later that I can use to assert what gossip did. But that would change the "created_at" timestamp and not the "stored_at". It's because the stored_at logic that I'm doing this awkward "create some ops, get the bookmark timestamp, create some more ops" pattern

crates/gossip/src/gossip.rs
ThetaSinner: You mean you didn't have a pet gossip as a child!? :)

I did actually document this, not often I can say that :) Is this okay? https://github.com/holochain/kitsune2/blob/main/crates/gossip/src/config.rs#L11

crates/gossip/src/respond/initiate.rs
neonphog: [do we need a library?](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0ccd064d4ca9c73882de1150592049f2) : )

neonphog (approved):

crates/gossip/src/gossip.rs
jost-s: Okay, I see.

crates/gossip/src/gossip.rs
jost-s: Fair enough!

PR #97, opened 2025-02-04
test(kitsune2_gossip): Limit the value we'll accept as a max op bytes…

ThetaSinner:
We're just trusting the value the remote sends us at the moment and trying to send them as many op ids as their op data limit permits. To avoid peers setting this value high and using us to get all their op data at once, set a limit for this value and apply the lower of the two values.

For example, we could set a 10MB limit on a low-powered node even though people will be requesting 100Mb.

### DIFF
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -26,7 +26,7 @@ pub struct K2GossipConfig {
     /// fast they'll be able to discover ops from us because it'll require multiple rounds to
     /// discover all the op ids they'd need to request from us.
     ///
-    /// Default: 250MB
+    /// Default: 100MB
     pub max_request_gossip_op_bytes: u32,
 
     /// The interval in seconds between initiating gossip rounds.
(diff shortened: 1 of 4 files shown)


c12i (dismissed):
LGTM

crates/gossip/src/config.rs
```diff
@@ -71,6 +82,7 @@ impl Default for K2GossipConfig {
     fn default() -> Self {
         Self {
             max_gossip_op_bytes: 100 * 1024 * 1024,
+            max_request_gossip_op_bytes: 250 * 1024 * 1024,
```
jost-s, line 85: I would expect this to match the default of max bytes we can request from someone else, ergo 100. What do you think?

ThetaSinner: Yeah I went back and forth on this a bit. My thinking is that you can get quite a few ops for 100Mb. But Holochain enforces a 1Mb limit on entry size. So you could get < 100 ops back with a 100Mb limit. I thought it maybe made sense to leave a little extra room for nodes that want to serve a DHT faster to its peer to return more data. 

Equally, this is a default... so maybe if a node wants to do that then it should just configure this limit up itself. Yeah okay, setting this to the same thing makes sense

jost-s (approved):

PR #99, opened 2025-02-05
Update storage arcs

ThetaSinner:
The meat of this is the update algorithm. I've done my best to keep it logical and well-commented.

Basically, we're trying to start from the common arc set and skip any mismatched sectors. Then we see if merging the matching sectors into our storage arc yields a longer arc that covers the existing one. Starting from an empty arc is treated as a special case. It took me hours of thinking to put that together so I've done by best to comment and test clearly!

The integration is also interesting. If we get a ring diff then we can update our arc because we know the disc is synced and we can examine the ring diff to see what sectors are matching. That's pretty sound logic I think, but we won't necessarily get a ring diff because the "what's new" operation will keep some agents in sync without a DHT diff. We can also use the receipt of a minimal snapshot that exactly matches ours, as a signal to claim our entire target arc. However, this only happens when we initiated. I don't think it's as reasonable to blindly trust a "NoDiff" message as a signal to claim our entire target arc. I'd rather we always have some information we can work off. So that means that agents who are catching up with the network should claim their storage arc as they go and agents who actively initiate rounds will also discover that they can update their storage arc. Agents that are only passively accepting rounds (perhaps because it's a small network and the timing works out that way) might end up not updating their storage arc for a while until something interrupts the flow, like a new agent joining or somebody going offline.

### DIFF
--- a/crates/dht/src/dht/snapshot.rs
+++ b/crates/dht/src/dht/snapshot.rs
@@ -57,6 +57,8 @@ pub enum DhtSnapshot {
     /// hashes.
     RingSectorDetails {
         /// Similar to the `ring_top_hashes` except the sector hashes are not combined.
+        ///
+        /// Organized by ring index in the first HashMap, then sector index in the second HashMap.
         ring_sector_hashes: HashMap<u32, HashMap<u32, bytes::Bytes>>,
         /// The end timestamp of the most recent full time slice.
         disc_boundary: Timestamp,
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -5,6 +5,7 @@ use crate::protocol::{
     ArcSetMessage, GossipMessage, K2GossipInitiateMessage,
 };
 use crate::state::GossipRoundState;
+use crate::storage_arc::update_storage_arcs;
 use crate::timeout::spawn_timeout_task;
 use crate::update::spawn_dht_update_task;
 use crate::{K2GossipConfig, K2GossipModConfig, MOD_NAME};
@@ -286,6 +287,29 @@ impl K2Gossip {
         *initiated_lock = Some(round_state);
 
         Ok(true)
+    }
+
+    pub(crate) async fn update_storage_arcs(
+        &self,
+        next_action: &kitsune2_dht::DhtSnapshotNextAction,
+        their_snapshot: &kitsune2_dht::snapshot::DhtSnapshot,
+        common_arc_set: kitsune2_dht::ArcSet,
+    ) -> K2Result<()> {
+        if !matches!(
+            next_action,
+            kitsune2_dht::DhtSnapshotNextAction::CannotCompare
+        ) {
+            // As long as the comparison was successful, use this diff to update storage arcs.
+            if let Err(e) = update_storage_arcs(
+                their_snapshot,
+                self.local_agent_store.get_all().await?,
+                common_arc_set.clone(),
+            ) {
+                tracing::error!("Error updating storage arcs: {:?}", e);
+            }
+        }
+
+        Ok(())
     }
 
     /// Handle an incoming gossip message.
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -292,7 +292,7 @@ impl K2Gossip {
     pub(crate) async fn update_storage_arcs(
         &self,
         next_action: &kitsune2_dht::DhtSnapshotNextAction,
-        their_snapshot: &kitsune2_dht::snapshot::DhtSnapshot,
+        their_snapshot: &kitsune2_dht::DhtSnapshot,
         common_arc_set: kitsune2_dht::ArcSet,
     ) -> K2Result<()> {
         if !matches!(
--- a/crates/gossip/src/respond/accept.rs
+++ b/crates/gossip/src/respond/accept.rs
@@ -96,12 +96,13 @@ impl K2Gossip {
 
         match accept.snapshot {
             Some(their_snapshot) => {
+                let their_snapshot: DhtSnapshot = their_snapshot.into();
                 let (next_action, _) = self
                     .dht
                     .read()
                     .await
                     .handle_snapshot(
-                        their_snapshot.into(),
+                        their_snapshot.clone(),
                         None,
                         common_arc_set.clone(),
                         // Zero because this cannot return op ids
@@ -115,6 +116,13 @@ impl K2Gossip {
                         if let Some(state) = lock.as_mut() {
                             state.stage = RoundStage::NoDiff;
                         }
+
+                        self.update_storage_arcs(
+                            &next_action,
+                            &their_snapshot,
+                            common_arc_set.clone(),
+                        )
+                        .await?;
 
                         Ok(Some(GossipMessage::NoDiff(K2GossipNoDiffMessage {
                             session_id: accept.session_id,
--- a/crates/gossip/src/storage_arc.rs
+++ b/crates/gossip/src/storage_arc.rs
@@ -6,6 +6,11 @@ use std::collections::HashSet;
 
 /// Update the storage arcs of local agents based on a DhtSnapshot.
 ///
+/// The snapshot may either be a [DhtSnapshot::RingSectorDetails] or a [DhtSnapshot::Minimal].
+/// A minimal snapshot may only be provided when the snapshot has already been checked to be
+/// identical to our own minimal snapshot. In that case we know that the entire common arc set
+/// was in sync.
+///
 /// Note that the DHT model prioritises syncing the disc. So if we've reached a ring diff then we
 /// can assume that the disc was synced at the point this diff was produced. This function requires
 /// the input [DhtSnapshot] to be a [DhtSnapshot::RingSectorDetails] variant.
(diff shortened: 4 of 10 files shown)


crates/dht/src/dht/snapshot.rs
```diff
@@ -250,8 +250,6 @@ impl DhtSnapshot {
                 },
             ) => {
                 if our_disc_boundary != other_disc_boundary {
-                    // TODO Don't expect the boundary to move during a comparison so treat this as
```
ThetaSinner, line 253: No longer relevant, we're permitting boundary updates during gossip rounds so this can happen and returning `CannotCompare` is the right action to take

crates/gossip/src/gossip.rs
```diff
@@ -302,6 +303,29 @@ impl K2Gossip {
         Ok(true)
     }
 
+    pub(crate) async fn update_storage_arcs(
```
ThetaSinner, line 306: Extracting common code that's used in a couple of places

crates/gossip/src/respond/accept.rs
```diff
@@ -90,12 +90,13 @@ impl K2Gossip {
...
         match accept.snapshot {
             Some(their_snapshot) => {
+                let their_snapshot: DhtSnapshot = their_snapshot.into();
                 let (next_action, _) = self
                     .dht
                     .read()
                     .await
                     .handle_snapshot(
-                        their_snapshot.into(),
+                        their_snapshot.clone(),
```
ThetaSinner, line 99: Cloning, which I'd prefer not to do with this type but we need to compare the snapshot before updating arcs and we can't pass a reference across the box future boundary here

crates/gossip/src/storage_arc.rs
```diff
@@ -0,0 +1,290 @@
...
+///
+/// Note that the DHT model prioritises syncing the disc. So if we've reached a ring diff then we
+/// can assume that the disc was synced at the point this diff was produced. This function requires
+/// the input [DhtSnapshot] to be a [DhtSnapshot::RingSectorDetails] variant.
+///
+/// The function looks for sectors within the common arc set that did not show up as a mismatch
+/// in the snapshot. Such sectors are considered synced. If any synced sectors can be added to the
+/// start or end of the current storage arc for a local agent, it will do so. Any synced sectors
+/// that cannot be added to the storage arc to produce a larger continuous arc are ignored.
+pub(crate) fn update_storage_arcs(
```
neonphog, line 22: > If any synced sectors can be added to the ***start or*** end of the current storage arc for a local agent, it will do so.

In legacy, storage arcs must start in the block that contains the agent's loc. Are we considering changing that here?

neonphog (changes requested):
All the code here is awesome, but:

- if storage arcs can now be something other than Empty and Full, it seems we are implicitly introducing sharding. Are we ready for that?
- parts of our theories of resilience depend on agents holding arcs relative to their agent location. (a) we never validate that in legacy, so it's not like this is strictly worse in the case of bad actors, and (b) since it is trivial to farm agent ids, maybe our theories are wrong anyways -- but we should explicitly choose before changing that : )

crates/gossip/src/storage_arc.rs
```diff
@@ -0,0 +1,290 @@
...
+            // When our current storage arc is empty, we're free to pick a new one that is
+            // contained within the target arc. Pick the largest one.
+            if let Some(new_arc) =
+                new_arcs.into_iter().max_by_key(|arc| arc.len())
+            {
+                local_agent.set_cur_storage_arc(new_arc);
+            }
+        } else {
+            // At this point, we have added any synced sectors to our current storage arc and ignored
+            // and synced sectors that aren't in our target arc. We can find the largest overlapping
```
jost-s, line 83: not ignored and synced, I believe?
```suggestion
            // sectors that aren't in our target arc. We can find the largest overlapping
```

jost-s (approved):
Just when I read the last test, I understood how this works. What do you think about a test where the target arc is smaller than the storage arc? Not that that should happen with our kitsune...

ThetaSinner:
I think it's easier to answer all questions in one big comment here.

> In legacy, storage arcs must start in the block that contains the agent's loc. Are we considering changing that here?

Not necessarily no. This is just about how I've gone about initializing from empty. Say I have a storage arc starting at my agent key which is in sector 30 and running to sector 60. In the first sync I do, I discover that I'm in sync for 30-32, 55-60 and 35-45. The algorithm chooses the longest arc to initialize from so we'll end up with 35-45 and as we get in sync for other sectors we'll expand towards 30-60. That does mean that while we're syncing, other nodes wouldn't be able to validate that we're starting from our agent key's sector. I could change this so that we require the arc to go from empty to an arc that starts at our agent key and then we'd only expand right.

> if storage arcs can now be something other than Empty and Full, it seems we are implicitly introducing sharding. Are we ready for that?

The target arc would only be full or empty. This is about incrementally growing our storage arc as we get in sync with the network. So yes we're declaring the storage arc as something other than empty or full but as Jost [said](https://github.com/holochain/kitsune2/pull/99#issuecomment-2639646841), it's the dynamic updating of the target arc that introduces sharding. That or allowing the arc to be pinned to some specific size.

All we're really doing here is trying to avoid sending `get` requests to nodes that aren't serving the target data yet.

> parts of our theories of resilience depend on agents holding arcs relative to their agent location

I think that's still valid. When it comes to sharding and setting a target arc to something other than full or empty, we can still set the target arc to start from the agent key. The way I've written this code, it won't be possible for other nodes to check that the agent is targeting the right arc but this code will still end up declaring a storage arc that starts at the agent key if the target arc does.

> If they couldn't, how could an agent have a full storage arc, since they start out with nothing in their op store?

There was just a single arc in K1, so your target and storage arc were one. You effectively declared your target arc up front and might not be able to serve data from it initially. One of many reasons for the endless timeouts we see in tests with 3+ conductors.

ThetaSinner:
> What do you think about a test where the target arc is smaller than the storage arc? Not that that should happen with our kitsune...

Yes, that's worth testing. It shouldn't do anything I believe because that would be a mistake somewhere else. The test confirms that it doesn't and leaves a comment about why it would be an unexpected input.

crates/gossip/src/storage_arc.rs
ThetaSinner: Correct, that's redundant - I updated this locally

jost-s (dismissed):
Sweet

neonphog:
> The target arc would only be full or empty. This is about incrementally growing our storage arc as we get in sync with the network. So yes we're declaring the storage arc as something other than empty or full but as Jost https://github.com/holochain/kitsune2/pull/99#issuecomment-2639646841, it's the dynamic updating of the target arc that introduces sharding.

Maybe this is becoming a semantic argument, but for me sharding is when you have to go to specific peers for specific data. It's easy enough to ignore zero arcs, but if the data is sharded (storage arcs are between empty and full) we have to know who to go to (as you point out in your comment about gets), so maybe we can't defer the peer discovery logic for finding who to "get" from?

jost-s:
That's a given anyway, that not all peers hold all data, even with a targeted full arc. The get call will need to look through the known peers' arcs to see where a piece of data can be found, no?

jost-s:
In theory, yes, but in reality a just-joined peer won't hold all data. In this theory only the empty arc peers would need to do a get to a full arc peer.
What is your idea, should the storage arc always be set to full or empty, regardless of the actual store's state? The target arc would not be needed in that case.

neonphog:
This is what I thought we were doing: In reality everybody joins the network with an empty arc. We still need the target arc, because we need to know if we're trying for full, or are just a mobile leecher that will always stay at empty. Then, once we've fetched all the data for full, we'd update our storage arc to full.

I'm not saying we can't do some interim step, I'm saying we hadn't been going to introduce sharding yet, and this feels to me like we're introducing sharding, so I want to make sure we've thought through all the consequences of that.

jost-s:
Understood. Does this introduce anything else, other than that going from empty to full is not a single step but multiple steps?

neonphog:
I'm not sure what you are asking. I've said everything I have to say.

ThetaSinner:
I see what you're thinking now. Yes okay so we might be publishing a partial arc while we're working towards being full arc. My understanding is that we were going to try to be "sharding ready" in K2 and we'll come back and add proper support for sharding soon and add it into Holochain soon after.

I will happily put this logic behind a feature flag and just declare a FULL arc once we have a gossip round without a diff. Alternatively, we could just do peer selection as "empty" or "not empty" rather than by overlap. We don't actually have to ensure that get requests succeed until we're intentionally doing sharding. That seems like a reasonable middle ground to be in. We might make a request early to a peer who isn't in sync yet and miss.. but that was already happening in K1 where peers were declaring arcs they didn't have yet. Preference between the two options? Or another option I haven't thought of?

jost-s:
> I'm not sure what you are asking. I've said everything I have to say.

I'm asking the question whether this progressive updating of arcs from empty to full introduces anything else than just that? You wanted to be clear on consequences of making this change, so I'm asking what are the consequences other than more incremental steps from empty to full?

@ThetaSinner That's what I was wondering, how does this work in legacy kitsune. It wasn't, I see. Until it's not clear what the consequences of growing storage arcs are, I think the feature flag is safer.

neonphog:
> > parts of our theories of resilience depend on agents holding arcs relative to their agent location
>
> I think that's still valid. When it comes to sharding and setting a target arc to something other than full or empty, we can still set the target arc to start from the agent key.

Peers make get requests based on the storage arc, not the target arc. If you can specify your storage arc as starting from anywhere, you can more easily claim authority over a dht item that you want to censor or alter. Again... not sure legacy actually prevents this, but that was the theory.

neonphog:
> so I'm asking what are the consequences other than more incremental steps from empty to full?

As far as I can tell, just what I've been saying, that now when we're making gets or anything else like trying to find an agent authority to talk to, we can't just ignore the empties and assume everyone else is valid, because they will be claiming storage arcs that may not cover the location we care about.

neonphog:
> Preference between the two options? Or another option I haven't thought of?

I'd lean toward keeping the current logic, but suppress updating the storage arc when it is anything other than Empty or Full. Then we can defer addressing the starting sector problem and for the time being avoid having any published partial storage arcs.

jost-s:
> As far as I can tell, just what I've been saying, that now when we're making gets or anything else like trying to find an agent authority to talk to, we can't just ignore the empties and assume everyone else is valid, because they will be claiming storage arcs that may not cover the location we care about.

Alright, got it. We're choosing between
1. If arcs are empty until they're full, an agent is "withholding data" that they could be serving while the arc is growing
2. If arcs are growing incrementally, gets need to select based on arc overlap who to get from

In both cases there can be scenarios where there's no peer to get from. I like David's suggestion to defer dealing with the aspects of the second case and unforeseens.

ThetaSinner:
I can remove the code and tests, they'll probably have to change again anyway. I'd prefer to keep them behind a feature flag because I think it'd give me a head start updating this later but I don't have to.

Does this look better, just wait for a full arc to come out of the expand check and then declare a full arc -> https://github.com/holochain/kitsune2/pull/99/commits/e8e5a4ce4fb016ec3b237503b1755855e1f35881

ThetaSinner:
For sure, I just needed to load into my head what the problems were with the way I'd done this :)

PR #100, opened 2025-02-05
test(kitsune): add integration test

jost-s:


crates/kitsune2/src/lib.rs
```diff
@@ -37,3 +36,215 @@ pub fn default_builder() -> Builder {
...
+            Box::pin(async {
+                let space_handler: DynSpaceHandler =
+                    Arc::new(HolochainSpaceHandler);
+                Ok(space_handler)
+            })
+        }
+    }
+
+    #[derive(Debug)]
+    struct HolochainSpaceHandler;
```
ThetaSinner, line 95: Would prefer to not have `Holochain*` names in here, hopefully we've done a reasonable job of decoupling here

crates/kitsune2/src/lib.rs
```diff
@@ -37,3 +36,215 @@ pub fn default_builder() -> Builder {
...
+        let (ops_1, op_ids_1) = create_op_list(100);
+        space_1
+            .op_store()
+            .process_incoming_ops(ops_1.clone())
+            .await
+            .unwrap();
+        let (ops_2, op_ids_2) = create_op_list(100);
+        space_2
+            .op_store()
+            .process_incoming_ops(ops_2.clone())
```
ThetaSinner, line 175: Process incoming ops will make the ops available for "what's new" but there will be no DHT diff happening unless gossip is informed about these ops through the space with "inform_ops_stored". That will work fine for this test because all the ops are new. Just wanted to mention that.

crates/kitsune2/src/lib.rs
```diff
@@ -37,3 +36,215 @@ pub fn default_builder() -> Builder {
...
+            .config
+            .set_module_config(&CoreBootstrapModConfig {
+                core_bootstrap: CoreBootstrapConfig {
+                    server_url: bootstrap_server_url.to_owned(),
+                    backoff_max_ms: 1000,
+                    ..Default::default()
+                },
+            })
+            .unwrap();
+        kitsune_builder.op_store = MemOpStoreFactory::create();
```
neonphog, line 114: Don't replace a factory after you've already set the default config... you end up with a config for the wrong factory.

```rust
let mut kitsune_builder = Builder {
    op_store: MemOpStoreFactory::create(),
    ..default_builder()
}.with_default_config().unwrap();
```

neonphog: Also, the MemOpStoreFactory *is* the default, so you don't need to do this at all.

neonphog (changes requested):
```
running 1 test
test test::two_node_gossip ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.01s
```

I wish cargo test would give times for the individual tests. Luckily, since there's only one in there, we can see that it is 10s! Not entirely unreasonable.

Couple cleanup suggestions.

crates/kitsune2/src/lib.rs
jost-s: Yeah, I had that here but noticed it didn't make a diff(erence).

neonphog (approved):
Woot, looks great!

jost-s:
@neonphog Sorry, just noticed the note about implementing your own peer meta store was deleted in this PR and I reverted that. Please re-approve.

jost-s:
Well damn, might we be seeing the issue on slow windows with tx5 being overloaded? https://github.com/holochain/kitsune2/actions/runs/13334116529/job/37245679121?pr=100#step:6:340

neonphog (approved):

neonphog:
> Well damn, might we be seeing the issue on slow windows with tx5 being overloaded?

The warnings that would indicate that are not showing up. Some other problem?

jost-s:
One agent initiates gossip, but the other end doesn't seem to receive the response.

The other agent initiates gossip, but immediately closes the connection https://github.com/holochain/kitsune2/actions/runs/13334116529/job/37245679121?pr=100#step:6:288

jost-s:
I've increased the delay for windows before adding ops and therefore starting to gossip. It's plausible that this is the issue you have mentioned before @neonphog, where the connection is not fully established yet and therefore no requests can even be sent. That's why there are no warnings from fetch or gossip, because the connections are closed straight away.

@ThetaSinner agreed to merge the test like this and come back to it if it turns out to be flaky or if we start seeing issues on windows with the integrated kitsune2.

neonphog (approved):

PR #101, opened 2025-02-05
Limit op data at sector granularity

ThetaSinner:
This one is slightly subtle. I initially applied the limit rather harshly at an "op boundary". If this falls within a really large sector/slice then it's possible that the limit isn't set high enough to ever retrieve that data. I think it's better to be less strict with the limit to ensure we always make progress when gossiping a diff.

The changes on this PR are all a fallout from that change.

### DIFF
--- a/crates/api/src/op_store.rs
+++ b/crates/api/src/op_store.rs
@@ -88,7 +88,6 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
         arc: DhtArc,
         start: Timestamp,
         end: Timestamp,
-        limit_bytes: Option<u32>,
     ) -> BoxFuture<'_, K2Result<(Vec<OpId>, u32)>>;
 
     /// Retrieve a list of ops by their op ids.
--- a/crates/dht/src/dht.rs
+++ b/crates/dht/src/dht.rs
@@ -98,7 +98,7 @@ pub trait DhtApi: 'static + Send + Sync + std::fmt::Debug {
         their_snapshot: DhtSnapshot,
         our_previous_snapshot: Option<DhtSnapshot>,
         arc_set: ArcSet,
-        max_op_data_bytes: u32,
+        max_op_data_bytes: i32,
     ) -> BoxFut<'_, K2Result<(DhtSnapshotNextAction, u32)>>;
 }
 
@@ -222,7 +222,7 @@ impl DhtApi for Dht {
         their_snapshot: DhtSnapshot,
         our_previous_snapshot: Option<DhtSnapshot>,
         arc_set: ArcSet,
-        max_op_data_bytes: u32,
+        max_op_data_bytes: i32,
     ) -> BoxFut<'_, K2Result<(DhtSnapshotNextAction, u32)>> {
         Box::pin(async move {
             if arc_set.covered_sector_count() == 0 {
@@ -346,9 +346,16 @@ impl DhtApi for Dht {
                 SnapshotDiff::DiscSectorSliceMismatches(
                     mismatched_slice_indices,
                 ) => {
+                    // Need to fetch in order by sector index
+                    let mut mismatched_slice_indices = mismatched_slice_indices
+                        .into_iter()
+                        .collect::<Vec<_>>();
+                    mismatched_slice_indices
+                        .sort_by_key(|(sector_index, _)| *sector_index);
+
                     let mut out = Vec::new();
                     let mut used_bytes = 0;
-                    for (sector_index, missing_slices) in
+                    for (sector_index, mut missing_slices) in
                         mismatched_slice_indices
                     {
                         let Ok(arc) = self
@@ -361,6 +368,9 @@ impl DhtApi for Dht {
                             );
                             continue;
                         };
+
+                        // Need to fetch in order by slice index
+                        missing_slices.sort();
 
                         for missing_slice in missing_slices {
                             let Ok((start, end)) = self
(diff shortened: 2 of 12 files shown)


crates/dht/src/dht.rs
```diff
@@ -98,8 +98,8 @@ pub trait DhtApi: 'static + Send + Sync + std::fmt::Debug {
         their_snapshot: DhtSnapshot,
         our_previous_snapshot: Option<DhtSnapshot>,
         arc_set: ArcSet,
-        max_op_data_bytes: u32,
-    ) -> BoxFut<'_, K2Result<(DhtSnapshotNextAction, u32)>>;
+        max_op_data_bytes: i32,
```
ThetaSinner, line 101: The limit is now soft, so we might have a negative number in the limit if we've already exceeded it.

ThetaSinner: I think we still want to return the next diff even if we're out of budget to return ops... Not certain about that but I think it still makes sense to proceed here.

crates/dht/src/dht.rs
```diff
@@ -346,9 +346,16 @@ impl DhtApi for Dht {
                 SnapshotDiff::DiscSectorSliceMismatches(
                     mismatched_slice_indices,
                 ) => {
+                    // Need to fetch in order by sector index
```
ThetaSinner, line 349: This wasn't exactly a bug but it was a little odd. We were letting the hashmap decide whatever to return data in. The tests I'd written were flaky like this once I split the op data over multiple slices.

ThetaSinner: So when retrieving all the data in one go, no problem. In terms of syncing everything from a large DHT and needing all of it. Sure we'll get it eventually. It makes a lot more sense to me for this to be predictable.

crates/dht/src/dht.rs
```diff
@@ -379,15 +389,14 @@ impl DhtApi for Dht {
...
-                                    start,
-                                    end,
-                                    Some(max_op_data_bytes - used_bytes),
+                                    arc, start, end,
                                 )
                                 .await?;
-                            out.extend(op_ids);
 
-                            used_bytes += ub;
+                            if (used_bytes + ub) as i32 <= max_op_data_bytes {
```
ThetaSinner, line 396: The new limiting logic

crates/api/src/op_store.rs
```diff
@@ -88,7 +88,6 @@ pub trait OpStore: 'static + Send + Sync + std::fmt::Debug {
         arc: DhtArc,
         start: Timestamp,
         end: Timestamp,
-        limit_bytes: Option<u32>,
```
ThetaSinner, line 91: Oh and bonus, we wanted these two out of the API for the op store, so 50% done :)

neonphog (approved):
Part of me wants to design some kind of paging protocol for sector syncing, but no, this is exactly what we want for this stage. Nice!

ThetaSinner:
> Part of me wants to design some kind of paging protocol for sector syncing

Likewise :D But it's a tricky thing to do when you have to work across sector boundaries and you haven't necessarily "got" the page because you're after the data not the ids and so it makes sense to request the same page from multiple agents until you've fetched enough data to be in sync with them... I think what we're ending up doing is actually reasonably close to what a paging version would look like. It's just not so clear somehow because it's hidden behind a diff check.

crates/dht/src/dht.rs
```diff
@@ -379,15 +389,14 @@ impl DhtApi for Dht {
...
+                                out.extend(op_ids);
+                                used_bytes += ub;
+                            }
                         }
                     }
```
jost-s, line 401: This can break early if max ops have been reached. There wouldn't be traces then.
```suggestion
                    } else {
                        break 'outer;
                    }
```

crates/gossip/src/state.rs
```diff
@@ -24,7 +24,10 @@ pub(crate) struct GossipRoundState {
     ///
     /// Note that it's actually [kitsune2_api::OpId]s that are exchanged. So this is a hint in
     /// terms of op data about how many op ids we should send back to the other peer.
-    pub peer_max_op_data_bytes: u32,
+    ///
+    /// The value is an `i32` because it's a soft limit and in order to send complete slices,
+    /// gossip may exceed the limit. That is recorded here as a negative number.
```
jost-s, line 29: What would a value like -100 indicate? The number of bytes can be the next complete slice greater than 100 bytes?

crates/gossip/src/state.rs
```diff
@@ -66,7 +69,7 @@ impl GossipRoundState {
             session_with_peer,
             started_at: tokio::time::Instant::now(),
             session_id,
-            peer_max_op_data_bytes,
+            peer_max_op_data_bytes: peer_max_op_data_bytes as i32,
```
jost-s, line 72: Doesn't this also have to change to an i32?

jost-s (commented):
Looks good. I have a few questions.

jost-s:
@ThetaSinner While I was reviewing this PR, I didn't notice it got merged. Please take a look at my questions anyway.

crates/dht/src/dht.rs
ThetaSinner: I was thinking of letting the loop proceed to look for a slice that's smaller and might fit into the limit. But maybe that's a lot of wasteful database queries. I think it'd be reasonable to break if the limit is reached or exceeded but probably not if the current slice is too big. I'll take a look at this

crates/gossip/src/state.rs
ThetaSinner: I don't think the config should no, that's just a limit value that shouldn't be set to a negative number. It's being assigned to a value that's being used to track actual usage against that limit. I think this is the right place to cast

PR #102, opened 2025-02-06
Do less work when the byte limit is exhausted after finding a DHT diff

ThetaSinner:
Follow up for https://github.com/holochain/kitsune2/pull/101


jost-s (approved):
Cool!

PR #103, opened 2025-02-06
docs: Remove complete or no longer relevant TODOs

ThetaSinner:


crates/core/src/factories/core_space.rs
```diff
@@ -308,7 +308,6 @@ impl Space for CoreSpace {
             // update our local map
             self.local_agent_store.add(local_agent.clone()).await?;
 
-            // TODO - inform gossip module of new join
```
ThetaSinner, line 311: Gossip doesn't need to know about newly joined agents. It will need to know about changing storage arcs at some point so receiving updated agent info from other peers is something we'll want to notify gossip about. There's another TODO for that.

But for our local agents, gossip always looks this up fresh when it starts, and I don't think it needs to know more than that. If all the agents leave, gossip just stops and waits to be shut down

neonphog (approved):
cool

PR #106, opened 2025-02-06
chore: Prepare manifests for publishing

ThetaSinner:


neonphog (approved):
woot. This seems so much nicer!

crates/api/Cargo.toml
jost-s, line 1: Fantastic! I didn't know this was possible.

PR #108, opened 2025-02-07
feat(kitsune2_gossip): Clean up bookmark handling

ThetaSinner:


crates/gossip/src/gossip.rs
```diff
@@ -244,7 +244,10 @@ impl K2Gossip {
             .peer_meta_store
             .new_ops_bookmark(target_peer_url.clone())
             .await?
-            .unwrap_or(UNIX_TIMESTAMP);
+            .unwrap_or_else(|| {
```
ThetaSinner, line 247: I was setting a 0 time for this when I was thinking we'd be incrementally syncing on first join but that's not right. We want to just request what's new that the DHT diff won't catch. We also reset this value if it's been a while since we gossiped with a peer so we really don't want to jump back to timestamp 0.

crates/gossip/src/respond.rs
```diff
@@ -155,12 +155,6 @@ impl K2Gossip {
         self.fetch
             .request_ops(decode_ids(accept_response.new_ops), from_peer.clone())
             .await?;
-        self.peer_meta_store
```
ThetaSinner, line 158: This was moved to `update_new_ops_bookmark` but didn't get removed here. Not too serious if the received value is accepted but there is a check in the other function. Certainly no need to do it twice

crates/gossip/src/respond.rs
```diff
@@ -266,6 +260,8 @@ impl K2Gossip {
             .map(|previous_bookmark| previous_bookmark <= updated_bookmark)
             .unwrap_or(true)
         {
+            // TODO Ideally we'd reset this if their arc set changes, to avoid missing ops in new
```
ThetaSinner, line 263: This bit is still relevant, moved to where the bookmark update happens

crates/gossip/src/respond/initiate.rs
```diff
@@ -87,7 +87,10 @@ impl K2Gossip {
             .peer_meta_store
             .new_ops_bookmark(from_peer.clone())
             .await?
-            .unwrap_or(UNIX_TIMESTAMP);
+            .unwrap_or_else(|| {
+                (Timestamp::now() - 2 * UNIT_TIME)
+                    .unwrap_or_else(|_| Timestamp::now())
+            });
```
neonphog, line 93: ditto the code comment here.

neonphog (approved):
Looks good :+1:

crates/gossip/src/gossip.rs
ThetaSinner: Yes, fair. Extracted a common function and an explanation comment

neonphog (approved):

PR #109, opened 2025-02-07
chore: publish crates

ThetaSinner:


Cargo.toml
```diff
@@ -96,7 +96,7 @@ prost-build = "0.13.3"
 # above this section.
 # --- dev-dependencies ---
 kitsune2_bootstrap_srv = { path = "crates/bootstrap_srv" }
-kitsune2_core = { path = "crates/core" }
+kitsune2_core = { version = "0.0.1-alpha", path = "crates/core" }
```
ThetaSinner, line 99: Publish complains if you don't set a version for dependencies, fair enough!

Makefile
```diff
@@ -26,3 +26,13 @@ test:
 
 proto:
 	cd crates/tool_proto_build && cargo run
+
+publish_all:
```
ThetaSinner, line 30: Add basic helper

Cargo.toml
jost-s: Can you use the workspace version here?

jost-s (changes requested):

Cargo.toml
ThetaSinner: I can't find anything in the docs about this and the `workspace.version` syntax is rejected by Cargo

neonphog (dismissed):

Cargo.toml
jost-s: Okay

jost-s:
Actually can you remove the holo keyword from the Cargo.toml? If we really want to decouple this from Holochain, we'd remove it from those keywords too.

ThetaSinner:
Removed Holo and added Kitsune to help group these crates with a keyword. I think as this is still under the Holochain org, leaving that in as a keyword is good for discovery

neonphog (approved):

PR #113, opened 2025-02-07
chore: Bump versions for publish

ThetaSinner:
Release today's changes for use in Holochain

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.0.1-alpha"
+version = "0.0.1-alpha.1"
 authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
 homepage = "https://github.com/holochain/kitsune2"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.0.1-alpha"
+version = "0.0.1-alpha.1"
```
neonphog, line 18: Unfortunately, cargo will auto-upgrade these like this. In the repos I manage, I've gone with just bumping the patch version `0.0.2-alpha`. But I'm pretty sure you can also delete the dot if you'd rather: `0.0.1-alpha1`

ThetaSinner: Ahh okay, I forgot about that. Unfortunately these are already published. But with this being the first versions, nothing is released on Holochain with it and no harm done. I'll make sure to bump the patch number next time

neonphog (approved):

jost-s (approved):

PR #115, opened 2025-02-10
docs(kitsune2): add links to module crates

jost-s:
Moving a couple of module level docs to crate level, because they weren't accessible due to the flat API structure. Added some static links to diagrams to the top-level crate.


neonphog:
> Moving a couple of module level docs to crate level, because they weren't accessible due to the flat API structure.

Hmm, I'm not sure I like that. I didn't think of that when I agreed to the flat structure...

Would it work to have an empty doc-only module like:

`agent.rs`
```
/// My top-level agent docs here...
#[cfg(doc)]
pub mod agent_docs {}
```

So the docs could remain in the modules they are related to?

jost-s:
Cool, yeah, I'll try that. I don't like it either and just wanted a quick now to make them available.

neonphog:
Alternately, if the module has a clear main struct associated with it, we could move the docs onto that struct?

zippy (dismissed):

jost-s:
How about this now? I'm quite happy like this, the doc modules are consistent in api and core now.

ThetaSinner:
Or maybe I should rethink the change I made. I'm fine with `pub mod` in the API as long as we then either export through a prelude or require users to import from the module and don't also `pub use` the same thing. It was just that I was trying to deal with

jost-s:
Documenting the private items is confusing, I think. `pub mod` is a good option, but also have a look at a local doc build of this branch.

ThetaSinner:
The docs do look tidy like that, as long as we're happy splitting module docs out into the docs module

jost-s:
I'm fine with returning to pub mods, but we can also decide that later when we come back to the docs.

crates/core/src/lib.rs
```diff
@@ -169,6 +169,9 @@ pub fn default_test_builder() -> Builder {
 
 pub mod factories;
 
+// #[cfg(any(doc, docsrs))]
+pub mod doc;
```
neonphog, line 173: why is the attr commented out?

jost-s: Oops, forgot to put that back in. It was commented so that links would be resolved in the IDE.

neonphog (approved):
woot : )

PR #116, opened 2025-02-11
Build/add nix package for bootstrap srv

jost-s:
- Add a nix flake with a package for building the bootstrap server.
- Add a job to ci to test the nix build.

resolves https://github.com/holochain/wind-tunnel/issues/144

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -48,12 +48,34 @@ jobs:
       - name: Test
         run: make test
 
+  nix-build:
+    needs: static
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+      fail-fast: false
+    
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v30
+
+      - uses: cachix/cachix-action@v15
+        with:
+          name: holochain-wind-tunnel
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - run: nix build --verbose -L .#bootstrap-srv
+
   ci_pass:
     if: ${{ always() }}
     runs-on: "ubuntu-latest"
     needs:
       - static
       - test
+      - nix-build
     steps:
       - name: check status
         uses: re-actors/alls-green@release/v1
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -48,12 +48,29 @@ jobs:
       - name: Test
         run: make test
 
+  nix-build:
+    needs: static
```
cdunster, line 52: Does this job need the `static` job? I guess it's usually for the `rust-cache` but this job doesn't use it.

cdunster (approved):
Just a small question but still happy to merge as-is

.github/workflows/test.yaml
```diff
@@ -48,12 +48,29 @@ jobs:
...
+      fail-fast: false
+    
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v30
+
+      - run: nix build --verbose -L .#bootstrap-srv
```
ThetaSinner, line 65: Quite a bit slower than other CI checks, can we add cachix?

ThetaSinner (commented):
How come this is building without the change to tx5?

jost-s:
Because the bootstrap server doesn't use tx5.

.github/workflows/test.yaml
jost-s: I thought it's better to only attempt to build it once `static` has succeeded, same as with the test.

.github/workflows/test.yaml
jost-s: yes

flake.nix
```diff
@@ -0,0 +1,40 @@
...
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-24.11";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+    crane.url = "github:ipetkov/crane";
+  };
+
+  outputs = inputs@{ flake-parts, crane, ... }:
+    flake-parts.lib.mkFlake { inherit inputs; } {
+      systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
```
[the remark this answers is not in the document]
jost-s: I see no harm with it, but okay.

ThetaSinner:
Right, yes of course!

jost-s:
Cachix added and build times are down to 1 min.

ThetaSinner (approved):

PR #117, opened 2025-02-12
More initiate tests

ThetaSinner:
Not much changing here, just adding more tests for the initiate responder

### DIFF
--- a/crates/dht/src/arc_set.rs
+++ b/crates/dht/src/arc_set.rs
@@ -10,8 +10,7 @@ use kitsune2_api::{DhtArc, K2Error, K2Result};
 use std::collections::HashSet;
 
 /// Represents a set of [DhtArc]s.
-#[derive(Debug, Clone)]
-#[cfg_attr(test, derive(PartialEq))]
+#[derive(Debug, Clone, PartialEq)]
 pub struct ArcSet {
     inner: HashSet<u32>,
 }
--- a/crates/gossip/src/respond/harness.rs
+++ b/crates/gossip/src/respond/harness.rs
@@ -2,10 +2,12 @@ use crate::gossip::{DropAbortHandle, GossipResponse, K2Gossip};
 use crate::peer_meta_store::K2PeerMetaStore;
 use crate::protocol::{deserialize_gossip_message, GossipMessage};
 use crate::K2GossipConfig;
+use base64::Engine;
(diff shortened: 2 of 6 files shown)


crates/dht/src/arc_set.rs
```diff
@@ -10,8 +10,7 @@ use kitsune2_api::{DhtArc, K2Error, K2Result};
 use std::collections::HashSet;
 
 /// Represents a set of [DhtArc]s.
-#[derive(Debug, Clone)]
-#[cfg_attr(test, derive(PartialEq))]
```
ThetaSinner, line 14: This is still only used for tests elsewhere, I could introduce a `test-utils` feature but is it worth it?

jost-s: Ah no, that's fine.

neonphog: There's nothing incorrect about PartialEq for this type, and compiler tree-shaking should delete the code if it's not used in production. I'm happy for things like this to exist on our types.

neonphog (approved):
:+1:

crates/gossip/src/respond/harness.rs
```diff
@@ -98,6 +100,33 @@ impl RespondTestHarness {
         }
     }
 
+    pub(crate) async fn remote_agent(
```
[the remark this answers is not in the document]
ThetaSinner: Yes that would do nicely, still saves a few lines to have a function but that's much tidier

crates/gossip/src/respond/initiate.rs
```diff
@@ -247,4 +363,73 @@ mod tests {
...
+                other_peer_url.clone(),
+                GossipMessage::Initiate(K2GossipInitiateMessage {
+                    session_id: test_session_id(),
+                    participating_agents: vec![],
+                    arc_set: Some(ArcSetMessage {
+                        value: arc_set.encode(),
+                    }),
+                    new_since: Timestamp::now().as_micros(),
+                    max_op_data_bytes: 5_000,
+                }),
```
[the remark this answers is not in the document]
ThetaSinner: Yes, the type wasn't clone but this probably won't be the last place that would be helpful. Updated

jost-s (approved):
:t-rex:

Issue #25, opened 2024-11-27
[k2] publish module


neonphog:
The publish module on the send side should be a simple queue with basic timeout and limited parallelism logic to sequentially send hashes to remotes.

On the receive side it will just add those hashes to the fetch queue.

All other complexity will be deferred until performance testing shows it is required.

Try using the real production publish module as the implementation in the core crate. We can always move it out if it gets too complicated.

jost-s:
Agent infos should be published too.

matthme:
Completed with #119 and #124

PR #119, opened 2025-02-12
Add publish module

matthme:
Resolves the publish op portion of https://github.com/holochain/kitsune2/issues/25


crates/core/src/factories/core_publish/test.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    core_publish_1
+        .publish_ops(
+            vec![incoming_op_id_1.clone(), incoming_op_id_2.clone()],
+            url_2,
+        )
+        .await
+        .unwrap();
+
+    let ops = tokio::time::timeout(
```
jost-s, line 91: For the timeout + loop + sleep construct, there is an `iter_check` in test_utils.

crates/core/src/factories/core_publish/test.rs
```diff
@@ -0,0 +1,129 @@
...
+        },
+    )
+    .await
+    .unwrap();
+
+    assert!(ops.len() == 2);
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn publish_to_invalid_url_does_not_error() {
```
jost-s, line 117: I understand the intention of the test, but it doesn't test that the outgoing publish task does not error, it tests that sending to that task doesn't error. It seems there's no way to test the publish doesn't error, unless you add some test only field.

crates/core/src/factories/core_publish/message_handler.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    #[tokio::test]
+    async fn publish_ops() {
+        let (incoming_publish_ops_tx, mut incoming_publish_ops_rx) =
+            tokio::sync::mpsc::channel(1);
+        let message_handler = PublishMessageHandler {
+            incoming_publish_ops_tx,
+        };
+        let peer = Url::from_str("wss://127.0.0.1:1").unwrap();
+        let requested_op_ids = vec![random_op_id()];
```
jost-s, line 113: use `create_op_id_list`?

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,246 @@
...
+                continue;
+            };
+        }
+    }
+}
+
+impl Drop for CorePublish {
+    fn drop(&mut self) {
+        for t in self.tasks.iter() {
+            t.abort();
```
jost-s, line 243: That's how the fetch module does it too, but I've seen that in many places it's just an `AbortHandle`. Nothing else of the task handle is used, so you could adopt that pattern.

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,246 @@
...
+    ) {
+        while let Some((op_ids, peer)) = response_rx.recv().await {
+            tracing::debug!(?peer, ?op_ids, "incoming publish ops");
+
+            // Retrieve ops to send from store.
+            if let Err(err) = fetch.request_ops(op_ids.clone(), peer).await {
+                tracing::warn!(
+                        "could not insert publish ops request into fetch queue: {err}"
+                    );
+                continue;
```
jost-s, line 234:
```suggestion
```

jost-s (commented):
Looking great! I've got a couple of suggestions.

crates/core/src/factories/core_publish/message_handler.rs
matthme: funnily enough I copied that part of the code over from your fetch module :) - gonna change it in both places.

crates/core/src/factories/core_publish/test.rs
matthme: What a blessing! That's a util I'll happily use.

crates/core/src/factories/core_publish/test.rs
matthme: > it tests that sending to that task doesn't error

How do you arrive at this conclusion? I think what this test actually *is* missing indeed is a sleep so that there is time for the publish to occur in the receiving end of the channel. But then it does indeed test the publish itself I think as I'm seeing the following tracing error being logged:

```
could not send publish ops: Connection Closed (src: None) op_id=rhfL-NCni9XNqVteNtTNzxYxpIDy-js2kwDN_Wedhbo peer_url=ws://notanexistingurl:80
```

matthme: I added a sleep now at the end of that test to make sure it makes it all the way to the publish step.

crates/core/src/factories/core_publish/message_handler.rs
matthme: addressed.

crates/core/src/factories/core_publish.rs
matthme: addressed.

crates/api/src/publish.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl From<PublishOps> for Vec<OpId> {
+    fn from(value: PublishOps) -> Self {
+        value.op_ids.into_iter().map(Into::into).collect()
+    }
+}
+
+/// Serialize list of op ids to request.
+pub fn serialize_publish_ops(value: Vec<OpId>) -> Bytes {
```
ThetaSinner, line 32: Does this one need to be public? Should be used through the function below?

crates/core/src/factories/core_publish.rs
matthme: Changed that now to only store the `AbortHandle`s

crates/api/src/publish.rs
```diff
@@ -0,0 +1,131 @@
...
+pub fn serialize_publish_ops_message(value: Vec<OpId>) -> Bytes {
+    let mut out = BytesMut::new();
+    let data = serialize_publish_ops(value);
+    let publish_message = K2PublishMessage {
+        publish_message_type: PublishMessageType::Ops.into(),
+        data,
+    };
+    publish_message
+        .encode(&mut out)
+        .expect("failed to encode publish ops message");
```
ThetaSinner, line 50: Just noticing that the fetch module does this too. I'd prefer bubbling this error up if something goes wrong. This isn't very likely to fail but I'd prefer to minimize panics

Can this be changed in the fetch module too @jost-s or did you have a reason for `expect`ing these?

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,247 @@
...
+use std::sync::Arc;
+use tokio::{
+    sync::mpsc::{channel, Receiver, Sender},
+    task::AbortHandle,
+};
+
+#[cfg(test)]
+mod test;
+
+mod message_handler;
```
ThetaSinner, line 34:
```suggestion
mod message_handler;

#[cfg(test)]
mod test;
```

Really minor, I prefer to find production modules before test ones

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,247 @@
...
+        &self,
+        op_ids: Vec<OpId>,
+        target: Url,
+    ) -> BoxFut<'_, K2Result<()>> {
+        Box::pin(async move {
+            // Insert requests into publish queue.
+            for op_id in op_ids {
+                if let Err(err) = self
+                    .outgoing_publish_ops_tx
+                    .send((op_id, target.clone()))
```
ThetaSinner, line 129: Any reason not to inform the other peer of all the op ids at the same time?

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    async fn outgoing_publish_ops_task(
+        mut outgoing_publish_ops_rx: Receiver<OutgoingPublishOps>,
+        space_id: SpaceId,
+        transport: DynTransport,
+    ) {
+        while let Some((op_id, peer_url)) = outgoing_publish_ops_rx.recv().await
+        {
+            // Send ops publish message to peer.
+            let data = serialize_publish_ops_message(vec![op_id.clone()]);
```
ThetaSinner, line 205: This is where it looks slightly odd to me to publish one op at a time. The protocol was set up to publish a list but we're always publishing one thing

crates/core/src/factories/core_publish.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    async fn incoming_publish_ops_task(
+        mut response_rx: Receiver<IncomingPublishOps>,
+        fetch: DynFetch,
+    ) {
+        while let Some((op_ids, peer)) = response_rx.recv().await {
+            tracing::debug!(?peer, ?op_ids, "incoming publish ops");
+
+            // Retrieve ops to send from store.
```
ThetaSinner, line 231:
```suggestion
            // Add incoming op ids to the fetch queue to let that retrieve the op data
```

crates/core/src/factories/core_publish/test.rs
ThetaSinner: Can we check in less than 3s? I'm on the fence about this. I do want us to check that sending doesn't error when we can't reach the other side and that does require waiting a bit but a fixed sleep will mean that we always have to wait 3s for this test to pass.

I have what I think might be a better suggestion that I didn't think of earlier. What about doing this "bad send" then following it with a valid send? That will then check that everything is still working after the thing that might have caused a fault?

ThetaSinner (commented):
Looking good! Have a few suggestions, mostly minor. I think the main thing is batching vs single op ids

crates/core/src/factories/core_publish/test.rs
matthme: Sounds good to me. Would it be too cheecky to sneak that into the `published_ops_can_be_retrieved` test before sending to the valid url?

crates/core/src/factories/core_publish.rs
matthme: Good point, will change that.

crates/api/src/publish.rs
jost-s: This can only fail in case of a memory corruption. There was a conversation with David over how to handle this on some PR. He left it up to me. If there's something wrong with the memory, I think it's justified to panic.

crates/core/src/factories/core_publish.rs
matthme: addressed this one - resolving conversation because I adopted the change 1:1.

crates/core/src/factories/core_publish.rs
matthme: addressed 1:1

crates/core/src/factories/core_publish/test.rs
jost-s: `publish_ops` returns a future `K2Result` that always returns `Ok(())`.

In the function body the op ids are sent to the `outgoing_publish_ops_tx` channel. If that fails, the error is logged. If the receiving task then fails to send something, the error is again logged. But in no place is an error returned.

crates/core/src/factories/core_publish.rs
matthme: addressed.

crates/api/src/publish.rs
matthme: addressed.

crates/api/src/publish.rs
matthme: Quoting from the prost docs:
> EncodeError always indicates that a message failed to encode because the provided buffer had insufficient capacity. Message encoding is otherwise infallible.

crates/core/src/factories/core_publish/test.rs
matthme: Yes that's correct. And the test tests that things can still proceed smoothly in that case. Do you think the behavior should be different or that we should have another test that tests something else additionally?

matthme: I renamed the test now and changed the logic to what @ThetaSinner proposed above (provided that I understood it correctly).

crates/core/src/factories/core_publish/test.rs
```diff
@@ -0,0 +1,156 @@
...
+            ])
+            .await
+            .unwrap();
+
+        if ops.len() == 2 {
+            return ops;
+        }
+    });
+
+    assert!(ops.len() == 2);
```
jost-s, line 155: Yes, that looks like a good test to me.
```suggestion
    iter_check!(1000, {
        let ops = op_store_2
            .retrieve_ops(vec![
                incoming_op_id_1.clone(),
                incoming_op_id_2.clone(),
            ])
            .await
            .unwrap();

        if ops.len() == 2 {
            break;
        }
    });
```

matthme: @ThetaSinner and I were discussing that the final assert is not needed really (when writing the other test which has the same logic in that regard) but we concluded that it would be good to let it there anyway for the purpose of readability since it's harder to understand that this check is already done implicitly by breaking the loop...

ThetaSinner: I still prefer that. I found too many tests in Holochain that use macros to wait for something but the final condition was slightly off and it took a long time to decode, a really clear assertion at the end of the test makes that hard to get wrong. I could live with it being removed in this case but in general, I think it's a helpful pattern

crates/tool_proto_build/src/main.rs
```diff
@@ -4,9 +4,10 @@ fn main() {
         .bytes(["."])
         .compile_protos(
             &[
-                "../api/proto/wire.proto",
```
ThetaSinner, line 7: Is there a reason to re-order these? The wire protocol is independent of the module protocols so having it first makes sense to me

ThetaSinner (approved):

crates/tool_proto_build/src/main.rs
matthme: It was just my instinct for alphabetical order at play here because I tend to find things easier if they are ordered alphabetically. But if it does not make sense from another perspective I can revert it...

matthme: I did not consider that the order may matter for the purpose of compilation...

crates/core/src/factories/core_publish/test.rs
jost-s: Maybe we can rename the macro to something that implies a condition to break the loop, so that assertion is clear. This iter_check pattern is used in the majority of places in this repo.

jost-s (approved):
Great!

crates/tool_proto_build/src/main.rs
ThetaSinner: I'm not sure that the order here matters but looking at the code behind that function call, it does maintain order using `Vec` and `for` through the process. I'd suggest making the semantic meaning clearer here rather than sorting alphabetically. How about something like

```
&[
    // Wire protocol for sending messages over the K2 transport
    "../api/proto/wire.proto",

    // Module protocols, as the content of wire messages
    "../api/proto/fetch.proto",
    // ...
]
```

Where the module messages can be sorted alphabetically. What do you think?

matthme: Sounds good - changed it accordingly.

ThetaSinner (approved):
Great, thank you!

PR #120, opened 2025-02-12
Showcase Chat

neonphog:
chat

- switches to rustyline... reedline kept resetting the terminal and it wasn't worth the time to figure it out.
- adds a super naive chat impl
- the second peer will be able to talk to the first right away, but the first will need to wait 1 min until its next bootstrap fetch


crates/kitsune2_showcase/src/app.rs
```diff
@@ -0,0 +1,132 @@
...
+                })
+            }
+
+            fn new_listening_address(
+                &self,
+                this_url: Url,
+            ) -> BoxFut<'static, ()> {
+                let print = self.0.clone();
+                Box::pin(async move {
+                    print.print_line(format!("Online at: {this_url}"));
```
ThetaSinner, line 60: I see this printed twice on startup, with the same value. Is that expected?

ThetaSinner (approved):
Nice, gave this a go locally and it works nicely

crates/kitsune2_showcase/src/app.rs
neonphog: Yes... sortof. There is a race condition where the address could be bound before we hook up our notifier, so we also publish the new address manually, hence often getting two of these events. It's better to often get two than to sometimes get none. We can address that at a later date if we come up with a solution to it.

jost-s:
Peers don't find each other for me - what is up?
image

neonphog:
> Peers don't find each other for me - what is up?

Most likely you are not manually running a kitsune2 bootstrap server and specifying the localhost address:port as a cli parameter.

jost-s (commented):
That must have been it! =D I was a bit hasty there :grimacing: 

I couldn't see the other peer, even after a minute. When restarting that 2nd peer, the second kept sending to the previous peer url as expected, and when that doesn't succeed and times out, the app panics. Is that desired?

image

neonphog:
heh, indeed the panic is not desirable. I'll update to instead just print the error.

neonphog:
@jost-s - k, this should work better: https://github.com/holochain/kitsune2/pull/120/commits/b010513815c16d397705ada30a5873e328de12b4

jost-s:
Yeah, great! And sending is in parallel as well.

One thing is still odd, my second node never sees the first node. For me messaging is always one way. When I shut down the second node and restart it, it is reversed. So the second can see the first, but the first never sees the second.

neonphog:
> One thing is still odd, my second node never sees the first node. For me messaging is always one way. When I shut down the second node and restart it, it is reversed. So the second can see the first, but the first never sees the second.

I still think it might fix eventually if you wait long enough (or double check that you typed the port correctly). But I'm not concerned about that for merging this, because that is a real problem that we are going to have to solve in terms of gossiping agent infos around or potentially explicitly passing at least the local ones in the preflight data when establishing connections.

jost-s (approved):
Of course, this is not specific to the app.

jost-s:
I kept the terminals open and now it does work.

PR #123, opened 2025-02-18
feat(kitsune2_bootstrap_srv): Add TLS support

ThetaSinner:
Following -> https://github.com/tokio-rs/axum/blob/main/examples/tls-rustls/src/main.rs


crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -5,19 +5,27 @@ use kitsune2_bootstrap_srv::*;
...
     #[arg(long)]
     pub production: bool,
+
     /// Output tracing in json format.
     #[arg(long)]
     pub json: bool,
-    // TODO - Implement the ability to specify TLS certificates
+
+    /// The path to a TLS certificate file.
+    #[arg(long, requires = "tls_key")]
```
ThetaSinner, line 23: Try to be helpful when specifying these on the command line, they cannot be specified separately

crates/bootstrap_srv/src/test.rs
```diff
@@ -151,7 +161,19 @@ impl<'lt> PutInfo<'lt> {
             },
         );
 
-        match ureq::put(&addr).send_string(&info) {
+        let agent_builder = ureq::builder();
+        let agent = if self.use_tls {
+            let client_config = ClientConfig::builder()
+                .dangerous()
+                .with_custom_certificate_verifier(Arc::new(NoVerify))
```
ThetaSinner, line 168: Just want to test with self-signed certificates. Server validity not relevant in a test. It'd be nice if they'd expose this for testing but they only seem to do this inside their own tests and expect everyone else to use valid certs. Could also try to provision certs as part of the test but doesn't seem that helpful

crates/bootstrap_srv/src/http.rs
```diff
@@ -132,17 +138,14 @@ fn tokio_thread(
...
             let mut addrs = Vec::with_capacity(config.addrs.len());
-            let mut servers = Vec::with_capacity(config.addrs.len());
+            let mut servers: Vec<BoxFut<'static, Result<()>>> =
+                Vec::with_capacity(config.addrs.len());
+
+            let shutdown_handle = axum_server::Handle::new();
 
             for addr in config.addrs {
-                let listener = match tokio::net::TcpListener::bind(addr).await {
+                let listener = match std::net::TcpListener::bind(addr) {
```
neonphog, line 148: This is a blocking call in an async context.
- can just use `axum_server::Server::bind` instead
- or wrap this in a `spawn_blocking`
- or use [tokio into_std](https://docs.rs/tokio/1.43.0/tokio/net/struct.TcpListener.html#method.into_std)

crates/bootstrap_srv/src/test.rs
neonphog: Yep, I've had to do this exact thing multiple times too... indeed wish there was a better way.

crates/bootstrap_srv/certs/test_cert.pem
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
neonphog, line 1: How long are these valid for? We could instead use tempfile and [rcgen](https://docs.rs/rcgen/latest/rcgen/fn.generate_simple_self_signed.html).

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
neonphog: oo, that's nice :+1:

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -5,19 +5,27 @@ use kitsune2_bootstrap_srv::*;
...
+    pub tls_cert: Option<std::path::PathBuf>,
+
+    /// The path to a TLS key file.
+    #[arg(long, requires = "tls_cert")]
+    pub tls_key: Option<std::path::PathBuf>,
```
neonphog, line 28: `tls_cert_pem_file` + `tls_key_pem_file` just to be [explicit](https://github.com/holochain/sbd/blob/main/rust/sbd-server/src/config.rs#L14-L26)?

neonphog (changes requested):
couple change requests, but looks good :+1:

jost-s (dismissed):

crates/bootstrap_srv/src/http.rs
ThetaSinner: I couldn't see a way to get the local address back from option 1, wrapped in `spawn_blocking`

crates/bootstrap_srv/certs/test_cert.pem
ThetaSinner: 3 years but yes, can use `rcgen` and it should keep working

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
ThetaSinner: Okay to solve with documentation on the CLI flag rather than a longer flag? Otherwise happy to make the rename and be consistent

neonphog: Yep, I'm cool with documentation :+1:

neonphog (approved):
woot

PR #125, opened 2025-02-20
simplify space notify

neonphog:
It turns out I don't need this extra layer at all, and actually can't use it. I just need to send to/recv from peer urls. Deletes all the not-needed code + protocols.


jost-s (approved):
Yay!

PR #126, opened 2025-02-21
chore: Bump Rust

ThetaSinner:
Update to Rust 1.85 which now supports Rust 2024


Cargo.toml
```diff
@@ -21,7 +21,7 @@ homepage = "https://github.com/holochain/kitsune2"
 keywords = ["holochain", "kitsune", "p2p", "dht", "networking"]
 categories = ["network-programming"]
 license = "Apache-2.0"
-edition = "2021"
+edition = "2024"
```
ThetaSinner, line 24: This is why there are so many changes, it's 90% formatter changes and a few places where lifetimes can now be elided.

crates/core/src/factories/core_publish.rs
```diff
@@ -315,8 +315,8 @@ impl CorePublish {
             // Add incoming op ids to the fetch queue to let that retrieve the op data
             if let Err(err) = fetch.request_ops(op_ids.clone(), peer).await {
                 tracing::warn!(
-                        "could not insert publish ops request into fetch queue: {err}"
-                    );
+                    "could not insert publish ops request into fetch queue: {err}"
```
jost-s, line 318: I like this.

jost-s (dismissed):

crates/tool_proto_build/src/main.rs
```diff
@@ -14,12 +16,16 @@ fn main() {
             &["../api/proto/"],
         )
         .expect("Failed to compile api protobuf protocol files");
-    std::env::set_var("OUT_DIR", "../core/proto/gen");
+    unsafe {
```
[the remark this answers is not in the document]
ThetaSinner: Yeah this is an interesting one. It's saying that the call is safe if you have checked that you're in a single-threaded context. Which we are here.

crates/core/src/factories/core_publish.rs
ThetaSinner: I think this was actually over-indented before, it's displaying like

```
if let Err(err) = fetch.request_ops(op_ids.clone(), peer).await {
      tracing::warn!(
          "could not insert publish ops request into fetch queue: {err}"
      );
};
```

For me locally

ThetaSinner: So the closing `);` is aligned with the call and the argument is indented. That's more in line with the formatting in other languages

crates/api/src/peer_store.rs
```diff
@@ -19,7 +19,7 @@ pub trait PeerStore: 'static + Send + Sync + std::fmt::Debug {
 
     /// Get all agents from the store.
     fn get_all(&self)
-        -> BoxFut<'_, K2Result<Vec<Arc<agent::AgentInfoSigned>>>>;
+    -> BoxFut<'_, K2Result<Vec<Arc<agent::AgentInfoSigned>>>>;
```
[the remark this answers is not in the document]
ThetaSinner: Yes I agree, I don't like how this looks

flake.nix
```diff
@@ -5,17 +5,23 @@
     nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-24.11";
     flake-parts.url = "github:hercules-ci/flake-parts";
     crane.url = "github:ipetkov/crane";
+    rust-overlay = {
```
ThetaSinner, line 8: Some updates needed to the flake to support updating to a newer Rust

flake.nix
```diff
@@ -32,6 +38,9 @@
           };
         in
         {
+          # Override the per system packages to include the rust overlay
+          _module.args.pkgs = import nixpkgs { inherit system; overlays = [ (import rust-overlay) ]; };
```
ThetaSinner, line 42: See https://github.com/hercules-ci/flake-parts/issues/106#issuecomment-1399041045

crates/tool_proto_build/src/main.rs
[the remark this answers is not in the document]
neonphog: :face_palm:

neonphog (approved):

PR #127, opened 2025-02-25
chore: Downgrade to Rust edition 2021

ThetaSinner:
I didn't realize that this would force all consumers to use Rust edition 2024. That will force us to upgrade Holochain and force everyone using Holochain to upgrade too. We don't need to be doing that work right now when we're trying to get 0.5 out. Let's plan to do this some other time.

All the code changes that were made to upgrade to 2024 are still valid on 2021 but the formatting is different so I had to `cargo fmt` to undo those changes.


mattyg (approved):

neonphog (approved):
Mostly irrelevant point of clarity: It doesn't force the use of the new edition, it forces the use of a newer rust version which in turn brings with it required clippy changes. But the result is the same. Approved : )

PR #129, opened 2025-02-25
Add bootstrap client

ThetaSinner:
Adds a new bootstrap client crate. Extracted the bootstrap PUT and GET calls from core and we'll be able to re-use this crate in Holochain where we connect to the bootstrap server for debug purposes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
   "crates/api",
+  "crates/bootstrap_client",
   "crates/bootstrap_srv",
   "crates/core",
   "crates/dht",
@@ -106,6 +107,7 @@ prost-build = "0.13.3"
 # above this section.
 # --- dev-dependencies ---
 kitsune2_bootstrap_srv = { path = "crates/bootstrap_srv" }
+kitsune2_bootstrap_client = { path = "crates/bootstrap_client" }
 kitsune2_core = { version = "0.0.1-alpha3", path = "crates/core" }
 kitsune2_test_utils = { path = "crates/test_utils" }
 
--- a/crates/bootstrap_client/Cargo.toml
+++ b/crates/bootstrap_client/Cargo.toml
@@ -20,6 +20,3 @@ ureq = { workspace = true, features = ["tls"] }
 kitsune2_core = { workspace = true }
 kitsune2_test_utils = { workspace = true }
 kitsune2_bootstrap_srv = { workspace = true }
-
-[features]
-default = []
--- a/crates/bootstrap_client/Cargo.toml
+++ b/crates/bootstrap_client/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "kitsune2_bootstrap_client"
+version.workspace = true
+description = "p2p / dht communication WAN discovery bootstrapping client"
+license.workspace = true
+homepage.workspace = true
+documentation = "https://docs.rs/kitsune2_bootstrap_client"
+authors.workspace = true
+keywords.workspace = true
+categories.workspace = true
+edition.workspace = true
+
+[dependencies]
+kitsune2_api = { workspace = true }
+
+tracing = { workspace = true }
+ureq = { workspace = true, features = ["tls"] }
+
+[dev-dependencies]
+kitsune2_core = { workspace = true }
+kitsune2_test_utils = { workspace = true }
+kitsune2_bootstrap_srv = { workspace = true }
+
+[features]
+default = []
--- a/crates/bootstrap_client/src/lib.rs
+++ b/crates/bootstrap_client/src/lib.rs
@@ -1,0 +1,49 @@
+//! A client for the Kitsune2 bootstrap server.
+
+#![deny(missing_docs)]
+
+use kitsune2_api::{AgentInfoSigned, DynVerifier, K2Error, K2Result, SpaceId};
+use std::sync::Arc;
+
+/// Send the agent info, for the given space, to the bootstrap server.
+pub fn put(
+    server_url: Arc<str>,
+    agent_info: Arc<AgentInfoSigned>,
+) -> K2Result<()> {
+    let url = format!(
+        "{server_url}/bootstrap/{}/{}",
+        &agent_info.space, &agent_info.agent
+    );
+
+    let encoded = agent_info.encode()?;
+    ureq::put(&url)
+        .send_string(&encoded)
+        .map_err(|e| K2Error::other_src("Failed to put agent info", e))?;
+
+    Ok(())
+}
+
+/// Get all agent infos from the bootstrap server for the given space.
+pub fn get(
+    server_url: Arc<str>,
+    space: SpaceId,
+    verifier: DynVerifier,
+) -> K2Result<Vec<Arc<AgentInfoSigned>>> {
+    let url = format!("{server_url}/bootstrap/{}", space);
+
+    let encoded = ureq::get(&url)
+        .call()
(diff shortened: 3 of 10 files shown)


crates/bootstrap_client/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+edition.workspace = true
+
+[dependencies]
+kitsune2_api = { workspace = true }
+
+tracing = { workspace = true }
+ureq = { workspace = true, features = ["tls"] }
+
+[dev-dependencies]
+kitsune2_core = { workspace = true }
```
ThetaSinner, line 20: Circular dependency, permitted in tests

crates/bootstrap_client/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+        .map_err(|e| K2Error::other_src("Failed to put agent info", e))?;
+
+    Ok(())
+}
+
+/// Get all agent infos from the bootstrap server for the given space.
+pub fn get(
+    server_url: Arc<str>,
+    space: SpaceId,
+    verifier: DynVerifier,
```
ThetaSinner, line 30: Require a verifier so that the client can only be used to retrieve agent infos that are considered valid by the provided verifier. Of course the user can provide a bad verifier but at least they're encouraged to think about what they're getting back.

crates/bootstrap_client/tests/connect.rs
```diff
@@ -0,0 +1,36 @@
+use kitsune2_bootstrap_srv::{BootstrapSrv, Config};
+use kitsune2_core::{Ed25519LocalAgent, Ed25519Verifier};
+use kitsune2_test_utils::enable_tracing;
+use std::sync::Arc;
+
+#[test]
+fn connect_with_client() {
```
ThetaSinner, line 7: I'd like to do a TLS test too but I can't disable certificate checks when `ureq` is used directly inside the `get`/`put` functions

Cargo.toml
```diff
@@ -1,6 +1,7 @@
 [workspace]
 members = [
   "crates/api",
+  "crates/bootstrap_client",
```
neonphog, line 4:
```suggestion
  "crates/bootstrap_cli",
```

to match `srv`?

crates/bootstrap_client/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+//! A client for the Kitsune2 bootstrap server.
+
+#![deny(missing_docs)]
+
+use kitsune2_api::{AgentInfoSigned, DynVerifier, K2Error, K2Result, SpaceId};
+use std::sync::Arc;
+
+/// Send the agent info, for the given space, to the bootstrap server.
+pub fn put(
```
neonphog, line 9:
```suggestion
pub fn blocking_put(
```

Do we want to call out explicitly that these are not async functions to help us remember we need to wrap them in spawn_blockings later?

crates/bootstrap_client/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    server_url: Arc<str>,
+    agent_info: Arc<AgentInfoSigned>,
```
neonphog, line 11:
```suggestion
    server_url: &str,
    agent_info: &AgentInfoSigned,
```

I think this will be a more generally useful api... to me it makes sense to have the complexity of moving the data into the spawn_blocking closure be a concern at that level when implementing the spawn_blocking.

crates/bootstrap_client/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    ureq::put(&url)
+        .send_string(&encoded)
+        .map_err(|e| K2Error::other_src("Failed to put agent info", e))?;
+
+    Ok(())
+}
+
+/// Get all agent infos from the bootstrap server for the given space.
+pub fn get(
+    server_url: Arc<str>,
```
neonphog, line 28:
```suggestion
    server_url: &str,
```

?

Cargo.toml
jost-s: With `cli` I'd expect a CLI with a binary.

crates/bootstrap_client/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    let encoded = agent_info.encode()?;
+    ureq::put(&url)
+        .send_string(&encoded)
+        .map_err(|e| K2Error::other_src("Failed to put agent info", e))?;
+
+    Ok(())
+}
+
+/// Get all agent infos from the bootstrap server for the given space.
+pub fn get(
```
neonphog, line 27:
```suggestion
pub fn blocking_get(
```

?

crates/bootstrap_client/tests/connect.rs
neonphog: I'm not worried. TLS is a really problem of our dependencies, not us.

neonphog (changes requested):
I vote for some api changes. Don't actually care about the client/cli name.

jost-s (dismissed):

crates/bootstrap_srv/src/test.rs
```diff
@@ -849,9 +849,6 @@ fn expiration_prune() {
     assert_eq!(1, res.len());
 }
 
-/// Certificates created with:
-/// > openssl genrsa 2048 > certs/test_key.pem
-/// > openssl req -x509 -days 1000 -new -key certs/test_key.pem -out certs/test_cert.pem
```
c12i, line 854: I think keeping this comment could still be valuable

crates/bootstrap_client/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+kitsune2_test_utils = { workspace = true }
+kitsune2_bootstrap_srv = { workspace = true }
+
+[features]
```
c12i, line 24: Perhaps a comment can be added here to indicate which features will be added in the future if this is already known

ThetaSinner: Oh good point, I copy-pasted this Cargo.toml and didn't tidy up properly. This can just be removed. I don't think we need any features for now

ThetaSinner: Done: https://github.com/holochain/kitsune2/pull/129/commits/531a406292a63bb0d27852900ca937266eec7d21

crates/bootstrap_srv/src/test.rs
ThetaSinner: I originally did generate certificates for the test and added this comment. As part of the review we switched to generating the certificates in the test with `rcgen`, so this comment should have been tidied up then

crates/bootstrap_client/src/lib.rs
ThetaSinner: Done: https://github.com/holochain/kitsune2/pull/129/commits/531a406292a63bb0d27852900ca937266eec7d21

crates/bootstrap_client/src/lib.rs
ThetaSinner: Done: https://github.com/holochain/kitsune2/pull/129/commits/531a406292a63bb0d27852900ca937266eec7d21

crates/bootstrap_client/src/lib.rs
ThetaSinner: Done: https://github.com/holochain/kitsune2/pull/129/commits/531a406292a63bb0d27852900ca937266eec7d21

neonphog (approved):
:guitar:

PR #132, opened 2025-02-28
feat: Support SBD in bootstrap

ThetaSinner:
Support running the SBD signalling service in the same process as the bootstrap server

### DIFF
--- a/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
+++ b/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
@@ -62,6 +62,8 @@ pub struct Args {
     #[arg(long)]
     pub prune_interval_ms: Option<u32>,
 
+    /// If specified, this server will only handle bootstrap requests,
+    /// dropping websocket upgrade requests from sbd clients.
     #[arg(long)]
     pub no_sbd: bool,
 
--- a/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
+++ b/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
@@ -61,6 +61,25 @@ pub struct Args {
     /// Defaults: testing = 10s, production = 60s
     #[arg(long)]
     pub prune_interval_ms: Option<u32>,
+
+    /// Use this http header to determine IP address instead of the raw
+    /// TCP connection details.
+    #[cfg(feature = "sbd")]
+    #[arg(long)]
+    pub sbd_trusted_ip_header: Option<String>,
+
+    /// Limit client connections.
+    #[cfg(feature = "sbd")]
+    #[arg(long)]
+    pub sbd_limit_clients: Option<i32>,
+
+    /// If set, rate-limiting will be disabled on the server,
+    /// and clients will be informed they have an 8gbps rate limit.
+    ///
+    /// Note that this is an SBD option, but when SBD is enabled, this applies to all connections.
+    #[cfg(feature = "sbd")]
+    #[arg(long)]
+    pub sbd_disable_rate_limiting: bool,
 }
 
 fn main() {
@@ -105,6 +124,18 @@ fn main() {
     }
     if let Some(ms) = args.prune_interval_ms {
         config.prune_interval = std::time::Duration::from_millis(ms as u64);
+    }
+    #[cfg(feature = "sbd")]
+    {
+        if let Some(header) = args.sbd_trusted_ip_header {
+            config.sbd.trusted_ip_header = Some(header);
+        }
+        if let Some(limit) = args.sbd_limit_clients {
+            config.sbd.limit_clients = limit;
+        }
+        if args.sbd_disable_rate_limiting {
+            config.sbd.disable_rate_limiting = true;
+        }
     }
 
     tracing::info!(?config);
--- a/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
+++ b/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
@@ -62,14 +62,15 @@ pub struct Args {
     #[arg(long)]
     pub prune_interval_ms: Option<u32>,
 
+    #[arg(long)]
+    pub no_sbd: bool,
+
     /// Use this http header to determine IP address instead of the raw
     /// TCP connection details.
-    #[cfg(feature = "sbd")]
     #[arg(long)]
     pub sbd_trusted_ip_header: Option<String>,
 
     /// Limit client connections.
-    #[cfg(feature = "sbd")]
     #[arg(long)]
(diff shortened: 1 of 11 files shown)


crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -61,6 +61,25 @@ pub struct Args {
...
+    #[arg(long)]
+    pub sbd_limit_clients: Option<i32>,
+
+    /// If set, rate-limiting will be disabled on the server,
+    /// and clients will be informed they have an 8gbps rate limit.
+    ///
+    /// Note that this is an SBD option, but when SBD is enabled, this applies to all connections.
+    #[cfg(feature = "sbd")]
+    #[arg(long)]
+    pub sbd_disable_rate_limiting: bool,
```
neonphog, line 82: I'm totally cool with how this is currently set up.

But just musing: we will eventually need to support rate-limiting for bootstrapping anyways. It doesn't feel bad to me to always compile sbd in as a non-optional dependency to use its rate-limiting code, then optionally ignore websockets if something like a `--no-sbd` cli flag is passed.

ThetaSinner: Yes great. I partly put the feature flag in as a caution thing, so that if I messed something up, we had an easy way to get rid of the code :D But if we want some of the functionality anyway and we're happy with this approach then I'm happy to make that change.

I'll jump into that on Monday :)

jost-s:
LGTM, but I can't try it out yet.

ThetaSinner:
> I can't try it out yet.

Updated to use a branch rather than a path, I need this running too

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -61,6 +61,25 @@ pub struct Args {
     /// Defaults: testing = 10s, production = 60s
     #[arg(long)]
     pub prune_interval_ms: Option<u32>,
+
+    #[arg(long)]
+    pub no_sbd: bool,
```
neonphog, line 66:
```suggestion
    /// If specified, this server will only handle bootstrap requests,
    /// dropping websocket upgrade requests from sbd clients.
    #[arg(long)]
    pub no_sbd: bool,
```

crates/bootstrap_srv/src/http.rs
```diff
@@ -111,40 +118,78 @@ struct Ready {
...
-            let app: Router = Router::new()
+            let sbd_config = Arc::new(config.sbd.clone());
+
+            let ip_rate = Arc::new(sbd_server::IpRate::new(sbd_config.clone()));
+
+            let c_slot = if !config.no_sbd {
+                Some(sbd_server::CSlot::new(sbd_config.clone(), ip_rate.clone()))
+            } else {
+                None
+            };
```
neonphog, line 153: Double negatives are confusing : )

```suggestion
            let c_slot = if config.no_sbd {
                None
            } else {
                Some(sbd_server::CSlot::new(sbd_config.clone(), ip_rate.clone()))
            };
```

crates/bootstrap_srv/src/http.rs
```diff
@@ -111,40 +118,78 @@ struct Ready {
...
                     "/bootstrap/:space/:agent",
                     routing::put(handle_boot_put),
-                )
+                );
+
+            let app = if !config.no_sbd {
+                app.route("/:pub_key", routing::get(crate::sbd::handle_sbd))
+            } else {
+                app
+            };
```
neonphog, line 167:
```suggestion
            let app = if config.no_sbd {
                app
            } else {
                app.route("/:pub_key", routing::get(crate::sbd::handle_sbd))
            };
```

crates/bootstrap_srv/src/http.rs
```diff
@@ -168,23 +213,41 @@ fn tokio_thread(
...
                 let app = app.clone();
                 let shutdown_handle = shutdown_handle.clone();
-                if let Some(tls_config) = &config.tls_config {
+                if let Some(tls_config) = &server_config.tls_config {
                     let tls_config = tls_config
                         .create_tls_config()
                         .await
                         .expect("Failed to create TLS config");
 
+                    let acceptor = RustlsAcceptor::new(tls_config);
```
neonphog, line 222: May not need to be part of this pr, but I feel like we're going to need another custom acceptor befor the tls one here to support doing the ip deny/ratelimit checks before tls upgrade when *not* in use trusted ip mode.

crates/bootstrap_srv/src/http.rs
```diff
@@ -168,23 +213,41 @@ fn tokio_thread(
...
                         .handle(shutdown_handle)
-                        .serve(app.into_make_service());
+                        .serve(app.into_make_service_with_connect_info::<SocketAddr>());
 
                     servers.push(Box::pin(s));
                 } else {
+                    let server = axum_server::Server::from_tcp(listener);
+
+                    let server = {
+                        server.acceptor(crate::sbd::SbdAcceptor::new(
```
neonphog, line 241: and probably for consistency would make sense to use that other custom acceptor before here as well even when not in tls mode

neonphog (approved):
Couple suggestions, one that could be a follow-on. Nice work!

jost-s:
This means that SBD server is using the same port as the bootstrap server, bootstrap over http and signal over ws, right?! So I can set the address and port using `--listen`. That's good.

jost-s:
Could there be the need to have bootstrap and signal server run on different ports? Maybe not while they're using different protocols.

neonphog:
> Could there be the need to have bootstrap and signal server run on different ports? Maybe not while they're using different protocols.

You can still run the sbd-server directly by itself then run this with the --no-sbd option if you need them on different ports.

crates/bootstrap_srv/src/http.rs
ThetaSinner: These run inside out. So the acceptor that's chained onto the RustlsAcceptor runs before the Rustls acceptor does. The implementation is "call the inner acceptor if set, then apply my behavior to the stream and return it". So I think this is behaving like you want?

ThetaSinner: Or maybe I'm not following the trusted IP logic

neonphog: Okay, cool, I was not understanding that correctly. So then, `preflight_ip_check` will be called in the accept code, and the ip will not actually be checked if we are in use_trusted_ip mode. Then, in the `handle_upgraded` it will be.

Thanks! I'm happy with this :+1:

neonphog (dismissed):

ThetaSinner:
I shouldn't merge this yet, it needs to point to a released sbd server

neonphog (approved):

PR #134, opened 2025-03-03
Add docker container

ThetaSinner:
Docker container for the bootstrap server. Currently deployed at `https://bootstrap2.holochain.org`.

We will have to actually start tagging our releases to produce bootstrap a container for each release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM rust:1.85-bookworm AS builder
+
+# Pick a directory to work in
+WORKDIR /usr/src/kitsune2
+
+# Copy the workspace's source code into the container
+ADD ["Cargo.toml", "Cargo.lock", "./"]
+ADD crates ./crates
+
+# Build the bootstrap server
+RUN cargo install --path ./crates/bootstrap_srv --bin kitsune2-bootstrap-srv
+
+FROM golang:1.24-bookworm AS lego-builder
+
+# Install the lego tool
+ENV GO111MODULE=on
+RUN go install github.com/go-acme/lego/v4/cmd/lego@latest
+
+FROM debian:bookworm-slim
+
+EXPOSE 443
+
+# Copy the built binary into the runtime container
+COPY --from=builder /usr/local/cargo/bin/kitsune2-bootstrap-srv /usr/local/bin/
+COPY --from=lego-builder /go/bin/lego /usr/local/bin/
+
+# Create a directory to store the TLS certificate and key
+RUN mkdir -p /etc/bootstrap_srv
+
+# Run the bootstrap server with production settings
+CMD ["kitsune2-bootstrap-srv", "--production", "--tls-cert", "/etc/bootstrap_srv/cert.pem", "--tls-key", "/etc/bootstrap_srv/key.pem"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,19 +10,12 @@ ADD crates ./crates
 # Build the bootstrap server
 RUN cargo install --path ./crates/bootstrap_srv --bin kitsune2-bootstrap-srv
 
-FROM golang:1.24-bookworm AS lego-builder
-
-# Install the lego tool
-ENV GO111MODULE=on
-RUN go install github.com/go-acme/lego/v4/cmd/lego@latest
-
 FROM debian:bookworm-slim
 
 EXPOSE 443
 
 # Copy the built binary into the runtime container
 COPY --from=builder /usr/local/cargo/bin/kitsune2-bootstrap-srv /usr/local/bin/
-COPY --from=lego-builder /go/bin/lego /usr/local/bin/
 
 # Create a directory to store the TLS certificate and key
 RUN mkdir -p /etc/bootstrap_srv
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+EXPOSE 443
+
+# Copy the built binary into the runtime container
+COPY --from=builder /usr/local/cargo/bin/kitsune2-bootstrap-srv /usr/local/bin/
+
+# Create a directory to store the TLS certificate and key
+RUN mkdir -p /etc/bootstrap_srv
+
+# Run the bootstrap server with production settings
+CMD ["kitsune2-bootstrap-srv", "--production", "--tls-cert", "/etc/bootstrap_srv/cert.pem", "--tls-key", "/etc/bootstrap_srv/key.pem"]
```
ThetaSinner, line 24: Provide a default command more as a hint than something usable

crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -109,6 +109,12 @@ fn main() {
         Config::testing()
     };
 
+    if args.tls_cert.is_some() || args.tls_key.is_some() {
```
ThetaSinner, line 112: Turns out we do need this. I have no desire to keep endlessly fiddling with which features are enabled and it's super fragile because we won't detect this on CI and it's a runtime decision which provider to pick. With multiple providers available, it defaults to picking none. Probably sensible but then let's just decide for it

crates/bootstrap_srv/src/http.rs
```diff
@@ -190,6 +190,8 @@ fn tokio_thread(
             let shutdown_handle = axum_server::Handle::new();
 
             for addr in server_config.addrs {
+                tracing::info!("Binding to: {}", addr);
```
ThetaSinner, line 193: Had some trouble here. We were binding to `0.0.0.0:443` and `[::]:443` but with dual stack available, binding to `[::]:443` hangs after we've successfully bound to `0.0.0.0:443`. 

I can remove these debug logs but a little extra debug help probably doesn't hurt

crates/bootstrap_srv/src/test.rs
```diff
@@ -853,8 +853,7 @@ fn expiration_prune() {
 #[test]
 fn start_with_tls() {
     // We have mixed features between ring and aws_lc so the "lookup by crate features" doesn't
-    // return a default. It's only `ureq` that does that and this shouldn't be needed in the
-    // server.
+    // return a default.
```
ThetaSinner, line 856: No longer true

crates/kitsune2_showcase/src/main.rs
```diff
@@ -7,11 +7,11 @@ mod readline;
 #[derive(clap::Parser)]
 struct Args {
     /// The signal server to use.
-    #[arg(long, default_value = "wss://sbd.holo.host")]
+    #[arg(long, default_value = "wss://bootstrap2.holochain.org")]
```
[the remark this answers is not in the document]
neonphog: I still kinda lightly vote for `dev-test-bootstrap2.holochain.org` to make it clear.

neonphog (approved):
Woot! Excited to have this.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+FROM debian:bookworm-slim
+
+EXPOSE 443
+
+# Copy the built binary into the runtime container
+COPY --from=builder /usr/local/cargo/bin/kitsune2-bootstrap-srv /usr/local/bin/
+
+# Create a directory to store the TLS certificate and key
+RUN mkdir -p /etc/bootstrap_srv
```
jost-s, line 21: Should this better be a regular user accessible folder like `/usr/local/etc`?

crates/core/src/factories/core_space.rs
```diff
@@ -355,6 +355,8 @@ impl Space for CoreSpace {
 
                         // add it to bootstrapping.
                         bootstrap.put(info);
+                    } else {
+                        tracing::trace!("Not updating agent info because we don't have a current url");
```
jost-s, line 359: Isn't that worth a warning or even error?

ThetaSinner: Possibly, we won't normally see it on startup I don't think. We generally seem to avoid giving out warnings for things that are likely to correct themselves and that the user can't fix. I'd quite like to make this a warning or at least info. Given that you phrased this request as a question, what do you think?

crates/kitsune2_showcase/src/main.rs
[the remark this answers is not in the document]
ThetaSinner: Cool, I can rename this. I don't think I've merged anything that references it yet so I just need to do that across a few open PRs

mattyg (dismissed):

crates/core/src/factories/core_space.rs
jost-s: Yes, info or warning is good. I've seen that pattern as well to not log an awful lot, but even after my little experience now dealing with kitsune problems, I'm all in favor for "more is better". We're nowhere near the tracing disaster that some of the holochain modules were.

jost-s (approved):
Sweet as

PR #135, opened 2025-03-03
separate call to register kitsune handler

neonphog:
I came to the part where conductor needs to handle events emitted by kitsune, but conductor can't pass itself into the kitsune builder because it doesn't exist as a struct yet, because it takes a kitsune instance, etc, etc...

This tiny bit of complexity here should save a bunch of complexity in holochain_p2p and conductor.


crates/api/src/kitsune.rs
```diff
@@ -57,6 +57,15 @@ pub type DynKitsuneHandler = Arc<dyn KitsuneHandler>;
 
 /// The top-level Kitsune2 api trait.
 pub trait Kitsune: 'static + Send + Sync + std::fmt::Debug {
+    /// Register the kitsune handler. This should be called exactly once,
+    /// before any other apis are invoked on the instance. The only reason
+    /// this isn't in the factory is to support a struct containing a
```
matthme, line 62: It took me a while to understand what you mean by this comment. If my interpretation is correct, it could potentially be rephrased to:
"The only reason this isn't in the factory is to support structs that contain a kitsune instance in one of their fields while at the same time acting as that instance's handler"

matthme:
It looks good to me but I'll let the approval to others that have better insight into kitsune.

jost-s (dismissed):

crates/api/src/kitsune.rs
ThetaSinner: Another way to describe this would be to think of this as a dependency injection system where you want a circular dependency between two objects. The dependency injection system can't create one or the other first, so you have to have a one way dependency and then inform the object about the reverse dependency after creation.

Not sure if that's helpful, it makes sense to me that way because I'm used to working with dependency injection so that's the simplest way to think about the way this is being set up.

crates/api/src/kitsune.rs
```diff
@@ -57,6 +57,15 @@ pub type DynKitsuneHandler = Arc<dyn KitsuneHandler>;
...
+    /// kitsune instance as well as acting as its handler.
```
ThetaSinner, line 63: Should we document that implementations of this trait should error if called more than once?

ThetaSinner (approved):
This looks good, I'd prefer that the library is flexible about this rather than having the complexity in Holochain

crates/api/src/kitsune.rs
neonphog: Awesome, thanks for the feedback! Let me know how this sounds: https://github.com/holochain/kitsune2/pull/135/commits/6c5610ff0a986bcb00aaaab26f26c3438b8db4cc

crates/api/src/kitsune.rs
neonphog: Done: https://github.com/holochain/kitsune2/pull/135/commits/6c5610ff0a986bcb00aaaab26f26c3438b8db4cc

jost-s (approved):

matthme (approved):

crates/api/src/kitsune.rs
matthme: sounds good, thank you :+1:

PR #139, opened 2025-03-07
feat: Reload TLS certificates

### LINKED ISSUE #136: Reload TLS certificates when they change
We're deploying the bootstrap server with `certbot` which will automatically renew certificates. We need to load those new certificates for that to be useful!

ThetaSinner:
Closes #136

### DIFF
--- a/crates/bootstrap_srv/src/http.rs
+++ b/crates/bootstrap_srv/src/http.rs
@@ -60,6 +60,7 @@ pub struct Server {
     addrs: Vec<std::net::SocketAddr>,
     receiver: HttpReceiver,
     h_send: HSend,
+    tls_reload_handle: Option<tokio::task::AbortHandle>,
     shutdown: Option<axum_server::Handle>,
 }
 
@@ -71,6 +72,9 @@ impl Drop for Server {
         }
         if let Some(t_join) = self.t_join.take() {
             let _ = t_join.join();
+        }
+        if let Some(tls_reload_handle) = self.tls_reload_handle.take() {
+            tls_reload_handle.abort();
         }
     }
 }
@@ -89,12 +93,14 @@ impl Server {
                 h_send,
                 addrs,
                 receiver,
+                tls_reload_handle,
                 shutdown,
             })) => Ok(Self {
                 t_join: Some(t_join),
                 addrs,
                 receiver,
                 h_send,
+                tls_reload_handle,
                 shutdown: Some(shutdown),
             }),
             Ok(Err(err)) => Err(err),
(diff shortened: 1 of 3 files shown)


jost-s (dismissed):
LGTM

crates/bootstrap_srv/src/tls.rs
```diff
@@ -28,4 +29,32 @@ impl TlsConfig {
         )
         .await
     }
+
+    /// Create a future that reloads the TLS configuration every hour.
```
mattyg, line 33: I'm trying to understand the utility of this. 

When the sysadmin renews their keys, they can then either restart the server to have it take effect immediately, or wait up to 1 hour for the server for the `reload_task`

Surely 1 hour is too much downtime for any real world deployment, so the sysadmin would need to just restart the service. Does it actually make sense to include this? If so, should we use something that watches for filesystem changes and reacts to that, rather than the 1 hr schedule, such as https://docs.rs/notify/latest/notify/ ?

ThetaSinner: We could watch and reload immediately but certbot will renew certificates 30 days before renewal. So we get plenty of chances to reload the certificates before the ones that are in use expire. I think even once a day would be frequent enough.

The linked example is just an example of replacing the certificates after a dummy time period. I suspect 20s was chosen so that somebody could run the example and see the certificates change.

In a certificate revocation scenario, that would be a different story and yes you'd want the certificates to update immediately. In that case I would expect to need to restart the service. If the certificate was somehow compromised then I wouldn't be too concerned about closing a few open connections and clearing the bootstrap data, it'd probably be a good thing.

I might be thinking about this the wrong way, but that's the picture I have in my head of this.

mattyg: Ah got it okay, this is good then.

mattyg (approved):

crates/bootstrap_srv/src/http.rs
```diff
@@ -152,6 +152,19 @@ fn tokio_thread(
                 Some(sbd_server::CSlot::new(sbd_config.clone(), ip_rate.clone()))
             };
 
+            let rustls_config = if let Some(tls_config) = server_config.tls_config {
+                let rustls_config = tls_config
+                    .create_tls_config()
+                    .await
+                    .expect("Failed to create TLS config");
+
+                tokio::task::spawn(tls_config.reload_task(rustls_config.clone()));
```
neonphog, line 161: Capture abort handle and shut down the task in the server shutdown logic.

neonphog (changes requested):
Abort the reload task on drop. Otherwise looks awesome!

crates/bootstrap_srv/src/http.rs
ThetaSinner: Do'h, of course :) Done!

neonphog (approved):
Woot :+1:

PR #142, opened 2025-03-11
space accessors

neonphog:
Need some additional space accessors on the kitsune type to do the preflight sending agent infos thing.

### DIFF
--- a/crates/api/src/kitsune.rs
+++ b/crates/api/src/kitsune.rs
@@ -71,9 +71,18 @@ pub trait Kitsune: 'static + Send + Sync + std::fmt::Debug {
         handler: DynKitsuneHandler,
     ) -> BoxFut<'_, K2Result<()>>;
 
+    /// List the active spaces.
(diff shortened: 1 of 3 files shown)


matthme (commented):
It looks good to me, given my limited big-picture-view of things. I'm wondering whether we could add a few more lines in the test in `core_kitsune.rs` as we're at it, which:
* call `list_spaces()` and confirms that it's a Vec of len 1 and contains `TEST_SPACE_ID`
* call `space_if_exists(TEST_SPACE_ID)` and verifies that it returns `Some(...)`
* call `space_if_exists(NON_EXISTING_SPACE_ID)` and verifies that it returns `None`.

neonphog:
>  add a few more lines in the test

Good call :+1:, done: https://github.com/holochain/kitsune2/pull/142/commits/dc6ec01de2c05828a69b75a92286878c9e0f0d9a

jost-s (approved):

matthme (approved):

PR #144, opened 2025-03-14
Avoid computing hashes for empty slices

ThetaSinner:
I knew this was doing a lot of work on startup until the first full time slice hash was computed but it turns out, it's a silly amount of work. Tests in `holochain_p2p` were taking 45s when they only need to take 0.2s. 

We already avoid recomputing earlier full slice hashes if we have a stored count. We also avoid recomputing partial slice hashes if we already know their value. However, we were spending a lot of initialisation time on computing slice hashes from the `UNIX_TIMESTAMP` that we can easily rule out by bounding the start time of the query.

Before I got this set up right, the exists tests did a lot of complaining... so I think they are sufficient. This doesn't add any new logic or states, it's just avoiding work. So if everything that worked before still works then we're good. It should just run much faster.

Test run for `kitsune2_dht` on `main`: `test result: ok. 81 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.59s`
and after these changes: `test result: ok. 81 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.23s`

So you can see the difference even in memory. With a real database, it's much more noticeable.

### DIFF
--- a/crates/core/src/factories/mem_op_store.rs
+++ b/crates/core/src/factories/mem_op_store.rs
@@ -315,7 +315,6 @@ impl OpStore for Kitsune2MemoryOpStore {
         arc: DhtArc,
     ) -> BoxFuture<'_, K2Result<Option<Timestamp>>> {
         Box::pin(async move {
-
             Ok(self
                 .read()
                 .await
--- a/crates/core/src/factories/mem_op_store.rs
+++ b/crates/core/src/factories/mem_op_store.rs
@@ -310,6 +310,32 @@ impl OpStore for Kitsune2MemoryOpStore {
         })
     }
 
+    fn earliest_timestamp_in_arc(
+        &self,
+        arc: DhtArc,
+    ) -> BoxFuture<'_, K2Result<Option<Timestamp>>> {
+        Box::pin(async move {
+            println!(
+                "Getting earliest from: {:?}",
+                self.read().await.op_list.len()
+            );
+
+            Ok(self
+                .read()
+                .await
+                .op_list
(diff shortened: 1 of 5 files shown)


crates/core/src/factories/mem_op_store.rs
```diff
@@ -310,6 +310,32 @@ impl OpStore for Kitsune2MemoryOpStore {
         })
     }
 
+    fn earliest_timestamp_in_arc(
+        &self,
+        arc: DhtArc,
+    ) -> BoxFuture<'_, K2Result<Option<Timestamp>>> {
+        Box::pin(async move {
+            println!(
+                "Getting earliest from: {:?}",
```
mattyg, line 319:
```suggestion
                "Getting earliest from of {} ops",
```

crates/core/src/factories/mem_op_store.rs
```diff
@@ -310,6 +310,32 @@ impl OpStore for Kitsune2MemoryOpStore {
...
+                self.read().await.op_list.len()
+            );
```
ThetaSinner, line 321:
```suggestion
```

crates/core/src/factories/mem_op_store.rs
ThetaSinner: Oh my bad, that was there for debugging

mattyg (dismissed):
awesome

neonphog (approved):

jost-s (approved):

PR #146, opened 2025-03-19
DhtArc::len -> DhtArc::arc_span

neonphog:


ThetaSinner (approved):

crates/api/src/arc.rs
```diff
@@ -153,13 +153,17 @@ impl DhtArc {
         }
     }
 
-    /// Get the length of the arc.
-    pub fn len(&self) -> u32 {
+    /// The distance from the inclusive beginning of the arc
```
jost-s, line 156: Now I read this, it doesn't make sense in conjunction with the tests. If both beginning and end are inclusive, `Arc(0, 0)` would be `1` and `Arc(0, 1)` would be `2`. End is excluded seems clearer.

neonphog: The data actually means inclusive, I don't feel like we should change that at this point. And we can't have the length return Arc(0, 0) == 1 because we will overflow when Arc(0, u32::MAX) because that value cannot fit in a u32. That's why we're changing the name of the function to span, because it's not really a length, it's length minus one.

jost-s: Sure, I didn't want to change the function, just the description to say "end excluded".

jost-s: I can live with that uneasiness if that feels incoherent =)

neonphog: Ahhh, I see what you're saying. Cool, done: https://github.com/holochain/kitsune2/pull/146/commits/52eba9dc50326326fbded9326e82356383efcb17

jost-s (approved):

crates/api/src/arc.rs
jost-s: Thanks!

PR #147, opened 2025-03-20
Gossip improvements

ThetaSinner:
Makes K2 gossip easier to debug by adding more trace logging

### DIFF
--- a/crates/dht/src/time.rs
+++ b/crates/dht/src/time.rs
@@ -183,7 +183,7 @@ impl TimePartition {
 
             // Index is one less than count, so we set the count to cover the previous index
             pt.full_slices = index.saturating_sub(1) as u64;
-            tracing::info!("Corrected full slices to {}", pt.full_slices);
+            tracing::debug!("Corrected full slices to {}", pt.full_slices);
         }
 
         // The end timestamp of the last full slice
--- a/crates/gossip/src/respond/accept.rs
+++ b/crates/gossip/src/respond/accept.rs
@@ -64,6 +64,7 @@ impl K2Gossip {
         // The remaining limit will be used for the DHT diff as required.
         if let Some(state) = lock.as_mut() {
             tracing::debug!(
+                ?accept.session_id,
                 "Used {}/{} op budget to send {} op ids",
                 used_bytes,
                 accept.max_op_data_bytes,
@@ -174,6 +175,7 @@ impl K2Gossip {
                             s => {
                                 // Other snapshot types are not expected at this point.
                                 tracing::error!(
+                                        ?accept.session_id,
                                     "unexpected snapshot type: {:?}",
                                     s
                                 );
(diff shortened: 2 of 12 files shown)


crates/dht/src/time.rs
```diff
@@ -183,7 +183,7 @@ impl TimePartition {
 
             // Index is one less than count, so we set the count to cover the previous index
             pt.full_slices = index.saturating_sub(1) as u64;
-            tracing::info!("Corrected full slices to {}", pt.full_slices);
+            tracing::debug!("Corrected full slices to {}", pt.full_slices);
```
ThetaSinner, line 186: This is really noisy, it gets logged for each sector. It's useful to see it happening if you're debugging but not at `info`

crates/gossip/src/respond/accept.rs
```diff
@@ -64,6 +64,7 @@ impl K2Gossip {
         // The remaining limit will be used for the DHT diff as required.
         if let Some(state) = lock.as_mut() {
             tracing::debug!(
+                ?accept.session_id,
```
ThetaSinner, line 67: This is the main change here, I wanted to be able to correlate log messages for gossip sessions

crates/gossip/src/respond/terminate.rs
```diff
@@ -26,9 +27,11 @@ impl K2Gossip {
                             "Unsolicited terminate message from: {from_peer}"
                         )));
                     }
-                }
 
-                true
+                    true
```
ThetaSinner, line 31: This could have wrongly treated a session as having been cleaned up as an initiated session when it was actually accepted. I believe this is what was causing the session timeouts I was seeing while debugging in Holochain.

ThetaSinner: Worst case scenario, the session has to time out if it's not cleaned up here. That wastes a little bit of time and claims a spot in the accepted sessions (we bounded that) but it's otherwise not very serious

cdunster: It might be nice to add some unit tests, as a separate PR, for this code and the code below to check that all branches are hit.

ThetaSinner: Yes, that is still on my list of things to do. I'll look for slack time on integration to come and work on this

crates/gossip/src/respond/accept.rs
ThetaSinner: Yes, at the end as a field

```
2025-03-20T15:35:49.643996Z DEBUG kitsune2_gossip::respond::accept: Used 128/104857600 op budget to send 1 op ids accept.session_id=b"?\xafj1\x8b\x93\x96U\xe0\xd2^\xe3"
```

cdunster: Ah okay, it really just gets tagged on the end then without any punctuation at all. It's still clear though, thanks.

cdunster (approved):

crates/gossip/src/respond/accept.rs
ThetaSinner: Oh the CLI it's printed in bold so it's easier to see :)

Issue #95, opened 2025-02-03
[k2] SPIKE network stats / dump / system state protocol


neonphog:
We either need to just hook up the network stats just as they were through legacy kitsune, or decide if we want a more versatile debugging/state of the system protocol that can send notifications of changes of system state. (e.g. op received / validated, peer discovered and has authority overlap, etc).

ThetaSinner:
I think we should keep the network dump similar to what it is now for tx5 data. Anything that we're returning that's specific to the peer metrics will have to be removed for now. When we get to https://github.com/holochain/holochain/issues/3315 for 0.6.

Exposing op metadata we should absolutely do too. It requires a bit of thought to translate op data into something that an app can understand in terms of actions/entries. But exposing the stored_at field and the validation status would be super. Again, I think we should tackle that later and just get Kitsune2 integrated. 

The `NetworkInfo` request on the app interface will break. We would need to update Kitsune2 to support tracking how many bytes are being sent between peers. This is currently used by `require_initial_gossip_activity_for_cell` in Sweettest and that is used in quite a few tests to check that two peers are connected and talking, so we probably can't do without that.

The `DumpNetworkMetrics` request to the admin interface dumps the legacy peer metrics that are stored by K1. This I think is okay to break because the content is specific to the K1 implementation

```
pub struct Metrics {
    /// Map of remote agents and gossip history with each.
    agent_history: HashMap<Arc<KitsuneAgent>, PeerAgentHistory>,

    /// Map of remote nodes and gossip history with each
    node_history: HashMap<NodeId, PeerNodeHistory>,

    /// Aggregate Extrapolated Dht Coverage
    agg_extrap_cov: RunAvg,

    // Number of times we need to force initiate
    // the next round.
    pub(crate) force_initiates: u8,
}
```

`DumpNetworkStats` is more useful and dumps agent information. This one I think gets used and we should look at maintaining. It gets tx5 connection information and combines it with agent infos to produce a peer connectivity report.

PR #150, opened 2025-03-21
feat: Add state dumps for fetch and gossip modules

ThetaSinner:
Contributes to #95 but adding state dumps for fetch and gossip. The gossip dump optionally includes a rather detailed DHT dump.

Note that dumping the DHT state one segment at a time would make it easier to compare. We might want to consider making that change when we go to sharding. But for now, the arc is either going to empty and the summary will be too or it'll be full and splitting by segment is a little pointless.

### DIFF
--- a/crates/api/src/fetch.rs
+++ b/crates/api/src/fetch.rs
@@ -1,12 +1,13 @@
 //! Kitsune2 fetch types.
 
-use crate::op_store;
 use crate::{
     builder, config, transport::DynTransport, BoxFut, DynOpStore, K2Result,
     OpId, SpaceId, Url,
 };
+use crate::{op_store, Timestamp};
 use bytes::{Bytes, BytesMut};
 use prost::Message;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 pub(crate) mod proto {
@@ -93,6 +94,9 @@ pub trait Fetch: 'static + Send + Sync + std::fmt::Debug {
         op_ids: Vec<OpId>,
         source: Url,
     ) -> BoxFut<'_, K2Result<()>>;
+
+    /// Get a state summary from the fetch module.
+    fn get_state_summary(&self) -> BoxFut<'_, K2Result<FetchStateSummary>>;
 }
 
 /// Trait object [Fetch].
@@ -119,6 +123,22 @@ pub trait FetchFactory: 'static + Send + Sync + std::fmt::Debug {
 
 /// Trait object [FetchFactory].
 pub type DynFetchFactory = Arc<dyn FetchFactory>;
+
+/// Summary of the fetch state.
+#[derive(Debug)]
+pub struct FetchStateSummary {
+    /// The op ids that are currently being fetched.
+    ///
+    /// Each op id is associated with one or more peer URL from which the op data could be
+    /// requested.
+    pub pending_requests: HashMap<OpId, Vec<Url>>,
+
+    /// The peer URL for nodes that are currently on backoff because of failed fetch requests.
+    ///
+    /// If peers are in here then they are not being used as potential sources in
+    /// [`FetchStateSummary::pending_requests`].
+    pub peers_on_backoff: HashMap<Url, Timestamp>,
+}
 
 #[cfg(test)]
 mod test {
--- a/crates/api/src/fetch.rs
+++ b/crates/api/src/fetch.rs
@@ -133,7 +133,7 @@ pub struct FetchStateSummary {
     /// requested.
     pub pending_requests: HashMap<OpId, Vec<Url>>,
 
-    /// The peer URL for nodes that are currently on backoff because of failed fetch requests.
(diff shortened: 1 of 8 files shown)


crates/gossip/src/summary.rs
```diff
@@ -0,0 +1,194 @@
...
+/// Gossip round state summary.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GossipRoundStateSummary {
+    /// The URL of the peer with which the round is initiated.
+    pub session_with_peer: Url,
+}
+
+/// Gossip state summary.
+#[derive(Debug, Serialize, Deserialize)]
+pub struct GossipSummary {
```
ThetaSinner, line 52: This summary is fairly specific to this gossip implementation. We could move it to the API and make the dump typed. I decided to do that for fetch because I think that's probably quite stable but this doesn't.

As it is, this is just returned as JSON.

neonphog (approved):

crates/api/src/fetch.rs
```diff
@@ -120,6 +124,22 @@ pub trait FetchFactory: 'static + Send + Sync + std::fmt::Debug {
...
+/// Summary of the fetch state.
+#[derive(Debug)]
+pub struct FetchStateSummary {
+    /// The op ids that are currently being fetched.
+    ///
+    /// Each op id is associated with one or more peer URL from which the op data could be
+    /// requested.
+    pub pending_requests: HashMap<OpId, Vec<Url>>,
+
+    /// The peer URL for nodes that are currently on backoff because of failed fetch requests.
```
mattyg, line 136:
```suggestion
    /// The peer URL for nodes that are currently on backoff because of failed fetch requests, and the timestamp when that backoff will expire.
```

crates/gossip/src/summary.rs
```diff
@@ -0,0 +1,194 @@
...
+                        .await?,
+                    peer_timeouts: self
+                        .peer_meta_store
+                        .peer_timeouts(url.clone())
+                        .await?,
+                },
+            );
+        }
+
+        serde_json::to_value(summary)
```
mattyg, line 191: Can we return the typed response? As a precursor to typing the conductor response too.

crates/gossip/src/summary.rs
mattyg: Why does it matter if its not stable? Holochain can still dump as json if we aren't ready to type its admin api dump responses.

crates/gossip/src/summary.rs
mattyg: Saw your comment above and replied there.

crates/gossip/src/gossip.rs
```diff
@@ -1115,4 +1122,46 @@ mod test {
...
+        let harness_1 = factory.new_instance().await;
+        harness_1.join_local_agent(DhtArc::FULL).await;
+
+        let harness_2 = factory.new_instance().await;
+        let local_agent_2 = harness_2.join_local_agent(DhtArc::FULL).await;
+        let op_2 = MemoryOp::new(Timestamp::now(), vec![2; 128]);
+        let op_id_2 = op_2.compute_op_id();
+        harness_2
+            .op_store
+            .process_incoming_ops(vec![op_2.clone().into()])
```
jost-s, line 1142: Not for this PR, but after you made me aware of it, I think it'd be better to have a test function to directly add ops to the store, instead of using this method.

crates/gossip/src/summary.rs
```diff
@@ -0,0 +1,139 @@
...
+                            },
+                            DhtSegmentState {
+                                disc_top_hash,
+                                disc_boundary,
+                                ring_top_hashes,
+                            },
+                        );
+                    }
+                    _ => {
+                        return Err(K2Error::other("unexpected snapshot type"))
```
jost-s, line 82: Panic here, because that's an implementation fault.

jost-s (approved):
Awesome data summaries!

crates/gossip/src/summary.rs
ThetaSinner: Yes -> https://github.com/holochain/kitsune2/pull/150/commits/9605bdb57e0db0a7d614bea3ebdc9fa4baba5e45

crates/gossip/src/gossip.rs
jost-s: Well this is not what a host would do when adding ops to the store. It's called when new ops are coming in and has side-effects. In general it seems better to have one method for ops coming in through gossip and one for the host writing new ops to the store.

jost-s (approved):

crates/gossip/src/gossip.rs
ThetaSinner: Right, fair. So it will have the effect of storing the op in the memory op store and also immediately make it "integrated". We should probably be able to test this both ways, like this where it's convenient to make data available right away and also where it's not visible yet.

PR #152, opened 2025-03-24
feat: Expose the transport from the `Kitsune` trait

ThetaSinner:
If we want to be able to get network stats, we need to access the `Transport` instance that Kitsune is using


neonphog (approved):
PR title misleading. I approve of exposing it from the top-level kitsune module.

ThetaSinner:
> PR title misleading. I approve of exposing it from the top-level kitsune module.

It really is, sorry

PR #158, opened 2025-03-26
fix: Bootstrap bad URLs

ThetaSinner:


crates/bootstrap_client/src/lib.rs
```diff
@@ -2,24 +2,27 @@
...
 ///
 /// Note the `blocking_` prefix. This is a hint to the caller that if the function is used in
 /// an async context, it should be treated as a blocking operation.
 pub fn blocking_put(
-    server_url: &str,
+    mut server_url: Url,
     agent_info: &AgentInfoSigned,
 ) -> K2Result<()> {
-    let url = format!(
-        "{server_url}/bootstrap/{}/{}",
```
ThetaSinner, line 17: When the `server_url` ends with a `/`, we ended up with `//` here and although that is a valid URL, the bootstrap server was returning a 404... Which I'm pretty sure is wrong, but I'd rather produce simple URLs and not argue with the HTTP server

crates/bootstrap_client/src/lib.rs
```diff
@@ -2,24 +2,27 @@
...
-        &agent_info.space, &agent_info.agent
-    );
+    server_url.set_path(&format!(
+        "bootstrap/{}/{}",
+        base64::prelude::BASE64_URL_SAFE_NO_PAD.encode(&**agent_info.space),
```
ThetaSinner, line 20: The display implementation can be overridden by the host, which we do in Holochain to print the space id as a DNA hash... Which rather messes this up :)

neonphog (approved):

PR #164, opened 2025-04-02
chore: Update to 0.1.0

ThetaSinner:
The integration into Holochain is nearing completion. It's looking like Kitsune2 is ready to be used at this point and we can commit to a proper version number.


neonphog (approved):

neonphog (approved):
yep, need those now

PR #167, opened 2025-04-03
refactor(core)!:process_incoming_ops in MemOpStore returns all op ids

jost-s:
This was creating a test problem where ops received multiple times were not removed from the request queue, so would be requested indefinitely.


ThetaSinner (approved):
Nice!

mattyg (approved):

PR #169, opened 2025-04-08
validation error if config disallows plain text, but plain text sbd used

neonphog:


ThetaSinner (approved):
Nice, that's a helpful change!

jost-s (approved):
Yes!

PR #176, opened 2025-04-10
feat: Broadcast agent info when declaring a new storage arc

### LINKED ISSUE #172: Publish a new agent info when updating storage arc

ThetaSinner:
Closes #172

### DIFF
--- a/crates/core/src/factories/core_space.rs
+++ b/crates/core/src/factories/core_space.rs
@@ -1,6 +1,7 @@
 //! The core space implementation provided by Kitsune2.
 
 use kitsune2_api::*;
+use std::collections::HashSet;
 use std::sync::{Arc, Mutex, Weak};
 
 /// CoreSpace configuration types.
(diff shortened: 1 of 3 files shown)


crates/core/src/factories/core_space.rs
```diff
@@ -374,7 +375,12 @@ impl Space for CoreSpace {
                         }
 
                         // add it to bootstrapping.
-                        bootstrap.put(info);
+                        bootstrap.put(info.clone());
+
+                        // and send it to our peers.
+                        if let Err(err) = broadcast_agent_info(peer_store, local_agent_store, publish, info).await {
```
ThetaSinner, line 381: Turns out we never hooked this up. We have the ability to broadcast agent info to our peers rather than waiting for them to bootstrap but we need to actually do that when updating our arc.

neonphog (approved):
Sweet. This is going to need to look different with sharding, but we're not thinking about that yet. :see_no_evil:

ThetaSinner:
Yeah I deliberately ignored the "who should we send the update to" and just blasted the network. I know that's noisy but we're on such small networks most of the time. Sub 20. It's not worth worrying about just yet.

crates/core/src/factories/core_space/test.rs
```diff
@@ -296,3 +296,125 @@ async fn space_local_agent_periodic_re_sign_and_bootstrap() {
...
+        .unwrap();
+
+    let h: DynKitsuneHandler = Arc::new(K);
+    let k1 = builder.build().await.unwrap();
+    k1.register_handler(h).await.unwrap();
+
+    let s1 = k1.space(TEST_SPACE_ID.clone()).await.unwrap();
+
+    // Join alice to the space and then remove her from the local agent store.
+    // This is a quick way to create a new agent info into the peer store for alice.
```
jost-s, line 400: Why is that?

jost-s (approved):
Great!

crates/core/src/factories/core_space/test.rs
ThetaSinner: I want a remote peer to push agent info to but don't want to set up a test harness that has two K2 instances

PR #178, opened 2025-04-10
feat(core): filter out requests for ops that are already in the store

### LINKED ISSUE #174: Fetch to filter out ops we already have before adding them to the queue

jost-s:
resolves #174

### DIFF
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -145,6 +145,7 @@ struct CoreFetch {
     state: Arc<Mutex<State>>,
     outgoing_request_tx: Sender<OutgoingRequest>,
     tasks: Vec<JoinHandle<()>>,
+    op_store: DynOpStore,
     #[cfg(test)]
     message_handler: DynTxModuleHandler,
 }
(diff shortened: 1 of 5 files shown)


neonphog (dismissed):
Looks good. Probably not worth checking again when re-adding ops after a failure, right?

jost-s:
You mean after a failure in `request_ops`? That would not be a different code path. This also catches the case we've seen the most where there are ops after the latest gossip bookmark a node has with a peer, so when gossiping again with them, the same ops are seen as new and passed to `fetch` again.

crates/core/src/factories/core_fetch.rs
```diff
@@ -166,20 +167,32 @@ impl Fetch for CoreFetch {
...
-                    .into_iter()
-                    .map(|op_id| (op_id.clone(), source.clone())),
-            );
-        }
-
         Box::pin(async move {
+            // Filter out requests for ops that are already in the op store.
+            let existing_op_ids = self
+                .op_store
+                .retrieve_ops(op_ids.clone())
```
ThetaSinner, line 174: This requires loading the ops out of the database and unpacking the msgpack blobs. Should we add an operation to the op store API that just checks if those ops are present?

ThetaSinner (commented):
Looks good, just a question about efficiency

crates/core/src/factories/core_fetch.rs
jost-s: I like that. I was hesitant to add another operation. But there's no good reason.

jost-s: How about now?

PR #179, opened 2025-04-10
Showcase Fixes

neonphog:


jost-s (commented):
Do we have tests that use a remote signal server to prove that webrtc connections can be established? Or does that rather belong into tx5?

neonphog:
> Do we have tests that use a remote signal server to prove that webrtc connections can be established? Or does that rather belong into tx5?

A remote signal server is irrelevant to webrtc connection establishment. The problem is when the nodes trying to connect to each other are not on the same machine. And there's no way to test that in CI.

neonphog:
.. technically not ***no*** way, but not something that is not easy to set up

crates/kitsune2_showcase/src/app.rs
```diff
@@ -78,28 +88,41 @@ impl App {
...
         Ok(Self {
             _k: k,
+            t,
             s,
             a,
             p: print,
         })
     }
 
+    pub async fn stats(&self) -> K2Result<()> {
```
ThetaSinner, line 120: Nice

ThetaSinner (approved):

mattyg (approved):

PR #180, opened 2025-04-11
Initiate faster after failure

### LINKED ISSUE #177: Retrigger gossip initiation if it fails
If we fail to initiate then we shouldn't wait the full initiation interval to try again, we should retry as soon as we discover the failure.

ThetaSinner:
Closes #177

This PR has created some flaky tests. The cause will be resolved by #182, so that should be merged before this one

### DIFF
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -2,6 +2,7 @@ use back_off::BackOffList;
 use kitsune2_api::*;
 use message_handler::FetchMessageHandler;
 use std::collections::HashMap;
+use std::sync::Weak;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex},
@@ -244,7 +245,7 @@ impl CoreFetch {
                     outgoing_request_tx.clone(),
                     outgoing_request_rx.clone(),
                     space_id.clone(),
-                    transport.clone(),
+                    Arc::downgrade(&transport),
                     config.re_insert_outgoing_request_delay_ms,
                 ));
             tasks.push(request_task);
@@ -255,7 +256,7 @@ impl CoreFetch {
             tokio::task::spawn(CoreFetch::incoming_request_task(
                 incoming_request_rx,
                 op_store.clone(),
-                transport.clone(),
+                Arc::downgrade(&transport),
                 space_id.clone(),
             ));
         tasks.push(incoming_request_task);
@@ -295,12 +296,19 @@ impl CoreFetch {
         outgoing_request_tx: Sender<OutgoingRequest>,
         outgoing_request_rx: Arc<tokio::sync::Mutex<Receiver<OutgoingRequest>>>,
         space_id: SpaceId,
-        transport: DynTransport,
+        transport: Weak<dyn Transport>,
         re_insert_outgoing_request_delay: u32,
     ) {
         while let Some((op_id, peer_url)) =
             outgoing_request_rx.lock().await.recv().await
         {
+            let Some(transport) = transport.upgrade() else {
+                tracing::info!(
+                    "Transport dropped, stopping outgoing request task"
+                );
+                break;
+            };
+
             let is_peer_on_back_off = {
                 let mut lock = state.lock().unwrap();
 
@@ -385,11 +393,18 @@ impl CoreFetch {
     async fn incoming_request_task(
         mut response_rx: Receiver<IncomingRequest>,
         op_store: DynOpStore,
-        transport: DynTransport,
+        transport: Weak<dyn Transport>,
         space_id: SpaceId,
     ) {
         while let Some((op_ids, peer)) = response_rx.recv().await {
             tracing::debug!(?peer, ?op_ids, "incoming request");
+
+            let Some(transport) = transport.upgrade() else {
+                tracing::info!(
+                    "Transport dropped, stopping incoming request task"
+                );
+                break;
+            };
 
             // Retrieve ops to send from store.
             let ops = match op_store.retrieve_ops(op_ids.clone()).await {
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -2,7 +2,6 @@ use back_off::BackOffList;
 use kitsune2_api::*;
 use message_handler::FetchMessageHandler;
 use std::collections::HashMap;
-use std::sync::Weak;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex},
@@ -296,7 +295,7 @@ impl CoreFetch {
         outgoing_request_tx: Sender<OutgoingRequest>,
         outgoing_request_rx: Arc<tokio::sync::Mutex<Receiver<OutgoingRequest>>>,
         space_id: SpaceId,
-        transport: Weak<dyn Transport>,
+        transport: WeakDynTransport,
         re_insert_outgoing_request_delay: u32,
     ) {
         while let Some((op_id, peer_url)) =
@@ -393,7 +392,7 @@ impl CoreFetch {
     async fn incoming_request_task(
         mut response_rx: Receiver<IncomingRequest>,
         op_store: DynOpStore,
-        transport: Weak<dyn Transport>,
+        transport: WeakDynTransport,
         space_id: SpaceId,
     ) {
         while let Some((op_ids, peer)) = response_rx.recv().await {
--- a/crates/core/src/factories/core_space.rs
+++ b/crates/core/src/factories/core_space.rs
@@ -214,6 +214,8 @@ struct CoreSpace {
 
 impl Drop for CoreSpace {
     fn drop(&mut self) {
+        tracing::trace!("Dropping core space");
+
         self.task_check_agent_infos.abort();
     }
 }
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -14,7 +14,7 @@ use bytes::Bytes;
 use kitsune2_api::*;
 use kitsune2_dht::{Dht, DhtApi};
 use std::collections::HashMap;
-use std::sync::{Arc, Weak};
+use std::sync::{Arc, OnceLock, Weak};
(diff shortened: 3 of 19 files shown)


crates/core/src/factories/core_fetch.rs
```diff
@@ -240,7 +241,7 @@ impl CoreFetch {
                     outgoing_request_tx.clone(),
                     outgoing_request_rx.clone(),
                     space_id.clone(),
-                    transport.clone(),
+                    Arc::downgrade(&transport),
```
ThetaSinner, line 244: To test this change, I needed offline agents. The simplest way to do that was to start a test harness and shut it down. At that point, I discovered we can't shut down K2 properly. I don't know if I caught everything but this was the common issue that prevented gossip from shutting down at least.

The modules I've made this change to register themselves as module handlers, so the transport has a reference to us. We can't then have an owned reference to the transport and expect everything to drop correctly.

crates/core/src/factories/core_space.rs
```diff
@@ -214,6 +214,8 @@ struct CoreSpace {
 
 impl Drop for CoreSpace {
     fn drop(&mut self) {
+        tracing::trace!("Dropping core space");
```
ThetaSinner, line 217: This was helpful for debugging, I'd like to keep this at `trace` level but I don't have to

crates/gossip/src/gossip.rs
```diff
@@ -118,11 +115,10 @@ pub(crate) struct K2Gossip {
...
     pub(crate) agent_verifier: DynVerifier,
-    pub(crate) response_tx: Sender<GossipResponse>,
-    pub(crate) _response_task: Arc<DropAbortHandle>,
-    pub(crate) _initiate_task: Arc<Option<DropAbortHandle>>,
-    pub(crate) _timeout_task: Arc<Option<DropAbortHandle>>,
-    pub(crate) _dht_update_task: Arc<Option<DropAbortHandle>>,
+    pub(crate) transport: Weak<dyn Transport>,
+    pub(crate) _initiate_task: Arc<std::sync::Mutex<Option<DropAbortHandle>>>,
+    pub(crate) _timeout_task: Arc<std::sync::Mutex<Option<DropAbortHandle>>>,
+    pub(crate) _dht_update_task: Arc<std::sync::Mutex<Option<DropAbortHandle>>>,
```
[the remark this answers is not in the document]
ThetaSinner: I could probably use a RefCell but I don't think it's that important here

crates/gossip/src/respond/accept.rs
```diff
@@ -25,13 +25,6 @@ impl K2Gossip {
         let (mut lock, initiated) =
             self.check_accept_state(&from_peer, &accept).await?;
 
-        // Only once the other peer has accepted should we record that we've tried to
-        // gossip with them. Otherwise, we risk lock each other out if we both record
-        // a last gossip timestamp and try to initiate at the same time.
-        self.peer_meta_store
```
ThetaSinner, line 31: Moved, see above

crates/gossip/src/respond/harness.rs
```diff
@@ -38,19 +39,20 @@ impl RespondTestHarness {
...
-        impl TxHandler for NoopHandler {}
-
-        let transport = builder
-            .transport
-            .create(builder.clone(), Arc::new(NoopHandler))
-            .await
-            .unwrap();
-
-        let (tx, rx) = tokio::sync::mpsc::channel(1);
+        let (tx, rx) = tokio::sync::mpsc::channel(100);
```
ThetaSinner, line 42: Now that the way the gossip module handles sending messages has changed, we're forced to mock the transport to be able to receive outgoing messages and check them in tests.

Well, we could use the transport, but we're forced to do something other than read the outgoing messages directly from the gossip module.

crates/gossip/src/timeout.rs
```diff
@@ -59,6 +69,14 @@ async fn remove_timed_out_rounds(
                     );
                 }
 
+                // If we failed to initiate, then we want to try again. Otherwise, an unavailable
+                // peer can prevent us from initiating gossip.
+                if matches!(state.stage, RoundStage::Initiated(_)) {
```
ThetaSinner, line 74: On a timeout in the initiate stage, trigger initiate to run again. I think that's the only time we should do it on a timeout.

crates/gossip/src/gossip.rs
```diff
@@ -279,11 +253,20 @@ impl K2Gossip {
             target_peer_url,
         );
 
-        send_gossip_message(
-            &self.response_tx,
-            target_peer_url,
+        // Record that we're initiating gossip with this peer.
```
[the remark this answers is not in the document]
ThetaSinner: https://github.com/holochain/kitsune2/issues/181

crates/core/src/factories/core_fetch.rs
```diff
@@ -295,12 +296,19 @@ impl CoreFetch {
         outgoing_request_tx: Sender<OutgoingRequest>,
         outgoing_request_rx: Arc<tokio::sync::Mutex<Receiver<OutgoingRequest>>>,
         space_id: SpaceId,
-        transport: DynTransport,
+        transport: Weak<dyn Transport>,
```
neonphog, line 299: This feels like it may be a pretty common pattern. Do you think we should create an official `pub type WeakTransport = Weak<dyn Transport>;` in the api crate?

crates/core/src/factories/core_fetch.rs
```diff
@@ -295,12 +296,19 @@ impl CoreFetch {
...
         re_insert_outgoing_request_delay: u32,
     ) {
         while let Some((op_id, peer_url)) =
             outgoing_request_rx.lock().await.recv().await
         {
+            let Some(transport) = transport.upgrade() else {
```
neonphog, line 305: wuuuuuut is this syntax ?!?/*&#* :exploding_head:

crates/core/src/factories/core_space.rs
neonphog: Seems like tracing to me :+1:

crates/gossip/src/gossip.rs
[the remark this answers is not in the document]
neonphog: RefCell isn't sync. But you could use [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html)s. But this is also fine.

neonphog (dismissed):
Nice, looks good to me :+1:

crates/gossip/src/gossip.rs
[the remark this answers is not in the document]
ThetaSinner: OnceLock would better express what this is doing, switched https://github.com/holochain/kitsune2/pull/180/commits/7d7f9b31be600c854209937010b3e71aecc2d7ed

crates/core/src/factories/core_fetch.rs
ThetaSinner: Cool right :) It lets you flatten out the let/if/let/else/return that is so often a useful pattern but ends up with nested code. I saw it in a diff on some other project when trying to track down a bug and decided it was worth remembering :D

crates/gossip/src/initiate.rs
```diff
@@ -58,14 +75,18 @@ pub fn spawn_initiate_task(
...
+                            rush_to_first_initiated_round = false;
                             tracing::info!("Initiated gossip with {}", url);
                         }
                         Ok(false) => {
                             // Don't log here, will already have logged the reason
                         }
                         Err(e) => {
                             tracing::error!("Error initiating gossip: {:?}", e);
+
+                            if let Err(err) = here_force_initiate.try_send(()) {
```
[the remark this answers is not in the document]
ThetaSinner: Actually, this is not so serious. We should only stay in this loop as long as there are peers we haven't attempted to gossip with recently. So it could be noisy if something is broken but it shouldn't spin endlessly.

crates/core/src/factories/core_fetch.rs
ThetaSinner: Yes, that's a good idea. That gives me a place to write about this in the code for people to find

https://github.com/holochain/kitsune2/pull/180/commits/5996eca03de6b117bc6a3123c269b08bb8ae4423
https://github.com/holochain/kitsune2/pull/180/commits/27adb70458409647d72a28478b67eb7ddc30882c
https://github.com/holochain/kitsune2/pull/180/commits/77ec3588227677845cfa3f6d2c96dcae3c50048f

neonphog: woot :+1:

neonphog (approved):

crates/gossip/src/gossip.rs
```diff
@@ -1009,4 +1018,103 @@ mod test {
...
+                    .await
+                    .unwrap()
+                    .is_some();
+            }
+
+            if all_tried {
+                break;
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
```
jost-s, line 1099: The sleep is included in the macro.
```suggestion
```

crates/gossip/src/gossip.rs
```diff
@@ -1009,4 +1018,103 @@ mod test {
...
+                .peer_meta_store
+                .last_gossip_timestamp(good_agent_url.clone())
+                .await
+                .unwrap()
+                .is_some()
+            {
+                break;
+            }
+
+            tokio::time::sleep(std::time::Duration::from_millis(10)).await;
```
jost-s, line 1114:
```suggestion
```

crates/gossip/src/harness.rs
```diff
@@ -264,6 +264,56 @@ impl K2GossipFunctionalTestHarness {
...
+                    };
+
+                    if agent.storage_arc == arc {
+                        break;
+                    }
+
+                    tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+                }
+            }
+        }).await.expect("Timed out waiting for agent to be in peer store with the requested arc");
```
jost-s, line 307: Use the `iter_check` here too?

PR #182, opened 2025-04-11
feat(kitsune2_gossip): Resolve conflicting initiates

### LINKED ISSUE #181: Handle initiate conflicts
When two peers try to initiate with each other, they will currently lock each other out and not proceed. We should add a conflict resolution mechanism for when this happens, so that the agents can proceed with one agent giving up on their initiation so that the other can continue.

ThetaSinner:
Closes #181

### DIFF
--- a/crates/gossip/proto/gossip.proto
+++ b/crates/gossip/proto/gossip.proto
@@ -90,6 +90,11 @@ message K2GossipInitiateMessage {
   // The DHT sectors covered by the union of the agents in the participating_agents list.
   ArcSetMessage arc_set = 11;
 
+  // A tie breaker so that conflicting initiates between a pair of peers can be resolved without a further message.
+  //
+  // This must be a random number, selected per gossip round.
+  uint32 tie_breaker = 12;
+
   // Request ops that are new since the given timestamp.
   int64 new_since = 20;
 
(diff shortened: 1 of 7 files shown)


crates/gossip/proto/gossip.proto
```diff
@@ -90,6 +90,11 @@ message K2GossipInitiateMessage {
   // The DHT sectors covered by the union of the agents in the participating_agents list.
   ArcSetMessage arc_set = 11;
 
+  // A tie breaker so that conflicting initiates between a pair of peers can be resolved without a further message.
+  //
+  // This must be a random number, selected per gossip round.
+  uint32 tie_breaker = 12;
```
neonphog, line 96: The webrtc algorithm does something similar but without having to send an extra bit of data by using the lexical ordering of the peer id. Do you think that would work here?

neonphog (approved):
Consider using the peer id as the tie breaker, but I'm also fine with this work as it's complete already.

crates/gossip/proto/gossip.proto
ThetaSinner: I did consider that. I was thinking that because of the asymmetry in the gossip, it'd be better to have a random choice. A peer can't strategically choose their peer id like they could choose their tie breaker but for honest peers, random is preferable I think

PR #185, opened 2025-04-11
refactor(api): remove config options unused warning

### LINKED ISSUE #184: Fix: log warning about config only when failed to set options
Kitsune outputs logs about configuration potentially not being applied if it contains non-existing options. This warning should not be output if the configuration is correct.

jost-s:
I commented in the issue:

> There's no way to know if the option is valid or not. Not in kitsune, not in Holochain. It would make the most sense to check for type correctness in kitsune and not import all of those types into Holochain and check there.
But the configuration was purposely built type-free and generic, so it would be bigger refactor to address the issue. My proposal is to just eliminate the warning, as it's not an informative warning.

Not an offense! The warning was added with the intention to identify deprecated options being set. But the way the configuration must be used doesn't lend itself to identifying such cases, because the configuration is usually constructed with defaults and then module configurations are updated. That means the warning is always output. It'll probably confuse developers for a while and then they and we will become numb to the warning.

resolves #184

### DIFF
--- a/crates/api/src/builder.rs
+++ b/crates/api/src/builder.rs
@@ -94,8 +94,6 @@ impl Builder {
             gossip.default_config(config)?;
             local_agent_store.default_config(config)?;
             publish.default_config(config)?;
-
-            config.mark_defaults_set();
         }
 
         Ok(self)
--- a/crates/api/src/config.rs
+++ b/crates/api/src/config.rs
@@ -51,7 +51,6 @@ impl Default for ConfigMap {
 
 struct Inner {
     map: ConfigMap,
-    are_defaults_set: bool,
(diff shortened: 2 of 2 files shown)


jost-s:
An alternative would be implement a function that constructs a config based on the set values and fills in the remaining values based on defaults. But with the JSON based config that provides the flexibility of updating config at runtime that seems very complicated.

jost-s:
Our (multiple) test configs don't set all default options and the test may override the option. That should be the reason why this has been showing up. Then it would be justified. I can't remember all the cases, but the logic seems sound. In production the warning doesn't up with a default conductor config or overrides of gossip, so it should all be in order.

PR #187, opened 2025-04-15
feat!: Tools for filtering local agents

ThetaSinner:


crates/api/src/local_agent_store.rs
```diff
@@ -17,6 +17,9 @@ pub trait LocalAgentStore: 'static + Send + Sync + std::fmt::Debug {
 
     /// Get a list of all local agents currently in the store.
     fn get_all(&self) -> BoxFut<'_, K2Result<Vec<DynLocalAgent>>>;
+
+    /// Check whether an agent is local.
+    fn is_agent_local(&self, agent: id::AgentId) -> BoxFut<'_, K2Result<bool>>;
```
ThetaSinner, line 22: I think breaking the API is preferable to always having to `get_all` and look for the agent in a vector when it's internally a hash map

crates/core/src/common.rs
```diff
@@ -0,0 +1,218 @@
...
+    AgentInfoSigned, DhtArc, DynLocalAgentStore, DynPeerStore, K2Result,
+};
+use std::collections::HashSet;
+use std::sync::Arc;
+
+/// Get all remote agents from the peer store.
+///
+/// Uses the [DynPeerStore] to get all agents and the [DynLocalAgentStore] to filter out local
+/// agents.
+pub async fn get_all_remote_agents(
```
ThetaSinner, line 11: This one is used by K2 core, the others are only used by Holochain in my local patched Holochain version

crates/core/src/factories/core_space.rs
```diff
@@ -509,20 +509,11 @@ async fn broadcast_agent_info(
     publish: DynPublish,
     agent_info_signed: Arc<AgentInfoSigned>,
 ) -> K2Result<()> {
-    let all_agents = peer_store.get_all().await?;
```
ThetaSinner, line 512: This is really a refactor, to move this code into the common module

crates/api/src/local_agent_store.rs
jost-s: Totally, break the API as much as you want while the iron is still hot.

jost-s (approved):
Sweet as

crates/core/src/common.rs
```diff
@@ -0,0 +1,218 @@
...
+        .get_all()
+        .await?
+        .into_iter()
+        .map(|a| a.agent().clone())
+        .collect::<HashSet<_>>();
+
+    Ok(agent_list
+        .into_iter()
+        .filter(|a| !all_local_agents.contains(&a.agent))
+        .collect())
```
neonphog, line 61:
```suggestion
    let all_local_agents = local_agent_store
        .get_all()
        .await?
        .into_iter()
        .filter_map(|a| {
            if all_local_agents.contains(&a.agent().agent) {
                None
            } else {
                Some(a.agent.clone())
            }
        })
        .collect::<HashSet<_>>();
```

Rather than creating 2 vecs

crates/core/src/common.rs
```diff
@@ -0,0 +1,218 @@
...
+    peer_store: DynPeerStore,
+    local_agent_store: DynLocalAgentStore,
+    loc: u32,
+    limit: usize,
+) -> K2Result<Vec<Arc<AgentInfoSigned>>> {
+    let near_location = peer_store.get_near_location(loc, limit).await?;
+    remove_local_agents(local_agent_store, near_location).await
+}
+
+async fn remove_local_agents(
```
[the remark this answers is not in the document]
ThetaSinner: Makes sense, removed

PR #189, opened 2025-04-16
feat!: Expose current URL instead of is_local_agent

ThetaSinner:
I think this is a more useful way of doing the check. The local agent we're sending to might have left the space and won't be in the local agent store but we still want to avoid sending to ourselves. 

I started looking at why the historical sync test was flaky. It turns out it was really not doing what it was supposed to do and I found some issues along the way. See the inline comments.

### DIFF
--- a/crates/core/src/factories/mem_op_store.rs
+++ b/crates/core/src/factories/mem_op_store.rs
@@ -102,7 +102,7 @@ impl From<MemoryOp> for Bytes {
 ///
 /// Test data should create [MemoryOp]s and not be aware of this type.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct MemoryOpRecord {
+pub struct MemoryOpRecord {
     /// The id (hash) of the op
     pub op_id: OpId,
     /// The creation timestamp of this op
@@ -140,13 +140,17 @@ impl From<Op> for MemoryOp {
     }
 }
 
+/// The in-memory op store implementation for Kitsune2.
+///
+/// Intended for testing only, because it provides no persistence of op data.
 #[derive(Debug)]
-struct Kitsune2MemoryOpStore {
+pub struct Kitsune2MemoryOpStore {
     _space: SpaceId,
     inner: RwLock<Kitsune2MemoryOpStoreInner>,
 }
 
 impl Kitsune2MemoryOpStore {
+    /// Create a new [Kitsune2MemoryOpStore].
     pub fn new(space: SpaceId) -> Self {
         Self {
             _space: space,
@@ -163,10 +167,13 @@ impl std::ops::Deref for Kitsune2MemoryOpStore {
     }
 }
 
+/// The inner state of a [Kitsune2MemoryOpStore].
 #[derive(Debug, Default)]
-struct Kitsune2MemoryOpStoreInner {
-    op_list: HashMap<OpId, MemoryOpRecord>,
-    time_slice_hashes: TimeSliceHashStore,
+pub struct Kitsune2MemoryOpStoreInner {
+    /// The stored op data.
+    pub op_list: HashMap<OpId, MemoryOpRecord>,
+    /// The time slice hashes.
+    pub time_slice_hashes: TimeSliceHashStore,
 }
 
 impl OpStore for Kitsune2MemoryOpStore {
--- a/crates/gossip/src/harness.rs
+++ b/crates/gossip/src/harness.rs
@@ -215,11 +215,11 @@ impl K2GossipFunctionalTestHarness {
                     tracing::info!("Waiting for op ids to be discovered: {} != {}. Have stored {}/{} and {}/{}", our_op_ids.len(), other_op_ids.len(), our_stored_op_ids.len(), our_op_ids.len(), other_stored_op_ids.len(), other_op_ids.len());
                 }
 
-                tokio::time::sleep(Duration::from_millis(50)).await;
+                tokio::time::sleep(Duration::from_millis(100)).await;
             }
         })
         .await
-        .expect("Timed out waiting for instances to sync");
+        .expect("Timed out waiting for ops to be discovered");
     }
 
     /// Wait for two instances to sync their op stores.
--- a/crates/gossip/src/harness.rs
+++ b/crates/gossip/src/harness.rs
@@ -1,22 +1,24 @@
 //! Functional test harness for K2Gossip.
 
+use crate::harness::op_store::K2GossipMemOpStoreFactory;
 use crate::peer_meta_store::K2PeerMetaStore;
 use crate::{K2GossipConfig, K2GossipFactory, K2GossipModConfig};
-use bytes::Bytes;
 use kitsune2_api::{
-    AgentId, AgentInfoSigned, BoxFut, Builder, Config, DhtArc, DynGossip,
-    DynOpStore, DynSpace, K2Error, K2Result, LocalAgent, MetaOp, OpId, OpStore,
-    OpStoreFactory, SpaceHandler, SpaceId, StoredOp, Timestamp, TxBaseHandler,
-    TxHandler, TxSpaceHandler, UNIX_TIMESTAMP,
+    AgentId, AgentInfoSigned, DhtArc, DynGossip, DynSpace, LocalAgent, OpId,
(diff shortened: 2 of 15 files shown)


crates/core/src/factories/mem_op_store.rs
```diff
@@ -102,7 +102,7 @@ impl From<MemoryOp> for Bytes {
 ///
 /// Test data should create [MemoryOp]s and not be aware of this type.
 #[derive(Debug, Clone, Serialize, Deserialize)]
-struct MemoryOpRecord {
+pub struct MemoryOpRecord {
```
ThetaSinner, line 105: Exposed for downstream testing

crates/gossip/src/harness.rs
```diff
@@ -201,7 +209,10 @@ impl K2GossipFunctionalTestHarness {
                 );
 
                 if our_op_ids == other_op_ids {
```
ThetaSinner, line 211: Some tracing improvements to make it easier to see how the test is progressing.

crates/gossip/src/harness.rs
```diff
@@ -322,13 +333,20 @@ pub struct K2GossipFunctionalTestFactory {
     space_id: SpaceId,
     /// The builder for creating new instances.
     builder: Arc<Builder>,
+    /// A sender for ops stored events.
+    ///
+    /// This must not be used for sending. If it is provided then the K2 mem op store is in use
+    /// and that will send ops through this channel. Use the sender to create new listeners so
+    /// that each space can update its DHT.
+    ops_stored_sender: Option<tokio::sync::broadcast::Sender<Vec<StoredOp>>>,
```
ThetaSinner, line 341: This is new and was a big part of why the test wasn't doing what it was supposed to do. When we store ops, we need to inform the DHT model. This most of the changes in the rest of this file are about making that happen

crates/gossip/src/harness.rs
```diff
@@ -395,10 +426,197 @@ impl K2GossipFunctionalTestFactory {
         let peer_meta_store =
             K2PeerMetaStore::new(space.peer_meta_store().clone());
 
+        if let Some(ref sender) = self.ops_stored_sender {
+            let space = space.clone();
+            let mut receiver = sender.subscribe();
+            tokio::spawn(async move {
```
ThetaSinner, line 432: Relay stored ops to the space's inform method

crates/gossip/src/harness.rs
```diff
@@ -395,10 +426,197 @@ impl K2GossipFunctionalTestFactory {
...
+        op_list: Vec<Bytes>,
+    ) -> BoxFut<'_, K2Result<Vec<OpId>>> {
+        Box::pin(async move {
+            let ops_to_add = op_list
+                .iter()
+                .map(|op| -> serde_json::Result<(OpId, MemoryOpRecord)> {
+                    let mut op = MemoryOpRecord::from(op.clone());
+
+                    // Behaviour difference from the inner implementation.
+                    op.stored_at = op.created_at;
```
ThetaSinner, line 519: I also needed to be able to create historical data without actually fiddling with the internals too much. This seemed like the lightest touch, to pretend that ops turned up at the point they were created. Almost as though they were authored.

It's not true to the real implementation because then other nodes syncing this data will do the same and the "what's new" won't behave as expected. That's why using this op store in the test hardness is optional.

crates/gossip/src/respond/accept.rs
```diff
@@ -130,6 +130,8 @@ impl K2Gossip {
                     DhtSnapshotNextAction::NewSnapshot(snapshot) => {
                         match snapshot {
                             DhtSnapshot::DiscSectors { .. } => {
+                                tracing::info!(?accept.session_id, "Found a disc mismatch, starting to compare sectors");
```
ThetaSinner, line 133: You can normally only see what route gossip sync takes at debug. I'd like to know with less output.

crates/gossip/src/respond/no_diff.rs
```diff
@@ -42,22 +42,26 @@ impl K2Gossip {
...
         no_diff: &K2GossipNoDiffMessage,
     ) -> K2GossipResult<()> {
-        let mut accepted_states = self.accepted_round_states.write().await;
-        if !accepted_states.contains_key(&from_peer) {
-            return Err(K2GossipError::peer_behavior(format!(
-                "Unsolicited NoDiff message from peer: {:?}",
-                from_peer
-            )));
-        }
+        let state = {
```
ThetaSinner, line 45: Okay, into the embarrassing part. When I got the historical sync running. I saw some deadlocks.

They weren't at the same point every time but clearly reproducible. It took me a while to understand why so let me try to explain.

The accepted states is a hashmap behind an rwlock. The inner values of the accepted states are also mutexes. This is so that you can look up an accepted session and work with its state safely without blocking all other accepted states. However, care must be taken because:

- In the responders (the functions that handle acting as a responder), we read the accepted states and clone the inner state. The inner state is then locked and kept locked while the function runs. If the session then finishes because we're sending a final message, we try to take out a write lock on the accepted states to remove this session.
- Here and in the session timeout task, we do things the other way round. We take out a write lock on the accepted states and then try to lock each inner state to check its start time, while the outer lock is still held. 

If you get unlucky (and it clearly doesn't take that much) then you can end up in a state where an responder function is running and has locked a session state but wants to lock the outer state to remove that session. At the same time, other code has decided to claim the write lock that the responder function needs and is waiting on the session state to be free, that the responder function is holding.

I've gone through all the uses of this locking pattern and I can't find any more cases to fix. Even this one I think is very unlikely to cause a problem. The tests stopped showing this problem when I fixed the timeout task, but I'm fixing the issue here too just to be safe.

I know this is the second problem like this I've found and I'd understand if I'm not trusted with locks any more. However, now that the test does what it's supposed to do, we do have a good load running through all of this code in a pretty realistic way... so anything that's likely to happen often should be caught.

crates/gossip/src/timeout.rs
```diff
@@ -86,9 +86,12 @@ async fn remove_timed_out_rounds(
     let mut remove = HashSet::new();
     {
         for (url, state) in accepted_round_states.iter() {
-            let state_lock = state.lock().await;
+            let Ok(state_lock) = state.try_lock() else {
```
ThetaSinner, line 89: This is what I mean about not waiting on both locks. If we can check this session then do. If it's currently in use then clearing it out will have to wait.

crates/gossip/tests/historical_load.rs
```diff
@@ -95,12 +117,10 @@ async fn historical_load() {
         "Completed rounds with agent 2 should have a value"
     );
 
-    let required_rounds =
-        total_size / K2GossipConfig::default().max_gossip_op_bytes as usize + 1;
     assert!(
-        completed_rounds_with_agent_2.unwrap() <= required_rounds as u32,
```
ThetaSinner, line 101: There is a behaviour change here and this is where I should have spent more time on this test in the first place. The disc sync has to complete before the disc top hash matches and the test will move on. That means it's not just a matter of a round or two to discover all the op ids. I'm see the op ids discovered quickly but then we have to fetch all the data (about 40Mb) before we can proceed to syncing the rings. So we can place an upper bound on this but not really an assertion because it depends on compute speed. Locally I'm seeing it complete ~7 rounds. The first couple sync disc ops and then it's waiting on fetch.

jost-s (dismissed):
Looks great to me. Nice digging and adding traces in useful places!

My suggestion regarding the op store doesn't hold up merging this.

crates/gossip/src/respond/no_diff.rs
ThetaSinner: It's certainly not the only place we've had problems with locks but this particular case wouldn't even necessarily be found in a long running test with WT so I'm glad it's caught but it does make me wonder what else might be lurking.

crates/gossip/src/harness.rs
ThetaSinner: That's a fair point, maybe gossip testing needs a more realistic op store... That would be more work to put in place but would actually be cleaner. Yes, let me have a think and then I'll come look at this. I'd like to finish what I'm in the middle of with a Holochain library right now.

jost-s (approved):
Yeah!

PR #193, opened 2025-04-22
chore: Add cargo make

### LINKED ISSUE #188: Add `cargo make` support

ThetaSinner:
Closes #188

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -13,7 +13,7 @@ args = ["clippy", "--all-targets", "--", "--deny=warnings"]
 
 [tasks.format-toml-check]
 description = "Check that all TOML files are formatted correctly"
-install_crate = true
+install_crate = { crate_name = "taplo-cli", version = "0.9.3", binary = "taplo", test_arg = "--help" }
 command = "taplo"
 args = ["format", "--check"]
 
@@ -53,7 +53,7 @@ cwd = "crates/tool_proto_build"
 
 [tasks.format-toml]
 workspace = false
-install_crate = true
+install_crate = { crate_name = "taplo-cli", version = "0.9.3", binary = "taplo", test_arg = "--help" }
 command = "taplo"
 args = ["format"]
 
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -33,8 +33,9 @@ command = "cargo"
 args = ["test", "--lib"]
 
 [tasks.test-integration]
-command = "cargo"
-args = ["test", "--no-run"]
+description = "Run integration tests"
+# This task is a placeholder, it will be overridden by the Makefile.toml in the crate
+workspace = true
 
 #
 # Tasks that modify content and are not run on CI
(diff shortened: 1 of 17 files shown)


Makefile.toml
```diff
@@ -0,0 +1,101 @@
...
+# Tests
+#
+
+[tasks.test-unit]
+command = "cargo"
+args = ["test", "--lib"]
+
+[tasks.test-integration]
+command = "cargo"
+args = ["test", "--no-run"]
```
mattyg, line 37: We don't want to run them?

ThetaSinner: I want to be able to run `cargo make test-integration` from the root but only some projects define integration tests. What this actually does is run `cargo make test-integration` in each project and so each project may override this. I can see if I can slim this down any more so that it doesn't look like this gets run.

mattyg (dismissed):

crates/bootstrap_client/Makefile.toml
```diff
@@ -0,0 +1,5 @@
+extend = "../../Makefile.toml"
+
+[tasks.test-integration]
+command = "cargo"
+args = ["test", "--test", "connect"]
```
neonphog, line 5: Is there any way to do this without having to enumerate the specific integration tests? I'm not sure I understand why the unit and integration tests are separate... it seems fine to just run them all at the same time.

Makefile.toml
```diff
@@ -0,0 +1,98 @@
...
+
+[tasks.proto]
+workspace = false
+command = "cargo"
+args = ["run", "--package", "tool_proto_build"]
+cwd = "crates/tool_proto_build"
+
+[tasks.format-toml]
+workspace = false
+install_crate = { crate_name = "taplo-cli", version = "0.9.3", binary = "taplo", test_arg = "--help" }
```
neonphog, line 57: Is there any way to make a dependency that installs taplo-cli so we don't have to ensure the version is the same in two locations?

neonphog (dismissed):
Whew that first run after a target delete is a doozy. Between the docs, clippy, and tests in each and every crate individually, it must build datachannel-sys dozens of times. And it now is 9.2G : )

Had a couple questions, but looks good as a first cut.

jost-s (dismissed):

crates/bootstrap_client/Makefile.toml
ThetaSinner: Not until cargo adds an option to do that. I separated them because that's how I tend to run them but I could just have `cargo make` run all of them and if I want to run them differently then I will run `cargo test --lib` myself?

Makefile.toml
ThetaSinner: Yes https://github.com/holochain/kitsune2/pull/193/commits/bfac650a47793a6c3565597636c0609c664981e6

crates/bootstrap_srv/Makefile.toml
```diff
@@ -0,0 +1,5 @@
+extend = "../../Makefile.toml"
+
+[tasks.test-integration]
```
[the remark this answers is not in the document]
ThetaSinner: As new tests are added? Yes

crates/bootstrap_client/Makefile.toml
jost-s: That would work for me as well. I usually only run all tests at the end.

jost-s (approved):

PR #194, opened 2025-04-23
feat: Add schema support for configuration

ThetaSinner:
Part of https://github.com/holochain/holochain/issues/4894

A lot of the content here is the example. I've added that to CI in a bit of an untidy way with a view to merging that with the `cargo make` PR.

### DIFF
--- a/crates/transport_tx5/src/lib.rs
+++ b/crates/transport_tx5/src/lib.rs
@@ -43,7 +43,7 @@ pub mod config {
     /// Configuration parameters for [Tx5TransportFactory](super::Tx5TransportFactory).
     #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
     #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
-    #[serde(rename_all = "camelCase", deny_unknown_fields)]
+    #[serde(rename_all = "camelCase")]
     pub struct Tx5TransportConfig {
         /// Allow connecting to plaintext (ws) signal server
         /// instead of the default requiring TLS (wss).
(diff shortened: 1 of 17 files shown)


crates/transport_tx5/src/lib.rs
```diff
@@ -76,11 +87,75 @@ pub mod config {
...
+            ice_transport_policy: Option<IceTransportPolicy>,
+            peer_identity: Option<String>,
+            rtcp_mux_policy: Option<RtcpMuxPolicy>,
+        }
+
+        let root = schemars::schema_for!(WebRtcConfig);
+        for def in root.definitions {
+            generator.definitions_mut().insert(def.0, def.1);
+        }
+        schemars::schema::Schema::Object(root.schema)
```
ThetaSinner, line 157: This is how I plan to integrate this into Holochain. Define a schema that mentions our known configuration items and permits unknown ones (see the example) and then just merges it into the parent schema.

mattyg (commented):
datachannel-rs does have a typed config struct: https://github.com/lerouxrgd/datachannel-rs/blob/master/src/config.rs#L9C12-L9C21

Does it make more sense just to propegate that up to the conductor config? Or if not, make our config type that we can convert into datachannel-rs'?

ThetaSinner:
Eugh grim and they haven't followed the same structure, which go-pion has helpfully stayed close to... https://github.com/pion/webrtc/blob/2de1ac42857c953947dffae9b23f71e114279e61/configuration.go#L14

Looking at tx5, the complete configuration of the underlying library isn't actually exposed either, even though it "claims" to be by being JSON. You can only pass ICE servers to datachannel.

Yeah let me head down to tx5 and see what I can do there.

ThetaSinner:
See https://github.com/holochain/tx5/pull/137

neonphog (approved):
Woot. Nice :+1:

Issue #131, opened 2025-02-26
Bootstrap: Avoid printing in library code


ThetaSinner:
When the bootstrap server is embedded as a library, it still prints out the address it started on -> https://github.com/holochain/kitsune2/blob/main/crates/bootstrap_srv/src/server.rs#L82

This logic should move to the bootstrap server binary so that projects that are embedding the bootstrap server can choose how they want to output this information.

anchalshivank:
@ThetaSinner https://github.com/holochain/kitsune2/pull/199

PR #199, opened 2025-04-29
refactor(bootstrap): move server address logging to binary

anchalshivank:
#131 
When the bootstrap server is embedded as a library, it currently prints the server address on startup, which is not desirable for library users.


ThetaSinner (commented):
Hi @anchalshivank, thank you for the PR!

So that we get the old behavior in the binary, could this new function be called once the server has started and before it starts waiting for a Ctrl+C signal? I think that should be possible here -> https://github.com/holochain/kitsune2/blob/main/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs#L159

ThetaSinner (approved):

PR #201, opened 2025-04-29
chore: Tidy dependencies

ThetaSinner:


crates/kitsune2/Cargo.toml
```diff
@@ -20,14 +20,13 @@ kitsune2_api = { workspace = true }
 kitsune2_core = { workspace = true }
 kitsune2_gossip = { workspace = true }
 kitsune2_transport_tx5 = { workspace = true }
-serde = { version = "1.0.219", features = ["derive"] }
+serde = { workspace = true, features = ["derive"] }
```
ThetaSinner, line 23: My IDE had "helpfully" added this, but we don't want to force the latest version of serde without a reason

jost-s: AU - artificial unintelligence

jost-s (approved):

crates/kitsune2/Cargo.toml
ThetaSinner: Haha! This was more of an autocomplete thing, like typing `derive(Seria...` triggered importing the library without me realizing. Could be an AI feature but I'm not aware that it is.

Issue #157, opened 2025-03-25
Bandwidth limiting by space


ThetaSinner:
This was sort of a feature in K1 although it was inconsistently enforced. We want to be able to both report and limit bandwidth usage by space. Ideally this is done automatically for a space and doesn't require each module to opt in.

Nice to have: Report usage by module?

ThetaSinner:
Initially, we'd like to track bandwidth by space and report it. Once we have that, we can look at adding limiting logic.

I think we can capture bytes in/out at the space level [here](https://github.com/holochain/kitsune2/blob/main/crates/core/src/factories/core_space.rs#L199) and [here](https://github.com/holochain/kitsune2/blob/main/crates/core/src/factories/core_space.rs#L463). A space roughly translates to an app or component of an app, so that is the useful level to capture usage at.

I'm not certain that these are the ideal points and capture all messages in and out for the space. I'm leaving that to be investigated as part of this issue, but they should be good starting points. It might be necessary to do this in the transport module (`kitsune2_transport_tx5`) where we route messages to the space, rather than doing this in the space directly.

Once we have the usage recorded, it can be exposed from the space similarly to what we've done on the [gossip](https://github.com/holochain/kitsune2/blob/main/crates/api/src/gossip.rs#L31), [fetch](https://github.com/holochain/kitsune2/blob/main/crates/api/src/fetch.rs#L100) and [transport](https://github.com/holochain/kitsune2/blob/main/crates/api/src/transport.rs#L161) APIs.

As an open question, can we attribute bandwidth usage to individual modules?

ThetaSinner:
Looking at this a bit closer in the integration that uses this code, I think it does have to be in the transport that we track this. The space only handles p2p agent messaging, not the traffic that K2 itself needs to run.

anchalshivank:
@ThetaSinner I found that we are sending/receiving data through any struct that implements this [trait](https://github.com/holochain/kitsune2/blob/33c0b10cc761a8471e49b5a94faf4ba0a18e01e6/crates/api/src/transport.rs#L141)

Now there are two ways to solve this:
1. We could edit all the implementations and keep track of how many bytes have been sent by just incrementing a counter
2. Or create a wrapper like:
```rust
pub struct BandwidthTracker<T: TxImp> {
   inner: T,
   bytes_sent: Arc<AtomicU64>,
   bytes_received: Arc<AtomicU64> 
}
```

Which approach do you prefer?

anchalshivank:
[...]

PR #205, opened 2025-05-03
feat(transport): add bandwidth tracking (send only) by space (#157)

anchalshivank:
Implemented tracking of bytes sent per space in the transport layer by wrapping implementations of the Transport trait. This enables basic observability of outbound traffic for each space.

Currently, only sent data is tracked; receiving metrics can be added in a follow-up.

Related to: #157


crates/api/src/transport.rs
```diff
@@ -222,6 +231,8 @@ pub trait Transport: 'static + Send + Sync + std::fmt::Debug {
 
     /// Dump network stats.
     fn dump_network_stats(&self) -> BoxFut<'_, K2Result<TransportStats>>;
+    /// Get Bandwidth Tracker
```
ThetaSinner, line 234:
```suggestion

    /// Get Bandwidth Tracker
```

ThetaSinner: Just adding a newline

crates/api/src/transport.rs
```diff
@@ -511,3 +547,172 @@ pub struct TransportConnectionStats {
...
+            module_stats: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    /// Records data sent for a given space and optional module.
+    ///
+    /// # Arguments
+    ///
+    /// * `space` - The unique identifier for the space.
+    /// * `module` - Optional module identifier.
```
ThetaSinner, line 625: Maybe mention that it's required when tracking module messages but allowed to be absent when sending messages at the space level, not via a module?

Not sure how to write that with fewer works :)

crates/core/src/factories/core_space.rs
```diff
@@ -284,6 +284,9 @@ impl CoreSpace {
             }
         }
     }
+    pub fn get_bandwidth_tracker(&self) -> Arc<BandwidthTracker> {
```
ThetaSinner, line 287:
```suggestion

    pub fn get_bandwidth_tracker(&self) -> Arc<BandwidthTracker> {
```

ThetaSinner: Another newline

ThetaSinner (commented):
This looks great! I really appreciate the quality documentation on the new types. We've made an effort to try to keep up with documentation on this project and it's great if we can keep that going :)

A couple of really minor comments. It'd also be good to take the `.idea` files back out of Git please. I'm a RustRover user myself, but we try to keep IDE files out of Git because we all use different tools on this project :)

crates/api/src/transport.rs
anchalshivank: Fixed

crates/api/src/transport.rs
```diff
@@ -189,7 +197,6 @@ pub trait Transport: 'static + Send + Sync + std::fmt::Debug {
         module: String,
         handler: DynTxModuleHandler,
     );
-
```
jost-s, line 192: Please keep this empty line between functions.

crates/api/src/transport.rs
```diff
@@ -511,3 +546,178 @@ pub struct TransportConnectionStats {
...
+        space: SpaceId,
+        module: Option<String>,
+        bytes: u64,
+    ) {
+        if let Ok(mut space_stats) = self.space_stats.write() {
+            space_stats
+                .entry(space.clone())
+                .or_insert_with(BandwidthStats::new)
+                .add_sent(bytes);
+        }
```
jost-s, line 637: I suggest to unwrap the write lock:
```suggestion
        self.space_stats.write()
            .unwrap()
            .entry(space.clone())
            .or_insert_with(BandwidthStats::new)
            .add_sent(bytes);
```
That's what we've done throughout this repository with sync Mutexes and RwLocks and what the [examples](https://doc.rust-lang.org/std/sync/struct.RwLock.html#examples) do too.

crates/api/src/transport.rs
```diff
@@ -511,3 +546,178 @@ pub struct TransportConnectionStats {
...
+    /// * `space` - The unique identifier for the space.
+    /// * `module` - Optional module identifier. Use `Some(module_id)` for module-specific messages, or `None` for space-level messages.
+    /// * `bytes` - Number of bytes sent.
+    pub fn track_received(
+        &self,
+        space: SpaceId,
+        module: Option<String>,
+        bytes: u64,
+    ) {
+        if let Ok(mut space_stats) = self.space_stats.write() {
```
jost-s, line 662: Same here

jost-s (changes requested):
The code reads very clear. Nevertheless please add a test for this feature. If you want guidance how to do that, give us a shout. There are unit tests for most features in this repo, so you can have a look at them for inspiration.

.idea/workspace.xml
cdunster, line 1: Should this file be committed? Is there something non-standard in this config that might help others who are using IntelliJ?

crates/api/src/transport.rs
```diff
@@ -222,6 +229,9 @@ pub trait Transport: 'static + Send + Sync + std::fmt::Debug {
 
     /// Dump network stats.
     fn dump_network_stats(&self) -> BoxFut<'_, K2Result<TransportStats>>;
+
+    /// Get Bandwidth Tracker
```
cdunster, line 233: Is it possible to elaborate a bit on this comment? Maybe something like "Get a reference to the struct that holds the stats of bandwidth usage of..."

crates/api/src/transport.rs
```diff
@@ -255,9 +266,15 @@ impl DefaultTransport {
             imp,
             space_map: hnd.space_map.clone(),
             mod_map: hnd.mod_map.clone(),
+            tracker: hnd.tracker.clone(),
         });
         out
     }
+
+    /// Get Bandwidth tracker
```
cdunster, line 274: Same here.

crates/api/src/transport.rs
```diff
@@ -321,14 +338,21 @@ impl Transport for DefaultTransport {
...
             })
             .encode()?;
-            self.imp.send(peer, enc).await
+            match self.imp.send(peer, enc).await {
+                Ok(()) => {
+                    self.tracker.track_sent(space, None, len);
+                    K2Result::Ok(())
+                }
+                Err(err) => K2Result::Err(err),
+            }
```
cdunster, line 355: I think that you could use the `?` operator here:

```suggestion
            self.imp.send(peer, enc).await?;
            self.tracker.track_sent(space, None, len);
            K2Result::Ok(())
```

Or even, a more functional approach using [inspect](https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect):

```suggestion
            self.imp
                .send(peer, enc)
                .await
                .inspect(|_| self.tracker.track_sent(space, None, len))
```

This is just a preference/opinion though, so it's up to you :slightly_smiling_face:

crates/api/src/transport.rs
```diff
@@ -340,20 +364,31 @@ impl Transport for DefaultTransport {
...
             })
             .encode()?;
-            self.imp.send(peer, enc).await
+            match self.imp.send(peer, enc).await {
+                Ok(()) => {
+                    self.tracker.track_sent(space, Some(module), len);
+                    K2Result::Ok(())
+                }
+                Err(err) => K2Result::Err(err),
+            }
```
cdunster, line 381: Same here as above

cdunster (commented):
Looking good. Thanks!

I particularly like the detailed and clear comments. I added a couple of suggestions.

crates/api/src/transport.rs
anchalshivank: Yups I will elaborate

anchalshivank:
I haven't added tests for this module because the behavior is often non-deterministic. For example, if I wait for 5 seconds, the node may gossip during that time, leading to inconsistent test results. See the relevant code here: https://github.com/anchalshivank/kitsune2/blob/bc941a0f3f15b42c478fa052ca11cf4484c17b2e/crates/kitsune2/src/lib.rs#L263. Suggestions for handling this kind of testing would be appreciated.

jost-s:
> I haven't added tests for this module because the behavior is often non-deterministic. For example, if I wait for 5 seconds, the node may gossip during that time, leading to inconsistent test results. See the relevant code here: https://github.com/anchalshivank/kitsune2/blob/bc941a0f3f15b42c478fa052ca11cf4484c17b2e/crates/kitsune2/src/lib.rs#L263. Suggestions for handling this kind of testing would be appreciated.

Sure I'll make suggestions. This kind of thing is inherent to this system, where you can't guarantee that something has happened. Our go to tool for this circumstance is to set a timeout and inside of the timeout repeatedly query what should be asserted until it succeeds. If time runs out, the timeout must be longer, until it is stable on CI. Or something can be improved in the test logic. Or it can also happen that a field must be exposed just for testing purposes.

There's a macro for this kind of repeated check https://github.com/holochain/kitsune2/blob/main/crates/test_utils/src/lib.rs#L42

Here's an example of this macro in use https://github.com/holochain/kitsune2/blob/main/crates/core/src/factories/core_fetch/test/incoming_request_queue.rs#L134

Instead of the sleeps (e. g. https://github.com/anchalshivank/kitsune2/blob/bc941a0f3f15b42c478fa052ca11cf4484c17b2e/crates/kitsune2/src/lib.rs#L316), use the `iter_check` macro, and that should make the test predicable.

ThetaSinner:
I like Jost's suggestion for this, and to prove it's working, it's not necessary to assert a specific value I think. An `iter_check!` that asserts "sent > 500" and "received > 500" does demonstrate that the value is increasing as expected when nodes talk to each other.

I've chosen the value 500 at random there, not necessarily useful. It's just about something that is "> 0" really.

anchalshivank:
Made minor changes to some of the previous test cases that were failing due to timeout issues. The application code remains unchanged. I investigated the failures but couldn't identify the exact cause, so I increased the timeout duration in the tests as a workaround.

anchalshivank:
@ThetaSinner  @jost-s  @cdunster  I’ve added a code snippet for the Bandwidth module. I believe the way I’ve integrated it into the existing code is appropriate. Once everyone agrees on this approach, I’ll proceed with implementing the logic for configuring bandwidth.

Now that the transport layer includes both the tracker and config parameters, monitoring and control can be handled more effectively at this level.

Initially, I considered integrating the bandwidth parameters into the k2GossipModule, but I decided it would be cleaner to create a separate struct instead, without modifying the existing code.

crates/gossip/src/harness.rs
```diff
@@ -118,7 +118,7 @@ impl K2GossipFunctionalTestHarness {
 
     /// Wait for the given ops to be in our op store.
     pub async fn wait_for_ops(&self, op_ids: Vec<OpId>) -> Vec<MemoryOp> {
-        tokio::time::timeout(Duration::from_millis(500), {
+        tokio::time::timeout(Duration::from_millis(5000), {
```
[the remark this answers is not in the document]
anchalshivank: The test cases for the 500 status code were failing, so I increased the timeout. However, I'm unable to debug the root cause, as I haven't made any changes to the relevant part of the code.

crates/kitsune2/src/lib.rs
```diff
@@ -182,7 +181,8 @@ mod test {
 
     #[tokio::test]
     async fn two_node_gossip() {
-        enable_tracing();
+        //If I comment it test passes else fails
+        // enable_tracing();
```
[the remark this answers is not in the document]
anchalshivank: yups

crates/api/src/transport.rs
```diff
@@ -101,7 +106,8 @@ impl TxImpHnd {
             data,
             ..
         } = data;
-
+        let len = data.len() as u64;
+        println!("Bandwidth config {:?}", self.config);
```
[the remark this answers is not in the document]
anchalshivank: I added them for debugging purpose. As this is not the final code

jost-s:
Is this ready for review again, @anchalshivank?

anchalshivank:
> Is this ready for review again, @anchalshivank?

Nope, I will make some changes by the end of the day.

anchalshivank:
> @anchalshivank Are you willing to finish this PR?

Yes I will finish it by tomorrow, I have been very busy these days

jost-s:
@anchalshivank I see you haven't finished this yet. This is just an observation, you don't need to explain yourself or make a promise. I'm sure there is always enough to do for you. =)

Are you okay with us completing this PR?

anchalshivank:
I need more time . If anyone else wants to complete this pr,  you can assign it to them.

ThetaSinner:
No activity in a long time, closing

PR #206, opened 2025-05-06
chore: Fix feature logic for automock

ThetaSinner:
Took me a while to realize why my IDE wasn't letting me navigate around the types with this annotation. We have different logic for when the import is available vs. the attribute applied. That is actually wrong, running the tests without that feature would be broken and it's fair enough that my IDE complains it can't find the import.


jost-s (approved):
Rightly so

PR #208, opened 2025-05-07
feat: Close spaces

ThetaSinner:
Depends on https://github.com/holochain/tx5/pull/142

Implements shutting down a space when it is no longer required:
- The space is removed from the `Kitsune` handler on request.
- The space handler and its module handlers are unregistered from the transport.
- Any peer connections that are only used by this space are closed.
- The shutdown of a space that has active local agents is not permitted. This helps avoid mistakes with shutting down spaces without stopping using them first, but it also gives the space a chance to broadcast a tombstone for the agent before shutting down.

I've added an integration test because it's meaningless to test this with the memory transport.


ThetaSinner:
There is a test `peer_connect_disconnect` that we have ignored for disconnect being broken in tx5. That doesn't actually appear to be the case. I think that test isn't preventing connections being re-established. Given that tx5 disconnect is required for this new test to work, it should be possible to revisit and fix that ignored test.

ThetaSinner:
Can't update versions here until https://github.com/holochain/kitsune2/pull/198 is in because that's bumping tx5 with other changes.

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,99 @@ impl Kitsune for CoreKitsune {
...
+                        ));
+                    }
+
+                    // Checks passed, remove our reference to the space.
+                    self.map.lock().unwrap().remove(&space);
+
+                    // Unregister the space and its module handlers from the transport.
+                    self.transport().await?.unregister_space(space).await;
+
+                    // Close any connections that were only used by this space.
```
cdunster, line 193: I think that this comment is a bit misleading because this code block only gets the connections and then they are removed/closed in the next code block.

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,99 @@ impl Kitsune for CoreKitsune {
...
+                                            }
+                                        }
+                                    }
+                                }
+
+                                Ok(())
+                            })
+                        },
+                    ))
+                    .await;
```
cdunster, line 239: To see if I understand: So the code above gets a list of peer URLs that are connected to this space, then this code goes through all of the spaces and removes from the list of peer URLs, any URL that is also used by another space, then the code below goes through the list that now only contains URLs that are only used by this space and disconnects from them? Is that correct?

Is there a reason not to just do the disconnect when looping through all of the spaces? So line 225 becomes:

```rust
if check_our_peers.contains(url) {
    transport
        .disconnect(
            url.clone(),
            Some("Space is being removed".to_string()),
        )
    .await;
}
```

Maybe because of the borrow checker? I've not tried it but just wondering.

crates/core/src/factories/mem_transport.rs
```diff
@@ -130,6 +130,12 @@ impl TxImp for MemTransport {
         })
     }
 
+    fn get_connected_peers(&self) -> BoxFut<'_, K2Result<Vec<Url>>> {
+        // The memory transport is always connected to everyone but doesn't
+        // expose who is connected here.
+        Box::pin(async move { Ok(vec![]) })
```
cdunster, line 136: Is the mem transport just used for testing? This seems like an invalid call error.

crates/transport_tx5/src/lib.rs
```diff
@@ -320,6 +320,17 @@ impl TxImp for Tx5Transport {
         })
     }
 
+    fn get_connected_peers(&self) -> BoxFut<'_, K2Result<Vec<Url>>> {
+        Box::pin(async {
+            Ok(self
+                .ep
+                .get_connected_peer_addresses()
+                .into_iter()
+                .filter_map(|u| Url::from_str(u.as_ref()).ok())
```
cdunster, line 329: We don't need ownership here, so I think the move done by `into_iter` is unnecessary. This should be able to become:

```suggestion
                .iter()
                .filter_map(|u| Url::from_str(u).ok())
```

mattyg (dismissed):

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,99 @@ impl Kitsune for CoreKitsune {
...
+            let fut = self.map.lock().unwrap().get(&space).cloned();
+            if let Some(fut) = fut {
+                if let Ok(s) = fut.await {
+                    // Just a point in time check, try to help the user out if they accidentally
+                    // try to remove a space with active local agents.
+                    // Those agents need to leave the space before it should be removed.
+                    if !s.local_agent_store().get_all().await?.is_empty() {
+                        return Err(K2Error::other(
+                            "Cannot remove space with local agents",
+                        ));
```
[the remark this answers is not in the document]
ThetaSinner: My thinking is that you wouldn't react to this error. It will prevent shutting down the space if there are agents present but that probably can't be fixed. If you're doing a deliberate shutdown then you do an agent leave and check whether any other agents are using the space. If not, you can shut it down. If an agent joins the space while you're doing those checks, then you shouldn't shut the space down and an error is acceptable.

Should it be an error that we can ignore? Possibly, and maybe that's a good enough reason to add a custom error?

crates/core/src/factories/core_kitsune.rs
ThetaSinner: I agree, this comment was written for an earlier version of the code

crates/core/src/factories/core_kitsune.rs
ThetaSinner: > above gets a list of peer URLs that are connected to this space

To this K2 instance. Connections are not made per space. The "space network" is a logical network.

>  Is that correct?

Yes

> Is there a reason not to just do the disconnect when looping through all of the spaces?

I'm not following, I think. We don't know that the peer connection was only used by this space until we have been through the complete list of other spaces

crates/core/src/factories/mem_transport.rs
ThetaSinner: Yes, it's only useful for testing. It is exposed outside of this repository but in Holochain we're mostly doing tests that use the real tx5 network transport. 

I could return an error, but then shutting down spaces will fail if done in K2 tests. Maybe it should?

crates/transport_tx5/src/lib.rs
ThetaSinner: Good spot, yes

crates/core/src/factories/core_kitsune.rs
ThetaSinner: https://github.com/holochain/kitsune2/pull/208/commits/dff8adfd0fb3d8f7dae77935a353b8669b041c64

crates/core/src/factories/core_kitsune.rs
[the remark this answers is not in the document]
neonphog: I'd be inclined to not worry about it until/unless there's a pull for actually branching on this specific error type.

crates/core/src/factories/core_space.rs
```diff
@@ -450,7 +450,22 @@ impl Space for CoreSpace {
                 }
 
                 // also send the tombstone to the bootstrap server
-                self.bootstrap.put(info);
+                self.bootstrap.put(info.clone());
+
+                // and send it to our peers.
+                if let Err(err) = broadcast_agent_info(
```
neonphog, line 456: Is this really what we want here? It's getting *all* agents from the peer_store, and putting this in the publish queue. At least for the last agent, if the space gets shut down right after, the publish module won't have a chance to actually publish it.

neonphog (approved):
Question about publishing agent info on leave. Otherwise LGTM.

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,101 @@ impl Kitsune for CoreKitsune {
...
+                        .lock()
+                        .unwrap()
+                        .iter()
+                        .map(|s| s.1.clone())
+                        .collect::<Vec<_>>();
+                    futures::future::join_all(spaces.into_iter().map(
+                        |fut| -> BoxFut<'_, K2Result<()>> {
+                            let connected_peer_urls =
+                                connected_peer_urls.clone();
+                            Box::pin(async move {
```
jost-s, line 215: Can this just be an `async` block instead of a boxed future?

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,101 @@ impl Kitsune for CoreKitsune {
...
+                                if let Ok(s) = fut.await {
+                                    let check_our_peers = connected_peer_urls
+                                        .read()
+                                        .unwrap()
+                                        .clone();
+                                    let space_agent_infos =
+                                        s.peer_store().get_all().await?;
+                                    for info in space_agent_infos {
+                                        if let Some(url) = &info.url {
+                                            if check_our_peers.contains(url) {
```
jost-s, line 225: Is it better to filter space_agent_infos and then remove all remaining urls under one write lock?

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,101 @@ impl Kitsune for CoreKitsune {
...
+                                    }
+                                }
+
+                                Ok(())
+                            })
+                        },
+                    ))
+                    .await;
+
+                    // Disconnect from any remaining peers. I.e. those that aren't in use by
```
jost-s, line 241: Nice!

jost-s (commented):
Looks good! Some questions

crates/core/src/factories/core_kitsune.rs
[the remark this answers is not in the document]
cdunster: Okay, that makes sense to me. Just wanted to understand the reasoning but basically we only create unique errors if we expect the user to handle or care about that particular error case.

crates/core/src/factories/core_kitsune.rs
cdunster: > I'm not following, I think. We don't know that the peer connection was only used by this space until we have been through the complete list of other spaces

Ah, yes, I misread it but with fresh eyes it makes sense now. Thanks

crates/core/src/factories/mem_transport.rs
cdunster: It could always return an unimplemented error or something, and then when we need to test shutting down spaces we can implement it as needed.

crates/core/src/factories/core_space.rs
ThetaSinner: No, not really. It would be better to:
- Only select peers we have an open connection with, we don't want to be opening new connections at this point
- Do this send immediately, with no channel, so that when this method completes there is no more work to do and it's safe to shut down the space.

I can work on that, but this is one of several places where we're going to wide and if I'm going to start solving problems like that then I think we want to start adding tests for it and for them to pass, we need to fix other places that cause us to connect to everyone. E.g. gossip, and the other use of this `broadcast_agent_info` method, and probably others.

Or I can do this change as a best effort to get it right so it's one less thing to tune later?

crates/core/src/factories/core_kitsune.rs
ThetaSinner: I don't *think* so. The return type `impl Future` isn't allowed by Rust and I can't make an `async` closure. I could define an actual async function to do this work and that might work but even then, I think Rust complains about the futures having different types when you try to put them in a collection and you end up needing to Box anyway

crates/core/src/factories/core_kitsune.rs
ThetaSinner: I don't know how I'd do that. The lock is local to this function and only created so that Rust is happy that we're not modifying the `HashSet` from multiple threads while these futures execute in parallel

crates/core/src/factories/core_kitsune.rs
```diff
@@ -170,6 +170,101 @@ impl Kitsune for CoreKitsune {
...
+                                    for info in space_agent_infos {
+                                        if let Some(url) = &info.url {
+                                            if check_our_peers.contains(url) {
+                                                connected_peer_urls
+                                                    .write()
+                                                    .unwrap()
+                                                    .remove(url);
+                                            }
+                                        }
+                                    }
```
jost-s, line 232: I mean like this for example, not for all spaces at once.
```suggestion
                                    let urls_to_remove = space_agent_infos
                                        .into_iter()
                                        .filter_map(|a| {
                                            if let Some(url) = &a.url {
                                                if check_our_peers.contains(url)
                                                {
                                                    Some(url.clone())
                                                } else {
                                                    None
                                                }
                                            } else {
                                                None
                                            }
                                        })
                                        .collect::<Vec<_>>();
                                    let mut lock =
                                        connected_peer_urls.write().unwrap();
                                    for url in urls_to_remove {
                                        lock.remove(&url);
                                    }
```

crates/kitsune2/tests/integration.rs
```diff
@@ -0,0 +1,350 @@
...
+                ops_2.len()
+            );
+            println!(
+                "space 2 actual ops received {}/expected {}",
+                actual_ops_2.len(),
+                ops_1.len()
+            );
+        }
+    });
+
```
jost-s, line 289: My comment didn't come through here yesterday. I suggest to remove the space here and assert it fails, because there are still local agents connected.

crates/core/src/factories/core_kitsune.rs
ThetaSinner: Oh I get you now! I'd misread the line number and was looking at the wrong lock. Yes, that would be better

ThetaSinner: https://github.com/holochain/kitsune2/pull/208/commits/d193030e8021ad9bafbcf05af0a4127ba055ecb1

crates/core/src/factories/mem_transport.rs
ThetaSinner: https://github.com/holochain/kitsune2/pull/208/commits/d193030e8021ad9bafbcf05af0a4127ba055ecb1

ThetaSinner: Makes sense, doesn't break anything, done :)

crates/kitsune2/tests/integration.rs
ThetaSinner: Good call, done https://github.com/holochain/kitsune2/pull/208/commits/7652800ce339ca77a0b7a127bafbeedf021ad092

jost-s (approved):
Sweet as!

cdunster (approved):

crates/core/src/factories/core_space.rs
neonphog: I'm cool merging as-is. It's just another item that will need to be addressed before we can scale past a handful of peers.

ThetaSinner: Agreed, yes it is. I'm quite tempted to start tackling that problem before Holochain is ready for it... but it will only really work in Holochain when `holochain_p2p` respects it too and it will be difficult to prevent applications breaking that logic.

PR #211, opened 2025-05-08
build: Simpler test target

ThetaSinner:
Make the test targets simpler to make them more maintainable.


neonphog (approved):
:raised_hands:

Makefile.toml
```diff
@@ -29,14 +29,9 @@ install_crate = false
...
-[tasks.test-unit]
+[tasks.test]
 command = "cargo"
-args = ["test", "--lib"]
-
-[tasks.test-integration]
-description = "Run integration tests"
-# This task is a placeholder, it will be overridden by the Makefile.toml in the crate
-workspace = true
+args = ["test"]
```
cdunster, line 34: I think that this is basically the default from `cargo-make` so you could remove this now: https://github.com/sagiegurari/cargo-make/blob/eacca6b741bbeb91372c6116fc600ba3fe9b8ce4/src/lib/descriptor/makefiles/rust.toml#L415-L425

neonphog (dismissed):

crates/kitsune2_showcase/Makefile.toml
```diff
@@ -1,9 +1,5 @@
 extend = "../../Makefile.toml"
 
-[tasks.test-unit]
+[tasks.test]
 clear = true
 script = "echo No unit tests"
```
cdunster, line 5: Does this need to be overridden? Won't it just report `0 tests run`?

Makefile.toml
ThetaSinner: That works in a single crate workspace but for some reason it doesn't seem to work when you have a workspace setup like this. I might have done something wrong, but if I remove this top level task it complains that the task doesn't exist

crates/kitsune2_showcase/Makefile.toml
ThetaSinner: Good point, removed

jost-s (approved):

Makefile.toml
cdunster: Ah, I see that `skip_core_tasks` is set to true, so there are no default tasks included. Then this is good, and I think it's clearer and more explicit to not include the defaults so ignore me :slightly_smiling_face:

crates/core/Makefile.toml
```diff
@@ -1,5 +1 @@
 extend = "../../Makefile.toml"
```
cdunster, line 1: This is probably my biggest pet peeve about `cargo-make`. There is the `CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE` env setting so that you don't need to do this, see [this section](https://sagiegurari.github.io/cargo-make/#automatically-extend-workspace-makefile). But that only works when running in the root directory and not when we run in the `crates/core` directory, for example. It might still be worth enabling `CARGO_MAKE_EXTEND_WORKSPACE_MAKEFILE` just in case we forget to add a `Makefile.toml` to a new crate or something. /rant

cdunster (approved):

crates/core/Makefile.toml
ThetaSinner: I agree this is untidy when you don't actually need to override any tasks, and yes I can set that https://github.com/holochain/kitsune2/pull/211/commits/24d72f1e6b1e6c163b67569535d82c71023b8267

Makefile.toml
ThetaSinner: Ahh, of course, that'd be why. Yeah I turned some stuff off otherwise the startup time is atrocious :)

PR #212, opened 2025-05-08
bootstrap-auth-test docker build

neonphog:


docker/kitsune2_bootstrap_srv_auth_test/start-cmd.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+test-auth-hook-server 8080 &
```
ThetaSinner, line 3: Multiple processes in the entry point of a container have caused me many headaches :/ 

This is backgrounding the hook server so that if it does the container won't tell us it's unhealthy? 

I'm wondering if this shouldn't be a separate container. We can run two containers on the same host if we don't want to run two servers. But that way we can look at logs separately and spot failure much more easily.

docker/kitsune2_bootstrap_srv_auth_test/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# Pick a directory to work in
+WORKDIR /usr/src/kitsune2
+
+# Copy the workspace's source code into the container
+ADD ["Cargo.toml", "Cargo.lock", "./"]
+ADD crates ./crates
+
+# Build the bootstrap server
+RUN cargo install --path ./crates/bootstrap_srv --example test-auth-hook-server
+RUN cargo install --path ./crates/bootstrap_srv --bin kitsune2-bootstrap-srv
```
ThetaSinner, line 12: Because of the comment below, I think this image should just contain the auth server and the boostrap server image can be deployed alongside

docker/kitsune2_bootstrap_srv_auth_test/start-cmd.sh
neonphog: Normally I'd agree... but the code for the test hook server is the epitome of trivial. It literally either says unauthorized or sends back random bytes. I'm not sure it *can* fail.

That being said, I'm fine to go with your suggestion if you want to move the complexity to the VM install side.

neonphog: Done: https://github.com/holochain/kitsune2/pull/212/commits/a2b1a1966dcc9688893074dae0546792cedd6cbc

docker/kitsune2_bootstrap_srv_auth_test/Dockerfile
neonphog: Done: https://github.com/holochain/kitsune2/pull/212/commits/a2b1a1966dcc9688893074dae0546792cedd6cbc

docker/kitsune2_bootstrap_srv_auth_test/start-cmd.sh
ThetaSinner: > test hook server is the epitome of trivial

That is a fair point that this one is very unlikely to fail. As an example deployment pattern then, it'd be helpful to go about it as though it can

ThetaSinner (approved):

PR #213, opened 2025-05-09
chore: support cargo-make in Nix devShell

cdunster:


Makefile.toml
```diff
@@ -94,7 +95,7 @@ install_crate = { crate_name = "taplo-cli", version = "0.9.3", binary = "taplo",
 
 [config]
 # Require at least this cargo-make version
-min_version = "0.37.24"
+min_version = "0.37.23"
```
cdunster, line 98: @ThetaSinner, is this okay? Was it just that `.24` was the latest and the version you tested with or was there another reason?

flake.nix
```diff
@@ -46,13 +46,20 @@
           devShells = {
             default = pkgs.mkShell {
               packages = with pkgs; [
-                rust
+                rustup
+                cargo-make
```
ThetaSinner, line 50: Ahh, I didn't think about that. I should really be testing things like this in the Nix shell when I'm adding new tooling.

ThetaSinner (dismissed):

flake.nix
cdunster: Nah, that's what you have me for :smile: We could also use Nix in the CI I guess.

ThetaSinner (approved):

jost-s (approved):

PR #219, opened 2025-05-16
feat(showcase): implement the showcase demo app

### LINKED ISSUE #107: Showcase demo kitsune2 app
- [x] uses all the default module factories
- [x] cli chats messages to everyone... boring but easy to implement / understand
- [x] writes tracing to a `[pid].log` file to make it easy to `cargo run` multiple instances from the same dir
- [x] `/share file.name` -  shares a file if it is less than 1K
- [x] `/list` - lists the files in the dht
- [x] `/fetch file.name` - downloads the file from the dht

cdunster:
Right now, it is heavily integrated with rustyline to act as a REPL. It would be nice to refactor some of the printing etc., so that we could use other interfaces in the future like creating a TUI with [Ratatui](https://crates.io/crates/ratatui) or something.

Closes #107.


crates/kitsune2_showcase/src/app/mod.rs
neonphog, line 1: The rust preference these days is to use `app.rs` rather than `app/mod.rs`. You can still have `app/file_data.rs`. But this way you don't have a bunch of files in your editor that are all named mod.rs : )

crates/kitsune2_showcase/src/readline.rs
```diff
@@ -1,94 +1,182 @@
...
-        if lines.blocking_send(line).is_err() {
-            break;
+    let line_editor = Arc::new(Mutex::new(line_editor));
+
+    loop {
+        let line_editor_2 = line_editor.clone();
+        let prompt = prompt.clone();
+        let line = tokio::task::spawn_blocking(async move || {
+            line_editor_2.lock().await.readline(&prompt)
+        })
```
neonphog, line 101: This isn't doing what you intended. The async closure is returning a future from the `spawn_blocking` so you are back in async land when the blocking readline is run.

You could instead use a normal std::sync::Mutex, instead of a tokio one like:

```rust
tokio::task::spawn_blocking(move || line_editor2.lock().unwrap().readline(&prompt))
```

Basically you can't ever use async closures with spawn_blocking.

crates/kitsune2_showcase/src/app/mod.rs
```diff
@@ -0,0 +1,391 @@
...
+                .send("No ops found".to_string())
+                .await
+                .expect("Failed to print message");
+        }
+
+        Ok(())
+    }
+
+    pub async fn fetch(&self, file_name: &str) -> K2Result<()> {
+        let op_ids = self
```
neonphog, line 345:
```suggestion
    pub async fn fetch(&self, file_name: &str) -> K2Result<()> {
        // This demo application is pulling the full list of local
        // ops in order to find the file name. In a real application
        // it would be more efficient to store the file at its CAS
        // address, then have a link stored at the hash of the
        // filename, so you could look up a single entry to find
        // the CAS address of the file data.
        let op_ids = self
```

neonphog (changes requested):
One async/blocking issue and a couple suggestions, but this looks awesome, nice work!

crates/kitsune2_showcase/src/readline.rs
cdunster: Good catch, thanks. Fixed in 37481d8fdec6e9445dd7a09a251eda4a6311578d to use `blocking_lock` instead so that I can still use the tokio Mutex, which is nicer everywhere else with async code.

crates/kitsune2_showcase/src/app/mod.rs
cdunster: Done: b9856b1f2b322f193f5b8f52fe9571bb96ce9f55

crates/kitsune2_showcase/src/readline.rs
neonphog: It's fine if there is a reason. But [read tokio's docs about it](https://docs.rs/tokio/1.45.1/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use), prefer standard Mutex when you can.

cdunster: I dropped the commit that made that change

neonphog (dismissed):
Awesome :+1:

crates/kitsune2_showcase/src/app/file_op_store.rs
```diff
@@ -0,0 +1,166 @@
...
+        Box::pin(async move {
+            let file_store_lookup = self.file_store_lookup.clone();
+            let out: DynOpStore = Arc::new(FileOpStore {
+                mem_op_store: self.mem_op_store.create(builder, space).await?,
+                file_store_lookup,
+            });
+            Ok(out)
+        })
+    }
+}
```
neonphog, line 56: Wrap the MemOpStore in an Arc to make it clone-able, then:

```suggestion
    fn create(
        &self,
        builder: Arc<Builder>,
        space: SpaceId,
    ) -> BoxFut<'static, K2Result<DynOpStore>> {
        let file_store_lookup = self.file_store_lookup.clone();
        let mem_op_store = self.mem_op_store.clone();
        Box::pin(async move {
            let out: DynOpStore = Arc::new(FileOpStore {
                mem_op_store: mem_op_store.create(builder, space).await?,
                file_store_lookup,
            });
            Ok(out)
        })
    }
}
```

crates/kitsune2_showcase/Cargo.toml
```diff
@@ -12,12 +12,19 @@ edition.workspace = true
...
 kitsune2 = { workspace = true }
 kitsune2_api = { workspace = true }
 kitsune2_core = { workspace = true }
 kitsune2_transport_tx5 = { workspace = true }
 rustls = { workspace = true }
 rustyline = { workspace = true, features = ["derive"] }
+serde = { workspace = true }
+serde_json = { workspace = true }
 sha2 = { workspace = true }
+strum = { version = "0.27.1", features = ["derive", "strum_macros"] }
```
neonphog, line 26: make it a workspace dep

crates/kitsune2_showcase/Cargo.toml
```diff
@@ -12,12 +12,19 @@ edition.workspace = true
...
 tokio = { workspace = true, features = ["full"] }
+tracing = { workspace = true }
+tracing-appender = "0.2.3"
```
neonphog, line 29: what is this for? if needed, please make it a workspace dep

neonphog (changes requested):
Couple comments, but looking good. For next time something like the op id lookup thing comes up, consider if it makes more sense to do it as a follow-on PR instead of adding it to the existing one. Smaller targeted PRs are easier to look through, and reviewers wouldn't have to look through the existing changes again to find the new bits : )

crates/api/src/op_store.rs
```diff
@@ -196,7 +196,7 @@ pub trait OpStoreFactory: 'static + Send + Sync + std::fmt::Debug {
         &self,
         builder: Arc<builder::Builder>,
         space: SpaceId,
-    ) -> BoxFut<'static, K2Result<DynOpStore>>;
+    ) -> BoxFut<K2Result<DynOpStore>>;
```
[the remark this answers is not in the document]
cdunster: But why is it so strict? Is there a need to enforce a static lifetime here?

crates/kitsune2_showcase/Cargo.toml
```diff
@@ -12,12 +12,19 @@ edition.workspace = true
 
 [dependencies]
 bytes = { workspace = true }
+chrono = "0.4.41"
```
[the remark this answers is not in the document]
cdunster: It provides formatting for pretty-printing a time like `2025-05-30T13:18:56.170358+02:00` instead of just a raw timestamp. I don't think that there's a nice way of doing this with std time

crates/kitsune2_showcase/Cargo.toml
cdunster: It's used to print the logs to a file.
Made workspace dep in e57aeb48d55677433734df6e489f5493846288e5

crates/kitsune2_showcase/Cargo.toml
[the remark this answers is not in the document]
cdunster: Made workspace dep in bfcea1b7ae5f98e499c86aa539837d9433729403

crates/kitsune2_showcase/src/app/file_op_store.rs
cdunster: ammended

crates/kitsune2_showcase/Cargo.toml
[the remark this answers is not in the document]
neonphog: oi rust... it's hard to get away from dependency bloat like this. But your logic is sound, since this is a user-facing tool :+1:

ThetaSinner (approved):
Looks great!

Issue #220, opened 2025-05-19
Improve initial sync time


ThetaSinner:
## Background

Currently, gossip initiation is controlled by several variables:

- `initial_initiate_interval_ms` which is used to accelerate the time to the first initiated gossip round. It's possible that on startup, gossip tries to run and finds no peers so it will go to sleep for 2 minutes. This configuration allows gossip to check more often for peers to gossip with up to the first successful round.
- `initiate_interval_ms` which controls how often the gossip module will attempt to initiate gossip after the first successful round.
- `initiate_jitter_ms` is for preventing synchronization where one peer in a pair is always the initiator. Because gossip is asymmetric, you want the initiator to vary. On a large network, this variable isn't really in play so it is a solution for smaller networks to avoid problems.
- `min_initiate_interval_ms` is how often the gossip module will accept permit gossip between two peers. Either after being an initiator or an acceptor, the last gossip time is recorded. The gossip module won't initiate or accept a gossip round with the same peer during this interval.

## Problem

The first gossip round isn't enough to declare a full storage arc. Getting the first round done fast (because of `initial_initiate_interval_ms`) helps get data from the existing network transferred to you. However, it is not until you have a gossip round where your data is in sync with the current gossip peer, that you can declare full arc. 

That can happen on a first gossip round if you happen to gossip when the network is very recently created (last 15 minutes) but that's likely to happen for some instances and not others. Which is what we're seeing in testing.

We are seeing that on larger networks, this behaves as intended. When there are more peers to gossip with and there isn't too much historical data, then your second gossip round is likely to result in picking a peer that you're in sync with and so you end up declaring full arc around the 2 minute mark after starting your program up.

Not having a full storage arc declared prevents other instances from pushing updates to you that they know you'll want to store. This is not a Kitsune2 mechanism, but something that can be built on top. So by optimizing how quickly full arc can be declared on a small network, we can make Holochain's publish mechanism behave better. But it also helps any other application using K2 that expects to see storage arcs matching targets arcs quickly on small networks.

## Possible solutions

The existing mechanism is: When my DHT is in sync (data older than 15 minutes ago) and I get back a gossip message of `NoDiff` during an initiated round, then I can treat that as a signal that my storage arc may be updated to match my target arc. (not entirely accurate in the sharded case, but when both nodes have a full target arc, this is accurate).

[...]

PR #226, opened 2025-05-23
feat: Permit receiving a burst of gossip initiations

ThetaSinner:
Part of #220

### DIFF
--- a/crates/gossip/src/burst.rs
+++ b/crates/gossip/src/burst.rs
@@ -159,7 +159,11 @@ mod tests {
             .get(&peer_url)
             .unwrap()
             .clone();
-        assert_eq!(config.initiate_burst_window_count as usize, open.len());
+        // The content of `open` should be recent timestamps but the actual number is unpredictable
+        // based on the timing of the sleeps above. Just check something is present.
+        // The main logic of the test is to be able to keep on accepting so the assertion in the
+        // loop above is the main check.
+        assert!(!open.is_empty(), "There should be some open timestamps");
     }
 
     #[tokio::test]


neonphog (changes requested):
Couple change requests. Otherwise looks good :+1:

crates/gossip/src/burst.rs
```diff
@@ -0,0 +1,215 @@
...
+            }),
+        }
+    }
+
+    pub(crate) fn check_accept(
+        &self,
+        peer_url: &Url,
+        timestamp: Timestamp,
+    ) -> bool {
+        let max_burst = Self::max_burst(&self.config);
```
zippy, line 62: Why are you re-computing these values with every accept?  Is it because the K2 config is allowed to be modified during runtime?

crates/gossip/src/config.rs
```diff
@@ -81,8 +81,32 @@ pub struct K2GossipConfig {
...
+    ///
+    /// Default: 3
+    #[cfg_attr(feature = "schema", schemars(default))]
+    pub initiate_burst_factor: u32,
+
+    /// The number of initiation windows that the burst is calculated over.
+    ///
+    /// For example, if this is set to `5`, the initiate interval is set to `1m` and the burst
+    /// factor is set to `3` then in 3 minutes, then in a given 3-minute window, 15 gossip rounds
+    /// can be accepted.
```
jost-s, line 104:
```suggestion
    /// For example, if this is set to `5`, the initiate interval is set to `1m` and the burst
    /// factor is set to `3`, then in a given 5-minute window, 15 gossip rounds
    /// can be accepted.
```

crates/gossip/src/config.rs
```diff
@@ -81,8 +81,32 @@ pub struct K2GossipConfig {
...
+    ///
+    /// Default: 5
+    #[cfg_attr(feature = "schema", schemars(default))]
+    pub initiate_burst_window_count: u32,
```
jost-s, line 108: With this factor, in a 5 min window 15 gossip rounds are accepted.
If the factor wasn't applied, in a 1 min window 3 gossip rounds would be accepted, over 5 min 15 gossip rounds. That's not exactly the same, but do you think we can do without the factor for simplicity's sake?

crates/gossip/src/burst.rs
ThetaSinner: Yes, because config can be changed. In theory gossip would cope reasonably well with config being changed but we haven't actually tried that out yet, just left the possibility open

crates/gossip/src/config.rs
jost-s: This was primarily about the window being 5 minutes, not 3.

Issue #42, opened 2024-12-03
[k2] Extend peer meta store


neonphog:
We need a way to understand cross-module some information about peer interactions:
- *DO NOW*: We want to know if a peer has recently been unresponsive, so we can skip even trying to communicate with them for a time period. Use the created_at timestamp from the agent info so that we can mark an agent as unreachable for that agent info's lifetime.
- *DO LATER*: We want to know if we've sent too much data to a specific peer lately, so we want to slow down comms to not use all their bandwidth
- *DO LATER*: We want to know if we've received too much data from a specific peer so we can close the connection with them to avoid all our bandwidth being used

We don't want this to be a "metrics" module.

A "Behavior" module to facilitate this type of information gathering should be designed.

- what all do we want to include in this module for the first iteration
- do these things need to be persisted, or should they be runtime memory only

---

## Implementation notes

We already have a pattern for doing something like this. See the `PeerMetaStore` and the gossip usage of this module as `K2PeerMetaStore` (maybe not the best name, should include gossip?). Add new functionality directly to the `PeerMetaStore` by extending the trait with provided methods for "peer unavailable" (get/set). We just need to take care to keep the key sets separate. The gossip module does this by prefixing keys with `gossip`. For this store we could prefix with `behavior` or `meta`.

For this issue, we should expose an API for recording offline peers and checking whether they are still on cooldown. This will support the next issue https://github.com/holochain/kitsune2/issues/221

ThetaSinner:
Use the peer meta store to store aggregated peer behavior metrics. We need to have at least "completed gossip rounds" for Sweettest's "require initial gossip" method.

ThetaSinner:
I think an unresponsive peer, like one that times out their gossip round or causes a timeout for us is a good use-case for putting a record into the peer meta store with an expiry. We can then quickly check whether we want to open a connection to them.

Bandwidth I think we might want to consider doing in-memory, within Kitsune

zippy:
About this: 

> DO LATER: We want to know if we've received too much data from a specific peer so we can close the connection with them to avoid all our bandwidth being used

If we close that connection, but then end up just asking someone else for that data on the next round, won't we just keep getting the same amount of unsynced data?   This may still make sense if this is also about being nice and asking too much from them.

PR #229, opened 2025-05-29
feat(api)!: add get/set to mark peer unresponsive in peer meta store

jost-s:
Part 1 of #42, which adds methods to the peer meta store to mark and check unresponsive peers, with default implementations.

Added simple test that uses these new methods.

### DIFF
--- a/crates/api/src/peer_meta_store.rs
+++ b/crates/api/src/peer_meta_store.rs
@@ -1,4 +1,5 @@
 use crate::{builder, config, BoxFut, K2Result, Timestamp, Url};
+use bytes::Bytes;
 use futures::future::BoxFuture;
 use std::sync::Arc;
 
@@ -6,21 +7,34 @@ use std::sync::Arc;
 ///
 /// This is expected to be backed by a key-value store that keys by space, peer URL and key.
 pub trait PeerMetaStore: 'static + Send + Sync + std::fmt::Debug {
-    /// Store a key-value pair for a given space and peer.
+    /// Store a key-value pair for a peer.
     fn put(
         &self,
         peer: Url,
         key: String,
-        value: bytes::Bytes,
+        value: Bytes,
         expiry: Option<Timestamp>,
     ) -> BoxFuture<'_, K2Result<()>>;
 
-    /// Get a value by key for a given space and peer.
+    /// Get a value by key for a peer.
     fn get(
         &self,
         peer: Url,
         key: String,
     ) -> BoxFuture<'_, K2Result<Option<bytes::Bytes>>>;
+
+    /// Mark a peer url unresponsive with an expiration timestamp.
+    ///
+    /// After the expiry timestamp has passed, the peer url is supposed to be removed
+    /// from the store.
+    fn mark_peer_unresponsive(
+        &self,
+        peer: Url,
+        expiry: Timestamp,
+    ) -> BoxFuture<'_, K2Result<()>>;
+
+    /// Check if a peer is marked unresponsive in the store.
+    fn is_peer_unresponsive(&self, peer: Url) -> BoxFuture<'_, K2Result<bool>>;
 
     /// Delete a key-value pair for a given space and peer.
     fn delete(&self, peer: Url, key: String) -> BoxFuture<'_, K2Result<()>>;
--- a/crates/core/src/factories/mem_peer_meta_store/test.rs
+++ b/crates/core/src/factories/mem_peer_meta_store/test.rs
@@ -105,3 +105,30 @@ async fn store_with_multiple_agents() {
         Some(timestamp_2)
     );
 }
+
+#[tokio::test]
+async fn store_unresponsive_peer() {
+    let factory = MemPeerMetaStoreFactory::create();
+    let store = factory
+        .create(
+            Arc::new(default_test_builder().with_default_config().unwrap()),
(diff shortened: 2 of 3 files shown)


crates/core/src/factories/mem_peer_meta_store/test.rs
```diff
@@ -39,6 +39,18 @@ impl TestPeerMetaStore {
 
         Ok(())
     }
+
+    async fn put_unresponsive_peer(
```
matthme, line 43: Is there a reason not to call this also `mark_peer_unresponsive`? I think I would find it more intuitive.

crates/api/src/peer_meta_store.rs
```diff
@@ -1,27 +1,48 @@
 use crate::{builder, config, BoxFut, K2Result, Timestamp, Url};
+use bytes::Bytes;
 use futures::future::BoxFuture;
 use std::sync::Arc;
 
 /// A store for peer metadata.
 ///
 /// This is expected to be backed by a key-value store that keys by space, peer URL and key.
 pub trait PeerMetaStore: 'static + Send + Sync + std::fmt::Debug {
-    /// Store a key-value pair for a given space and peer.
```
matthme, line 9: I think removing this here makes sense but maybe it might still be worth having written somewhere (maybe in the top-level comment of the trait) that the `PeerMetaStore` is per space?

matthme: If it is per space, that is. But I thought so.

crates/api/src/peer_meta_store.rs
```diff
@@ -1,27 +1,48 @@
...
 
+    /// Mark a peer url unresponsive with an expiration timestamp.
+    //
+    // TODO After the expiry the peer url will be removed from the store.
+    fn mark_peer_unresponsive(
+        &self,
+        peer: Url,
+        expiry: Timestamp,
+    ) -> BoxFuture<'_, K2Result<()>> {
+        self.put(peer, "unresponsive".to_string(), Bytes::new(), Some(expiry))
```
matthme, line 34:
```suggestion
        self.put(peer, "transport:unresponsive".to_string(), Bytes::new(), Some(expiry))
```

matthme: I hold that suggestion lightly. I just saw that in the gossip module the `gossip:` prefix was used and thought maybe we'd want to be consistent. But also, it's defined here in the `api` crate and not in the transport crate so maybe that prefix would not be appropriate anyway.

crates/api/src/peer_meta_store.rs
```diff
@@ -1,27 +1,48 @@
...
+    }
+
+    /// Check if a peer is unresponsive.
+    fn is_peer_unresponsive(&self, peer: Url) -> BoxFuture<'_, K2Result<bool>> {
+        Box::pin(async move {
+            self.get(peer, "unresponsive".to_string())
```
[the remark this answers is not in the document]
matthme: Same as above. I'm also happy for this to be ignored.

crates/api/src/peer_meta_store.rs
ThetaSinner: It is per space. Our understanding of how that was going to work was still evolving when the documentation was written but now every module except the transport is created per space. 

It wouldn't be a bad thing to document here but it seems like something more for top-level documentation somewhere

crates/api/src/peer_meta_store.rs
```diff
@@ -1,27 +1,48 @@
...
         key: String,
     ) -> BoxFuture<'_, K2Result<Option<bytes::Bytes>>>;
 
+    /// Mark a peer url unresponsive with an expiration timestamp.
+    //
+    // TODO After the expiry the peer url will be removed from the store.
+    fn mark_peer_unresponsive(
+        &self,
+        peer: Url,
+        expiry: Timestamp,
```
ThetaSinner, line 32: Perhaps we could pass an `Arc<AgentInfoSigned>` here so that the caller doesn't have to think about the expiry time? They're going to have to look at up the agent info anyway to work out the expiry, may as well just pass it here for consistency?

crates/core/src/factories/mem_peer_meta_store/test.rs
ThetaSinner: Slightly different suggestion. The `inner` field is visible at the module level so could actually do without these methods and just use `store.inner.mark_peer_unresponsive`?

crates/api/src/peer_meta_store.rs
matthme: Okay, I'm gonna resolve the conversation then.

jost-s:
Thanks for the comments, all addressed in the latest push. The change was small enough to do it all in one.

matthme (approved):

crates/api/src/peer_meta_store.rs
ThetaSinner: Another possible motivation for passing it in. I think we might consider storing the signature as the value of this key. That way we can tell apart a new agent info from re-receiving an existing agent info when deciding whether to un-mark. Though maybe we just do that against the peer store and not inside the meta store logic.

ThetaSinner (approved):
Looks good. Leaving the discussion about passing in the agent info open for now but I'm fine with getting this released in 0.2.x and tested in HC 0.6 before we make a final decision and backport this to 0.1.x for use in HC 0.5

PR #230, opened 2025-05-30
Faster initiate

ThetaSinner:
Closes https://github.com/holochain/kitsune2/issues/220

Depends on #227

### DIFF
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -42,7 +42,7 @@ pub struct K2GossipConfig {
     /// This value should be set to a small value to keep the initial sync fast, but large enough to
     /// avoid using an unreasonable amount of CPU time.
     ///
-    /// Default: 5000 (1s)
+    /// Default: 1000 (1s)
     #[cfg_attr(feature = "schema", schemars(default))]
     pub initial_initiate_interval_ms: u32,
 
--- a/crates/gossip/src/config.rs
+++ b/crates/gossip/src/config.rs
@@ -34,13 +34,15 @@ pub struct K2GossipConfig {
 
     /// The initial interval in milliseconds between initiating gossip rounds.
     ///
-    /// This controls how often Kitsune will check for a peer to gossip with for its first gossip
-    /// round. Once a gossip round has been successfully initiated, this interval will no longer be
-    /// used. The value is used as an upper bound for an exponential backoff. The backoff runs from
-    /// 10ms, with a factor of 1.2, up to this value. Once the maximum value is reached, it
-    /// continues to be used as the upper bound for the backoff.
+    /// This value is used to prevent a busy loop for initiating gossip when trying to complete the
+    /// initial sync for an agent. The primary signal for when to initiate again is the fetch queue
+    /// draining. However, if no peers are available and the fetch queue is empty, initiation would
+    /// keep trying to initiate a gossip round immediately, which would result in a busy loop.
     ///
-    /// Default: 5000 (5s)
+    /// This value should be set to a small value to keep the initial sync fast, but large enough to
+    /// avoid using an unreasonable amount of CPU time.
+    ///
+    /// Default: 5000 (1s)
     #[cfg_attr(feature = "schema", schemars(default))]
     pub initial_initiate_interval_ms: u32,
 
@@ -140,7 +142,7 @@ impl Default for K2GossipConfig {
         Self {
             max_gossip_op_bytes: 100 * 1024 * 1024,
             max_request_gossip_op_bytes: 100 * 1024 * 1024,
-            initial_initiate_interval_ms: 5000,
+            initial_initiate_interval_ms: 1000,
             initiate_interval_ms: 120_000,
             initiate_jitter_ms: 10_000,
             min_initiate_interval_ms: 300_000,
@@ -163,16 +165,6 @@ impl K2GossipConfig {
     /// The interval between initiating gossip rounds.
     pub(crate) fn initiate_interval(&self) -> std::time::Duration {
         std::time::Duration::from_millis(self.initiate_interval_ms as u64)
-    }
-
-    /// The minimum amount of time that must be allowed to pass before a gossip round can be
-    /// initiated by a given peer.
-    // #[deprecated(
-    //     since = "0.1.9",
-    //     note = "This is being replaced by a burst mechanism instead."
-    // )]
-    pub(crate) fn min_initiate_interval(&self) -> std::time::Duration {
-        std::time::Duration::from_millis(self.min_initiate_interval_ms as u64)
     }
 
     /// The timeout for a gossip round.
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -1075,11 +1075,6 @@ mod test {
         iter_check!(1000, 20, {
             let mut all_tried = true;
 
-            // let summary = harness_1.gossip.get_state_summary(GossipStateSummaryRequest {
-            //     include_dht_summary: false
-            // }).await.unwrap();
-            // println!("Gossip summary: {:#?}", summary);
-
             for agent in &junk_agents {
                 all_tried &= harness_1
                     .peer_meta_store
--- a/crates/gossip/src/gossip.rs
+++ b/crates/gossip/src/gossip.rs
@@ -879,6 +879,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[ignore = "This is the wrong place to test this. Check it as a unit test of the gossip message handlers instead. Cannot prevent multiple gossip rounds happening fast any more."]
     async fn respect_size_limit_for_new_ops_and_dht_ring_diff() {
         enable_tracing();
 
@@ -1040,10 +1041,8 @@ mod test {
         )
         .await;
 
-        // The second harness has an empty arc set, so it will be lower priority for target
-        // selection in gossip initiate.
         let harness_2 = factory.new_instance().await;
-        let agent_2 = harness_2.join_local_agent(DhtArc::Empty).await;
+        let agent_2 = harness_2.join_local_agent(DhtArc::FULL).await;
 
         // Create some agent info to use with the first harness so it has to work through some
         // unavailable agents.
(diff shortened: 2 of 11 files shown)


crates/gossip/src/gossip.rs
```diff
@@ -879,6 +879,7 @@ mod test {
     }
 
     #[tokio::test]
+    #[ignore = "This is the wrong place to test this. Check it as a unit test of the gossip message handlers instead. Cannot prevent multiple gossip rounds happening fast any more."]
```
ThetaSinner, line 882: Will need to write a new test for this and not make it an integration test where tasks are running

crates/gossip/src/gossip.rs
```diff
@@ -1040,10 +1041,8 @@ mod test {
         )
         .await;
 
-        // The second harness has an empty arc set, so it will be lower priority for target
-        // selection in gossip initiate.
         let harness_2 = factory.new_instance().await;
-        let agent_2 = harness_2.join_local_agent(DhtArc::Empty).await;
```
ThetaSinner, line 1046: This was a trick to make the test work previously, but the selection logic changing means it no longer works. Instead, this agent is only added to the peer store later, after gossip rounds with the unavailable agents has been attempted.

crates/gossip/tests/initial_sync.rs
ThetaSinner, line 1: This checks the new changes are working. See that we can expect each agent to declare full arc within 5s and we aren't having to set the gossip configuration parameters at all.

The test actually runs in a couple of seconds.

crates/gossip/src/config.rs
```diff
@@ -34,13 +34,15 @@ pub struct K2GossipConfig {
...
+    /// This value is used to prevent a busy loop for initiating gossip when trying to complete the
+    /// initial sync for an agent. The primary signal for when to initiate again is the fetch queue
+    /// draining. However, if no peers are available and the fetch queue is empty, initiation would
+    /// keep trying to initiate a gossip round immediately, which would result in a busy loop.
     ///
-    /// Default: 5000 (5s)
+    /// This value should be set to a small value to keep the initial sync fast, but large enough to
+    /// avoid using an unreasonable amount of CPU time.
+    ///
+    /// Default: 5000 (1s)
```
matthme, line 45:
```suggestion
    /// Default: 5000 (5s)
```

crates/gossip/src/summary.rs
```diff
@@ -130,6 +130,8 @@ impl K2Gossip {
                         .peer_meta_store
                         .peer_timeouts(url.clone())
                         .await?,
+                    is_offline: agent.is_tombstone,
```
matthme, line 133: Why not call the field `is_tombstone`? It seems to me like "offline" is a superset of "tombstone", is that correct?

matthme: On a tangential note, looking at the `PeerStore` trait, it's not clear whether `self.peer_store.get_all()` returns only a single `AgentInfoSigned` per `AgentId` or can also return multiple. I see that the `MemPeerStore` is implemented in a way that it will only ever store a single `AgentInfoSigned` per `AgentId` but it is not required by the trait and there is also no documentation on the `PeerStore` trait that would say anything about that being the intention.

I was just noticing this because `summary.peer_meta.insert()` above would overwrite any existing ones and the result of `self.peer_store.get_all()` may not have been sorted so `agent.is_tombstone` could potentially return different results for the same `Url` depending on how the `Vec` returned by `self.peer_store.get_all()` happens to be sorted.

crates/test_utils/src/agent.rs
```diff
@@ -143,6 +163,15 @@ impl AgentBuilder {
         self
     }
 
+    /// Set the tombstone flag for the agent.
+    ///
+    /// This is a signal from this agent that it is planning to go offline or otherwise doesn't
+    /// want to be contacted anymore.
+    pub fn with_tombstone(mut self, is_tombstone: bool) -> Self {
```
matthme, line 170: To not have to break the `AgentBuilder` later, should we add a `with_[...]` for all other remaining fields as well? `wit_agent()`, `with_space()`, `with_created_at()`, `with_expires_at()`. Could also be a separate PR but might be nice to do as part of the same release.

crates/gossip/tests/initial_sync.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    let harness_2 = factory.new_instance().await;
+    harness_2.join_local_agent(DhtArc::FULL).await;
+
+    // Wait for data to be synced.
+    harness_1
+        .wait_for_sync_with(&harness_2, Duration::from_secs(10))
+        .await;
+
+    // Then both agents should have reach full arc.
```
matthme, line 59:
```suggestion
    // Then both agents should have reached full arc.
```

crates/gossip/tests/initial_sync.rs
```diff
@@ -0,0 +1,125 @@
...
+        TEST_SPACE_ID,
+        true,
+        // Use default config for this test
+        Some(Default::default()),
+    )
+    .await;
+
+    let harness_1 = factory.new_instance().await;
+    let agent_1 = harness_1.join_local_agent(DhtArc::FULL).await;
+    harness_1
```
matthme, line 83: It was a bit hard for me to spot the difference between this and the above test. As I understand it now, the notion of "existing network" in the name of this test refers to there being one agent with declared full arc, is that correct? Maybe it could help to add a small comment here.

crates/gossip/src/initiate.rs
```diff
@@ -18,74 +17,101 @@ pub fn spawn_initiate_task(
...
-            }
-
             let Some(gossip) = gossip.upgrade() else {
                 tracing::info!("Gossip instance dropped, stopping initiate task");
                 break;
             };
 
+            let Ok(local_agents) = gossip.local_agent_store.get_all().await else {
+                tracing::warn!("Failed to get local agents, pausing and then retrying");
+                tokio::time::sleep(initial_initiate_interval).await;
```
matthme, line 42: What's the reasoning for taking the `initial_initiate_interval` here? It may not matter at all anyway but when looking at [the implementation](https://github.com/holochain/kitsune2/blob/2ca14f5885a2b186b673d3664dca79664bb4afb6/crates/core/src/factories/core_local_agent_store.rs#L39-L43) of `get_all()` it seems that it cannot fail anyway (except when it panics but that wouldn't show up here). So I'm assuming that this sleep logic here is intended for potential different kinds of implementations where it *could* fail? Would we expect it to fail in the beginning stage of gossip only, hence why the `initial_initiate_interval`?

crates/gossip/src/initiate.rs
```diff
@@ -356,9 +354,8 @@ mod tests {
         let harness = Harness::create().await;
```
matthme, line 355: This is unrelated to the PR but I'm wondering whether we should also assert in this test that `harness.local_agent_store.get_all().await.len() == 0`?

crates/gossip/src/initiate.rs
```diff
@@ -18,74 +17,101 @@ pub fn spawn_initiate_task(
...
+                continue;
+            };
+
+            if local_agents.is_empty() {
+                tracing::warn!("No local agents available, skipping initiation");
+                tokio::time::sleep(initial_initiate_interval).await;
+                continue;
+            }
+
+            if local_agents.iter().all(|a| a.get_cur_storage_arc() == a.get_tgt_storage_arc()) {
```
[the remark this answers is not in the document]
ThetaSinner: Hmm, that's an interesting one. Yes that is a valid edge case but I'm wondering whether that's okay. If the existing local agent was synced, then the new local agent is joining the same space and already has all the content in the other local agent's arc. So for full arc, a new agent joining doesn't cause any new data to need to be synced and publish to us will happen because of the other agent declaring full sync already. So I think this edge case is actually specific to sharding where the new local agent might want to store a different DHT section to the current local agent. 

I'm tempted to defer dealing with this until we start looking at sharding because it's going to require a test with partial arcs to make sense and while we can do that, we haven't got much testing for partial arcs outside of the DHT crate.

crates/gossip/src/summary.rs
matthme: > We only accept the insert if the incoming agent info is valid and newer than the current one

Yes correct, in the actual implementation we're using that is the case. It is just not enforced by the `PeerStore` *trait* so in principle, this *could* be implemented differently without being aware that this has implications here downstream because there's nothing in the trait's documentation telling, that the `insert()` method should ensure a single `AgentInfoSigned` per `AgentId`.

crates/gossip/src/initiate.rs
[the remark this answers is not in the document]
matthme: Okay, sounds reasonable. Should we add a comment in the code that we don't forget about it?

crates/gossip/tests/initial_sync.rs
ThetaSinner: Yep, reading this back it's not clear enough. Added more comments to the test https://github.com/holochain/kitsune2/pull/230/commits/846819386204d3f9f26ab367d8c02ea9d0f552b6

crates/test_utils/src/agent.rs
ThetaSinner: https://github.com/holochain/kitsune2/pull/230/commits/846819386204d3f9f26ab367d8c02ea9d0f552b6

matthme (approved):

crates/gossip/src/initiate.rs
```diff
@@ -18,74 +17,110 @@ pub fn spawn_initiate_task(
...
+                // to be drained and then go ahead with the initiation.
+                let (tx, rx) = futures::channel::oneshot::channel();
+                gossip.fetch.notify_on_drained(tx);
+                rx.await.ok();
+
+                // If the fetch queue is empty and there's nobody to gossip with then this loop
+                // would just spin. So use the initial initiate interval to wait a short amount of
+                // time before trying to initiate again.
+                // For this reason, this configuration value should be set to a small value!
+                tokio::time::sleep(initial_initiate_interval).await;
```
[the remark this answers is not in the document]
ThetaSinner: Possibly. We don't really have a way to set configuration ourselves that other devs then aren't able to change. We've put all this behind an `advanced` field in Holochain's configuration to discourage changing values randomly.

I did document this in a similar way on the field to how I've written the code comment here. Is that okay?

jost-s (approved):

Issue #221, opened 2025-05-19
Record peer connection failures in the behavior store


ThetaSinner:
Depends on: https://github.com/holochain/kitsune2/issues/42

This requires a change to the `transport_tx5` module. It will need to get a reference to the peer behavior module and when a peer connections fails, record it with the behavior module. 

I suggest that the transport module cannot control how long the peer is put on timeout for. This logic should be contained within the behavior store. The transport module simply records the failure.

This issue is part of the solution to a larger issue -> https://github.com/holochain/holochain/issues/4978

ThetaSinner:
Notes: 
- Possibly need a sanity check that we're online, otherwise we might block peers because we can't reach them, not because they're unavailable. See `tx5_online`. <- do as a follow up
- When coming back online after connection issues on your end, will have to produce a new agent info. Could use that as a signal (only when our pub key changes, not just arc?) to clear all unavailable peers.
- Keep simple for now, and mark peers unreachable if connection attempts time out or the connection fails.

---

Plan:
- When changing peer url, clear unreachable peers
- Later, when online check changes from failing to passing, clear unreachable peers
- Now, mark peers unreachable when there is any network error, until their agent info expires

jost-s:
resolved by #233

PR #233, opened 2025-06-03
Record peer connection failures in the peer meta store

matthme:
Addresses https://github.com/holochain/kitsune2/issues/221

May need additional integration tests at the holochain level with an actual peer meta store implementation.

### DIFF
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -134,8 +134,9 @@ impl TxImpHnd {
         }
     }
 
-    /// Call this whenever a connection to a peer fails to get established
-    /// or when sending a message to a peer fails.
+    /// Call this whenever a connection to a peer fails to get established,
+    /// sending a message to a peer fails or we get a disconnected event from
+    /// a peer.
     pub fn mark_peer_unresponsive(&self, peer: Url) {
         for (space_id, space_handler) in self.space_map.lock().unwrap().iter() {
             if let Err(e) = space_handler.mark_peer_unresponsive(peer.clone()) {
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -136,8 +136,8 @@ impl TxImpHnd {
     }
 
     /// Call this whenever a connection to a peer fails to get established,
-    /// sending a message to a peer fails or we get a disconnected event from
-    /// a peer.
+    /// sending a message to a peer fails or when we get a disconnected
+    /// event from a peer.
     pub fn mark_peer_unresponsive(&self, peer: Url) {
         for (space_id, space_handler) in self.space_map.lock().unwrap().iter() {
             if let Err(e) = space_handler.mark_peer_unresponsive(peer.clone()) {
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -10,7 +10,6 @@ use std::sync::{Arc, Mutex, Weak};
 /// Construct using ([TxImpHnd::new]), with a high-level [DynTxHandler],
 /// then call [DefaultTransport::create] to return the high-level handler
 /// from the [TransportFactory].
-#[derive(Debug)]
 pub struct TxImpHnd {
     handler: DynTxHandler,
     space_map: Arc<Mutex<HashMap<SpaceId, DynTxSpaceHandler>>>,
@@ -144,6 +143,12 @@ impl TxImpHnd {
                 tracing::error!("Failed to mark peer with url {peer} as unresponsive in space {space_id}: {e}");
             };
         }
+    }
+}
+
+impl std::fmt::Debug for TxImpHnd {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f,"TxImpHnd {{ handler: {:?}, space_map: [{} entries], mod_map: [{} entries] }}", self.handler, self.space_map.lock().unwrap().len(), self.mod_map.lock().unwrap().len() )
     }
 }
 
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -133,6 +133,16 @@ impl TxImpHnd {
             }
         }
     }
+
+    /// Call this whenever a connection to a peer fails to get established
+    /// or when sending a message to a peer fails.
+    pub fn mark_peer_unresponsive(&self, peer: Url) {
+        for (space_id, space_handler) in self.space_map.lock().unwrap().iter() {
+            if let Err(e) = space_handler.mark_peer_unresponsive(peer.clone()) {
+                tracing::error!("Failed to mark peer with url {peer} as unresponsive in space {space_id}: {e}");
+            };
+        }
+    }
 }
 
 /// A low-level transport implementation.
@@ -447,6 +457,12 @@ pub trait TxSpaceHandler: TxBaseHandler {
         data: bytes::Bytes,
     ) -> K2Result<()> {
         drop((peer, space, data));
+        Ok(())
+    }
+
+    /// Mark a peer as unresponsive in the space's peer meta store
+    fn mark_peer_unresponsive(&self, peer: Url) -> K2Result<()> {
+        drop(peer);
         Ok(())
     }
 }
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -137,15 +137,12 @@ impl TxImpHnd {
     /// Call this whenever a connection to a peer fails to get established,
     /// sending a message to a peer fails or when we get a disconnected
     /// event from a peer.
-    pub fn mark_peer_unresponsive(
-        &self,
-        peer: Url,
-    ) -> BoxFut<'_, K2Result<()>> {
+    pub fn set_unresponsive(&self, peer: Url) -> BoxFut<'_, K2Result<()>> {
         let space_map = self.space_map.lock().unwrap().clone();
         Box::pin(async move {
             for (space_id, space_handler) in space_map.iter() {
                 if let Err(e) =
-                    space_handler.mark_peer_unresponsive(peer.clone()).await
+                    space_handler.set_unresponsive(peer.clone()).await
                 {
                     tracing::error!("Failed to mark peer with url {peer} as unresponsive in space {space_id}: {e}");
                 };
@@ -477,7 +474,7 @@ pub trait TxSpaceHandler: TxBaseHandler {
     }
 
     /// Mark a peer as unresponsive in the space's peer meta store
-    fn mark_peer_unresponsive(&self, peer: Url) -> BoxFut<'_, K2Result<()>> {
+    fn set_unresponsive(&self, peer: Url) -> BoxFut<'_, K2Result<()>> {
         drop(peer);
         Box::pin(async move { Ok(()) })
     }
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -10,6 +10,7 @@ use std::sync::{Arc, Mutex, Weak};
 /// Construct using ([TxImpHnd::new]), with a high-level [DynTxHandler],
 /// then call [DefaultTransport::create] to return the high-level handler
 /// from the [TransportFactory].
+#[derive(Debug)]
 pub struct TxImpHnd {
     handler: DynTxHandler,
     space_map: Arc<Mutex<HashMap<SpaceId, DynTxSpaceHandler>>>,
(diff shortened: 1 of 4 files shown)


crates/transport_tx5/src/test.rs
```diff
@@ -525,9 +536,161 @@ async fn preflight_send_recv() {
...
+
+    assert!(r.is_err());
+
+    // Now peer 2 goes offline and we check that it gets marked as unresponsive
+    drop(t2);
+
+    // We need to wait for a while in order for the webrtc connection to get
+    // disconnected. If this test becomes flaky, this waiting period may need
+    // to be increased a bit.
+    tokio::time::sleep(std::time::Duration::from_secs(7)).await;
```
matthme, line 669: This is a bit unfortunate but due to the nature of webrtc, the connection will only close 6-7 seconds after the peer goes offline so any send attempt earlier than that will look like it succeeded from the sender side. Hence why we need to wait a bit here.

crates/transport_tx5/src/test.rs
```diff
@@ -525,9 +536,161 @@ async fn preflight_send_recv() {
     .unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn nonexistent_peer_marked_unresponsive() {
```
matthme, line 540: Not sure this test is necessary but it was the simpler one to write first =)

crates/transport_tx5/src/lib.rs
```diff
@@ -442,6 +444,7 @@ async fn evt_task(
                         continue;
                     }
                 };
+                handler.mark_peer_unresponsive(peer_url.clone());
```
matthme, line 447: Marking peers as unresponsive also in case of a "Disconnected" event. I don't know how to test this independently from the `mark_peer_unresponsive()` call above when the send fails because they both get invoked in the test scenario that I set up.

crates/transport_tx5/src/lib.rs
```diff
@@ -316,17 +319,16 @@ impl TxImp for Tx5Transport {
...
-            self.ep
-                .send(peer.clone(), data.to_vec())
-                .await
-                .map_err(|e| {
-                    K2Error::other_src(
-                        format!("tx5 send error to peer at url {peer}"),
-                        e,
-                    )
-                })
+            if let Err(e) = self.ep.send(peer_url, data.to_vec()).await {
```
matthme, line 324: No error distinction for now.

crates/core/src/factories/core_space.rs
```diff
@@ -198,6 +198,39 @@ impl TxSpaceHandler for TxHandlerTranslator {
...
     }
+
+    fn mark_peer_unresponsive(&self, peer: Url) -> K2Result<()> {
+        let core_space = self.1.upgrade().ok_or(K2Error::Other {
+            ctx: "CoreSpace had been dropped.".into(),
+            src: DynInnerError(None),
+        })?;
+        tokio::task::spawn(async move {
+            // Only add a peer as unreachable to the peer meta store if it is
+            // also in the peer store. That's because this method may be called
```
ThetaSinner, line 209: Is that necessary to assume? yes the transport doesn't know, but it's looking up the right space based on the space id. So while this is technically possible, I think it's okay to accept the input

crates/core/src/factories/core_space.rs
```diff
@@ -198,6 +198,39 @@ impl TxSpaceHandler for TxHandlerTranslator {
...
+                Some(agent_info) => {
+                    if let Err(e) = core_space
+                        .peer_meta_store
+                        .mark_peer_unresponsive(
+                            agent_info.url.clone().unwrap(),
+                            agent_info.expires_at,
+                        )
+                        .await
+                    {
+                        tracing::error!("Failed to mark peer at url {:?} as unresponsive in the peer meta store: {e}", agent_info.url);
```
ThetaSinner, line 224: We use an inconsistent style for this through K2 but I'm starting to prefer the way that this is rendered if you use:

`tracing::error!(?err, "some message with more context {}", agent_info.url);`

Entirely a style preference for how the output looks and not a "must do" comment :)

crates/transport_tx5/src/lib.rs
ThetaSinner: I don't know about this. Sometime (I hope reasonably) soon, we will actually be closing connections we don't need. I don't think it's the case that we shouldn't re-open this connection if we need to.

If the peer was online and able to successfully notify us of closing their connection then it's valid to try to connect to them again on the same peer URL I think.

crates/transport_tx5/src/test.rs
ThetaSinner: Seems valid to me. This ties in with what you were asking about with "what if we get a peer info that's bogus and just produced to waste our time", but better put than my quick summary there :D 

I think we absolutely want to know that will be marked unavailable. 

Generally, I like simple tests that check things that seem obvious. Because just occasionally they aren't and it would have been easy to write a test for them and catch the problem

crates/transport_tx5/src/test.rs
```diff
@@ -525,9 +536,161 @@ async fn preflight_send_recv() {
...
+            bytes::Bytes::from_static(b"anyone here?"),
+        )
+        .await;
+
+    // We expect it to have timed out because no peer should be reachable at above URL
+    assert!(res.is_err());
+
+    // We expect the mark_peer_unresponsive() method on the TxSpaceHandler trait to have
+    // been invoked
+    let r = tokio::time::timeout(std::time::Duration::from_secs(5), async {
```
ThetaSinner, line 579: Nice test logic

crates/transport_tx5/src/test.rs
```diff
@@ -525,9 +536,161 @@ async fn preflight_send_recv() {
...
+    .await
+    .unwrap()
+    .unwrap();
+
+    println!("{r:?}");
+    assert_eq!(b"hello", r.2.as_ref());
+
+    // Check that the peer has not been marked as unresponsive
+    let r = tokio::time::timeout(std::time::Duration::from_secs(3), async {
+        unresp_recv.recv().await
```
ThetaSinner, line 657: Consider using `unresp_recv.try_recv` here? I think a quick point in time check is acceptable because we got a response?

crates/transport_tx5/src/test.rs
ThetaSinner: That's interesting. So "not getting a response" isn't enough. We actually do need the connection to be come invalid...

Okay, well we can come back and look for other ways to force the connection to break faster if we have problems here. It is a bit slow then this test but it's an important thing to test.

I am wondering if we should consider moving slower tests like this to the integration tests so that `cargo test --lib` stays fast. But then we probably don't want to do that on this PR because the test harness will have to be exposed from the crate to be used in integration tests.

ThetaSinner (changes requested):
This is looking good! I am requesting changes because of:
- Possibly can avoid fetching all peers from the peer store
- Marking agents unreachable if they close their connection isn't necessarily correct, I want to check what's going on there

crates/transport_tx5/src/lib.rs
matthme: Okay, so we should rather remove this line, right?

crates/core/src/factories/core_space.rs
matthme: 3efccaaf4434f6b38331ecd477ae48f224553f6d

crates/core/src/factories/core_space.rs
```diff
@@ -198,6 +198,39 @@ impl TxSpaceHandler for TxHandlerTranslator {
     ) -> K2Result<()> {
         self.0.recv_notify(peer, space, data)
     }
+
+    fn mark_peer_unresponsive(&self, peer: Url) -> K2Result<()> {
+        let core_space = self.1.upgrade().ok_or(K2Error::Other {
```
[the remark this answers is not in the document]
matthme: 3efccaaf4434f6b38331ecd477ae48f224553f6d

crates/api/src/transport.rs
```diff
@@ -10,6 +10,7 @@ use std::sync::{Arc, Mutex, Weak};
 /// Construct using ([TxImpHnd::new]), with a high-level [DynTxHandler],
 /// then call [DefaultTransport::create] to return the high-level handler
 /// from the [TransportFactory].
+#[derive(Debug)]
```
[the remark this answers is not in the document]
matthme: > was the content immediately useful?

No, not really. I only added the Debug trait there because it was required in order to be able to add it as a field to the `Tx5Transport` struct. I can look into a custom debug implementation.

crates/transport_tx5/src/lib.rs
matthme: Removed: 40e7afd790424b6091df035fdf7f0213ceebee6e

crates/transport_tx5/src/test.rs
```diff
@@ -45,6 +48,7 @@ impl Test {
                 tx5_transport: config::Tx5TransportConfig {
                     signal_allow_plain_text: true,
                     server_url: format!("ws://127.0.0.1:{}", this.port),
+                    timeout_s: tx5_timeout_s.unwrap_or(60),
```
jost-s, line 51:
```suggestion
                    timeout_s: tx5_timeout_s.unwrap_or_default(),
```

crates/transport_tx5/src/test.rs
```diff
@@ -525,9 +536,157 @@ async fn preflight_send_recv() {
     .unwrap();
 }
 
+#[tokio::test(flavor = "multi_thread")]
```
jost-s, line 539: I know it's the style of the existing tests here, but I invite you to break that pattern and use more descriptive identifiers for variables. One letter names always require me to look twice what something is.

crates/transport_tx5/src/test.rs
ThetaSinner: Won't that pick the default for the type, rather than for the field? Not sure what behavior to expect there

matthme: Yes, I think so too. It would pick the default for the type which would be `0`. Not ideal that I cannot use the actual intended default here easily though I agree...but I didn't see a better way.

crates/api/src/transport.rs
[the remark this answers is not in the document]
matthme: 5f6518a464928c22c3c1e39925f775ff8f522173

crates/core/src/factories/core_space.rs
jost-s: We just talked about this and I was mistaken. We identified three ways of doing this
1. the way it's implemented on this PR
2. mark URLs unresponsive in all spaces' peer stores (might become a lot?)
3. add peer stores to the transport, so that it can search through them to find where the peer with the URL is

I think how it's done on this PR is fine.

ThetaSinner: I see, because the space isn't available in the transport at this point. Should that be changed though? This `send` is only called from `send_space_notify` and `send_module`, which doesn't pass the space down because it's in the encoded message. You could decode the message on this code path but it'd probably be cheaper to just pass the space id. Then you do know which space to target and don't have to broadcast this?

crates/transport_tx5/src/test.rs
ThetaSinner: Hmm, I'm not sure how to explain that. Perhaps the connection was idling even before the drop happened so the reduced wait before the drop means the same total amount of time is needed for the connection to get dropped. I'm not sure!

crates/core/src/factories/core_space.rs
jost-s: Aah, yes, that's best, I think.

matthme: > Then you do know which space to target and don't have to broadcast this?

I actually think we'd want to broadcast it anyway because the peer url likely doesn't only belong to a single space and we'd want to mark that url as unreachable for all spaces that it's part of I assume. At least that was my thinking.

matthme: If we think it makes sense to forward the space id anyway since we're going to break the API with the next release I'm fine with that. But I think we still want to iterate through all spaces. I should however probably update the doc comment then because wanting to mark a url as unreachable for all spaces that it's part of is the primary reason then and not the lack of knowledge of the space id.

crates/core/src/factories/core_space.rs
```diff
@@ -198,6 +198,29 @@ impl TxSpaceHandler for TxHandlerTranslator {
...
+            ctx: "CoreSpace had been dropped.".into(),
+            src: DynInnerError(None),
+        })?;
+        tokio::task::spawn(async move {
+            let peers = core_space.peer_store.get_all().await?;
```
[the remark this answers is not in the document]
matthme: Okay, sounds like we don't need the new method then :+1:

ThetaSinner (dismissed):

PR #234, opened 2025-06-10
Update API for peer meta store unresponsive URLs

jost-s:


ThetaSinner (commented):
Looks good, one request for `expect` -> `Err`

crates/api/src/peer_meta_store.rs
```diff
@@ -25,16 +31,46 @@ pub trait PeerMetaStore: 'static + Send + Sync + std::fmt::Debug {
...
+            .await?;
+            Ok(())
+        })
+    }
 
-    /// Check if a peer is marked unresponsive in the store.
-    fn is_peer_unresponsive(&self, peer: Url) -> BoxFuture<'_, K2Result<bool>>;
+    /// Get the timestamp of when a peer URL was marked unresponsive, if it is present in the
+    /// store.
+    fn get_unresponsive_url(
```
matthme, line 59:
```suggestion
    fn get_unresponsive_for_url(
```

matthme: I would actually still find `is_peer_unresponsive()` appropriate as a name.

jost-s: How about this?

matthme (dismissed):

crates/api/src/peer_meta_store.rs
jost-s: I renamed to `get_unresponsive` and `mark_unresponsive`, but then though it's more consistent to match the get/set pattern and ended up with `get_unresponsive` and `set_unresponsive`.

ThetaSinner (approved):

crates/api/src/peer_meta_store.rs
```diff
@@ -25,16 +31,50 @@ pub trait PeerMetaStore: 'static + Send + Sync + std::fmt::Debug {
...
+                Some(expiry),
+            )
+            .await?;
+            Ok(())
+        })
+    }
 
-    /// Check if a peer is marked unresponsive in the store.
-    fn is_peer_unresponsive(&self, peer: Url) -> BoxFuture<'_, K2Result<bool>>;
+    /// Get the timestamp of when a peer URL was marked unresponsive, if it is present in the
```
matthme, line 57:
```suggestion
    /// Get the timestamp of when a peer URL was last marked unresponsive, if it is present in the
```

matthme (dismissed):

Issue #222, opened 2025-05-19
Gossip to filter out peers that have failed to connect


ThetaSinner:
Depends on: https://github.com/holochain/kitsune2/issues/42
and is probably best done after: https://github.com/holochain/kitsune2/issues/221

In the gossip initiation task, we have peer selection logic that takes into account storage arc and how recently we gossiped with our peers. This logic should be updated to include filtering out peers that we've recently failed to connect to.

We should simply query the peer behavior to see if an agent is marked as unreachable and if so, they are not a valid target for gossip.

ThetaSinner:
- Could check for unreachable agent during the gossip process, before sending responses.

PR #235, opened 2025-06-11
Filter unresponsive peers in gossip

matthme:
Adresses #222

### DIFF
--- a/crates/gossip/src/initiate.rs
+++ b/crates/gossip/src/initiate.rs
@@ -187,7 +187,7 @@ async fn select_next_target(
         using_overlapping_agents = false;
     }
 
-    let mut possible_target = select_least_recently_gossiped(
+    let mut possible_target = select_responsive_and_least_recently_gossiped(
         all_agents,
         peer_meta_store.clone(),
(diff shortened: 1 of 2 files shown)


matthme:
I've thought of that possibility too but I discarded it for the following 2 reasons:
- in the `select_responsive_and_least_recently_gossiped()` function we can piggy-back on the `for agents in all_agents` loop that we're already doing. A `remove_unresponsive_agents()` function would have to do another separate, dedicated for loop which seemed a bit redundant
- the `remove_local_agents()` function is called multiple times and we'd have to be careful if any of the logic related to that changes that we'd also make sure to always filter out unresponsive agents at the same time. There would be the possibility to convert it to a single function `remove_local_and_unresponsive_agents()` that would do both. But that function is currently not async and does not have access to the peer meta store (which of course could be changed too)

So long story short, I didn't see much functional advantage between either of the solutions but found it much more ergonomic and a bit more computationally efficient to pack it into the peer selection function, assuming that this function would be called in any scenario anyway.

jost-s (approved):
ok

PR #237, opened 2025-06-13
Filter unresponsive URLs in fetch and publish modules

jost-s:
- Omit unresponsive URLs in publish
- Omit fetch requests to URLs set unresponsive in peer meta store
- Replace existing back off in fetch module by handling unresponsive URLs in peer meta store
- Remove back off crate backon from Kitsune2

I favor this solution in fetch over the previous back off mechanism. It integrates with other modules and gives greater control in tests.

### DIFF
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -1,6 +1,7 @@
 use kitsune2_api::*;
 use message_handler::FetchMessageHandler;
 use std::collections::HashMap;
+use std::sync::MutexGuard;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex},
@@ -341,28 +342,10 @@ impl CoreFetch {
             // from state and no request will be sent and the request
             // will not be re-inserted into the queue.
             {
-                let mut state_lock = state.lock().expect("poisoned");
-                if !state_lock
-                    .requests
-                    .contains(&(op_id.clone(), peer_url.clone()))
-                {
-                    // Check if the fetch queue is drained.
-                    //
-                    // Either requests in the state is empty, in which case the queue is
-                    // drained and no further request is sent, or it is not empty and the
-                    // next request will be sent.
-                    if state_lock.requests.is_empty() {
-                        // Notify all listeners that the fetch queue is drained.
-                        for notify in
-                            state_lock.notify_when_drained_senders.drain(..)
-                        {
-                            if notify.send(()).is_err() {
-                                tracing::warn!(
-                                    "Failed to send notification on drained"
-                                );
-                            }
-                        }
-                    }
+                let lock = state.lock().expect("poisoned");
+                if !lock.requests.contains(&(op_id.clone(), peer_url.clone())) {
+                    // Check if the fetch queue is drained and notify listeners.
+                    Self::notify_listeners_if_queue_drained(lock);
 
                     continue;
                 }
@@ -422,12 +405,26 @@ impl CoreFetch {
                             tracing::warn!(
                                 "could not re-insert fetch request for op {op_id} to peer {peer_url} into queue: {err}"
                             );
-                            // Remove op id/peer url from set to prevent build-up of state.
+                            // Remove request from set to prevent build-up of state.
                             lock.requests.remove(&(op_id, peer_url));
+
+                            Self::notify_listeners_if_queue_drained(lock);
                         }
                     }
                 }
             });
+        }
+    }
+
+    fn notify_listeners_if_queue_drained(mut state: MutexGuard<State>) {
+        // Check if the fetch queue is drained.
+        if state.requests.is_empty() {
+            // Notify all listeners that the fetch queue is drained.
+            for notify in state.notify_when_drained_senders.drain(..) {
+                if notify.send(()).is_err() {
+                    tracing::warn!("Failed to send notification on drained");
+                }
+            }
         }
     }
 


crates/api/src/peer_meta_store.rs
```diff
@@ -29,6 +29,12 @@ pub trait PeerMetaStore: 'static + Send + Sync + std::fmt::Debug {
         key: String,
     ) -> BoxFuture<'_, K2Result<Option<bytes::Bytes>>>;
 
+    /// Get all peers for a given key.
+    fn get_all_by_key(
```
jost-s, line 33: A new function for the peer meta store to get all URLs with values for a given meta key, e. g. to get all unresponsive URLs.

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
 
-                lock.back_off_list.is_peer_on_back_off(&peer_url)
-            };
+            // If peer URL is set as unresponsive, do not send request or re-insert
+            // request into queue, and remove all remaining requests from state.
+            let peer_url_unresponsive = peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some());
+            if peer_url_unresponsive {
```
jost-s, line 337: new check in fetch

crates/core/src/factories/core_fetch.rs
```diff
@@ -418,31 +387,33 @@ impl CoreFetch {
...
-                async move {
-                    tokio::time::sleep(Duration::from_millis(
-                        re_insert_outgoing_request_delay as u64,
-                    ))
-                    .await;
-                    if let Err(err) = outgoing_request_tx
-                        .try_send((op_id.clone(), peer_url.clone()))
-                    {
-                        tracing::warn!(
+            if !peer_url_unresponsive {
```
jost-s, line 390: This is new too. Before requests were always re-inserted into the request queue. Now they're only re-inserted if the peer is not unresponsive. My thinking is that if a URL is unresponsive, there's no point to keep the requests in the queue, and if the URL becomes available again, gossip will happen again and insert missing ops in the request queue again. If the URL is renewed for the peer, it's good that the requests are no longer in the queue.

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
+            let peer_url_unresponsive = peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some());
+            if peer_url_unresponsive {
+                state
+                    .lock()
+                    .unwrap()
+                    .requests
+                    .retain(|(_, a)| *a != peer_url);
```
jost-s, line 342: This is new. Before the back off had to go through all intervals before removing all requests for the unresponsive URL. Now if the URL is unresponsive, all of the peer URL's requests are removed straight away. See below for my reasoning.

crates/core/src/factories/core_publish/test.rs
```diff
@@ -1,121 +1,38 @@
...
+    MockTransport, Publish, Signer, SpaceHandler, SpaceId, Timestamp,
+    TxBaseHandler, TxHandler, TxSpaceHandler, Url,
 };
 use kitsune2_test_utils::{
     agent::{TestLocalAgent, TestVerifier},
     enable_tracing, iter_check,
     space::TEST_SPACE_ID,
 };
-
-use crate::factories::{core_publish::CorePublish, MemoryOp};
```
jost-s, line 15: This is just moved to the bottom.

crates/core/src/factories/core_publish.rs
```diff
@@ -282,11 +293,22 @@ impl CorePublish {
...
             outgoing_publish_ops_rx.recv().await
         {
+            // Check if peer URL to publish to is unresponsive.
+            if peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some())
+            {
+                // Peer URL is unresponsive, do not publish.
+                continue;
```
jost-s, line 309: new check for unresponsiveness

crates/core/src/factories/core_publish.rs
```diff
@@ -104,6 +104,7 @@ impl PublishFactory for CorePublishFactory {
         space_id: SpaceId,
         fetch: DynFetch,
         peer_store: DynPeerStore,
+        peer_meta_store: DynPeerMetaStore,
```
jost-s, line 107: Peer meta store now added to publish

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
+            }
 
-                lock.back_off_list.is_peer_on_back_off(&peer_url)
-            };
+            // If peer URL is set as unresponsive, do not send request or re-insert
+            // request into queue, and remove all remaining requests from state.
+            let peer_url_unresponsive = peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some());
```
matthme, line 336: Would there potentially be use in logging errors here instead of silently converting them to a boolean?

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
+                .requests
+                .contains(&(op_id.clone(), peer_url.clone()))
+            {
+                continue;
+            }
 
-                lock.back_off_list.is_peer_on_back_off(&peer_url)
-            };
+            // If peer URL is set as unresponsive, do not send request or re-insert
+            // request into queue, and remove all remaining requests from state.
```
matthme, line 332:
```suggestion
            // request into queue, and remove all remaining requests for that peer from state.
```

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
+                state
+                    .lock()
+                    .unwrap()
+                    .requests
+                    .retain(|(_, a)| *a != peer_url);
+            }
 
-            // Send request if peer is not on back off list.
-            if !is_peer_on_back_off {
+            if !peer_url_unresponsive {
```
matthme, line 345: Maybe convert this to an `else` here? Seems a bit prettier to me but is a detail.

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
```diff
@@ -273,200 +269,170 @@ async fn filter_requests_for_held_ops() {
...
+        .expect("Timed out waiting for drained notification")
+        .unwrap();
+
+    // Add control agent's ops to set.
+    fetch
+        .request_ops(op_list_2, responsive_url.clone())
+        .await
+        .unwrap();
+
+    // Wait for a request attempt to the unresponsive URL, which should remove all of its requests
```
matthme, line 411: I'm not clear what this check is testing here, it seems redundant. Or at least sending another message to the responsive url above seems redundant. Since `notify_on_drained` had been emitted earlier I would naturally expect this `iter_check!` here to return immediately since requests of the unresponsive peer had already been filtered out. Or am I missing something?

ThetaSinner (commented):
Looking good. A handful of minor requests and a couple of things that I think would be useful to tweak. At a high level though the code changes look logical, the tests are thorough and the code has been tidied up nicely around the changes :)

crates/core/src/factories/core_fetch.rs
jost-s: I didn't think so. You guys do, so I'll change it.

crates/core/src/factories/core_fetch.rs
jost-s: done 421d26a9bb6b9fd6e9f058a92de69a8de1b0b2dc

crates/core/src/factories/core_fetch.rs
```diff
@@ -337,22 +315,34 @@ impl CoreFetch {
...
+            // If peer URL is set as unresponsive, do not send request or re-insert
+            // request into queue, and remove all remaining requests from state.
+            let peer_url_unresponsive = peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some());
+            if peer_url_unresponsive {
+                state
+                    .lock()
+                    .unwrap()
```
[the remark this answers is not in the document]
jost-s: I see this is the new pattern. That outputs
```
poisoned: PoisonError { .. }
```
instead of
```
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```
It slightly reduces the legibility. Do you prefer `expect` anyway?

crates/core/src/factories/core_fetch.rs
jost-s: Clearly so!

crates/core/src/common.rs
```diff
@@ -61,11 +112,33 @@ async fn filter_local_agents(
...
+        )
+        .await?;
+
+    Ok(agent_list
+        .into_iter()
+        .filter(|a| {
+            if let Some(url) = a.url.clone() {
+                !unresponsive_urls.contains_key(&url)
+            } else {
+                true
```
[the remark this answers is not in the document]
jost-s: Yes, I agree. This function will probably just be called to send something to the agents, and if they don't have a URL, they can't be contacted anyway.

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
jost-s: No, you're not missing something. The refactor of this existing test was incomplete. I could have just checked for the drained queue and not check the other part. I've changed it so that there's the control agent to ensure that only the unresponsive agent's requests are dropped.

crates/core/src/factories/core_publish.rs
```diff
@@ -282,11 +293,22 @@ impl CorePublish {
...
+            if peer_meta_store
+                .get_unresponsive(peer_url.clone())
+                .await
+                .is_ok_and(|maybe_value| maybe_value.is_some())
+            {
+                // Peer URL is unresponsive, do not publish.
+                continue;
+            }
+
             let Some(transport) = transport.upgrade() else {
```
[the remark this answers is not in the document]
jost-s: Okay

crates/core/src/factories/mem_peer_meta_store.rs
```diff
@@ -46,14 +47,32 @@ impl PeerMetaStore for MemPeerMetaStore {
...
-    ) -> BoxFuture<'_, K2Result<Option<bytes::Bytes>>> {
+    ) -> BoxFuture<'_, K2Result<Option<Bytes>>> {
         let inner = self.inner.clone();
         Box::pin(async move {
             let inner = inner.lock().await;
             Ok(inner.get(&peer).and_then(|entry| entry.get(&key).cloned()))
         })
     }
 
+    fn get_all_by_key(
```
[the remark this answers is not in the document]
jost-s: Important point. The autoindex does cater for lookups to the first column of a compound primary key, but not the second column. That's coherent with other indexes. So we need to add another index on `meta_key` to the database definition.

crates/api/src/fetch.rs
```diff
@@ -141,12 +142,6 @@ pub struct FetchStateSummary {
     /// Each op id is associated with one or more peer URL from which the op data could be
     /// requested.
     pub pending_requests: HashMap<OpId, Vec<Url>>,
-
-    /// The peer URL for nodes that are currently on backoff because of failed fetch requests, and the timestamp when that backoff will expire.
-    ///
-    /// If peers are in here then they are not being used as potential sources in
-    /// [`FetchStateSummary::pending_requests`].
-    pub peers_on_backoff: HashMap<Url, Timestamp>,
```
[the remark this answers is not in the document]
jost-s: Sure

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
```diff
@@ -502,21 +466,16 @@ async fn fetch_queue_notify_on_last_op_fetched() {
...
-    // The queue should still be working, so the notification should not be sent yet.
-    assert!(rx.try_recv().unwrap().is_none());
-
-    tokio::time::timeout(Duration::from_secs(5), rx)
+    tokio::time::timeout(Duration::from_secs(1), rx)
         .await
         .expect("Timed out")
         .unwrap();
-
-    assert!(fetch.state.lock().unwrap().requests.is_empty());
```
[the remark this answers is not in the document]
jost-s: It went much further. The queue drained notification wasn't working correctly. As described above the test was only making sure that the notification is sent when all requests fail - when all URLs are unresponsive. When requests succeeded, however, the check whether the queue is drained was skipped, and the comment that said it was safe to do so was mistaken (because it was based on the test with unresponsive URLs).

I've added a test, so now there is one which makes sure notification happens when URLs are responsive and when they're unresponsive. This allowed me to streamline and simplify the flow of the outgoing request task. Happy days!

crates/core/src/factories/core_fetch.rs
jost-s: https://github.com/holochain/kitsune2/pull/237/commits/8a271e2922698eb67a49092d67d953975aaf332e Although, that's gone now anyway.

crates/core/src/common.rs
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/kitsune2/pull/237/commits/80c2ae64426b35d8f1e151e9327b08ecea19c310

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
```diff
@@ -273,200 +269,170 @@ async fn filter_requests_for_held_ops() {
...
+    // Request ops from another URL which is not unresponsive, so that this request
+    // can be awaited to have happened, which implies that the previous
+    // publish to the unresponsive URL has been omitted.
+    fetch
+        .request_ops(op_list_2, unresponsive_url.clone())
+        .await
+        .unwrap();
 
-        lock.back_off_list.first_back_off_interval_ms
+    // Timeout after waiting 100 ms for fetch to request from the responsive URL.
```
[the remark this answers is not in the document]
jost-s: https://github.com/holochain/kitsune2/pull/237/commits/d58da19fbf98431ea39459ed4e86c9aa310b37a9

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
jost-s: https://github.com/holochain/kitsune2/pull/237/commits/22f1116d641015c15d94fe6456332f004d559662

crates/core/src/factories/core_fetch.rs
jost-s: https://github.com/holochain/kitsune2/pull/237/commits/63a2a02dec5815830e0a7852f9da1099a3769a3d

crates/core/src/factories/core_fetch.rs
```diff
@@ -324,98 +301,100 @@ impl CoreFetch {
...
-                            .back_off_list
-                            .has_last_back_off_expired(&peer_url)
+                    // Check if the fetch queue is drained.
+                    //
+                    // Either requests in the state is empty, in which case the queue is
+                    // drained and no further request is sent, or it is not empty and the
+                    // next request will be sent.
+                    if state_lock.requests.is_empty() {
+                        // Notify all listeners that the fetch queue is drained.
+                        for notify in
```
ThetaSinner, line 355: I don't see how this logic can be complete. The only case where we will send a notification is when the request coming in from the queue is no longer in the list of ops to fetch. 

I'm also wondering if this can't lead to duplicate notification attempts for the same queue empty event. If we have 5 fetch requests for the same peer in the queue+state and we discover they are offline when processing the first request then we'll clear the state. When processing the second request we'll run this code and then continue processing the queue. For items 3, 4 and 5 we'll also try to notify.

I can see that there's a case here that wasn't covered where the state gets emptied by an unresponsive check and we never reach the point of sending a notification. But I'm not clear how there isn't a case where the queue clears successfully, never having something in the queue that isn't in state, and we now also miss sending a notification in that case?

Or maybe I just don't understand how this works. Do we expect the request to be put back in the queue without waiting for a response. The response arriving clears the state and then when we next process the queued request, we hit this state?

crates/core/src/factories/core_fetch.rs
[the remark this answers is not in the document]
jost-s: Done https://github.com/holochain/kitsune2/pull/237/commits/f4536f47970431a0d055bad730aa6b44f1c50d50

crates/core/src/factories/core_fetch.rs
```diff
@@ -408,18 +408,23 @@ impl CoreFetch {
...
                                 "could not re-insert fetch request for op {op_id} to peer {peer_url} into queue: {err}"
                             );
-                        // Remove op id/peer url from set to prevent build-up of state.
-                        state
-                            .lock()
-                            .expect("poisoned")
-                            .requests
-                            .remove(&(op_id, peer_url));
+                            // Remove op id/peer url from set to prevent build-up of state.
+                            lock.requests.remove(&(op_id, peer_url));
```
matthme, line 426: Do we need to add a check here if after this removal as well whether the requests are empty and send a queue drained notification?

crates/core/src/factories/core_fetch.rs
```diff
@@ -324,98 +302,83 @@ impl CoreFetch {
...
+            // Do nothing if op id is no longer in the set of requests to send.
+            //
+            // If the peer URL is unresponsive, the current request will have been removed
+            // from state and no request will be sent and the request
+            // will not be re-inserted into the queue.
+            {
+                let lock = state.lock().expect("poisoned");
+                if !lock.requests.contains(&(op_id.clone(), peer_url.clone())) {
+                    // Check if the fetch queue is drained and notify listeners.
+                    Self::notify_listeners_if_queue_drained(lock);
```
ThetaSinner, line 348:
```suggestion
                    Self::notify_listeners_of_queue_drained(lock);
```

PR #240, opened 2025-06-17
Backport faster initial sync 0.1

ThetaSinner:


jost-s (commented):
Aren't these breaking changes since they're adding a new fn to the trait and new struct fields without a default?

PR #244, opened 2025-06-18
chore: Update Cargo.toml use repository instead of homepage

szabgab:
According to the [manifset](https://doc.rust-lang.org/cargo/reference/manifest.html) it seems the `repository` field is preferable.  More explanation https://github.com/szabgab/rust-digger/issues/91

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,8 @@ resolver = "2"
 [workspace.package]
 version = "0.2.8"
 authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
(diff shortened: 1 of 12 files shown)


szabgab:
You can try to use the [Rust Digger](https://rust-digger.code-maven.com/) to find the crates with such issues. You can locate them [by owner](https://rust-digger.code-maven.com/people) or some of the [recent lists](https://rust-digger.code-maven.com/stats) or if you have a suggestion how to select these crates I might be able to build you a custom report.

PR #246, opened 2025-06-19
Prepare next release

holochain-release-automation2:
Automated preparation for the next release, triggered by cdunster.


jost-s (approved):
There are going to be more changes for the upcoming release. We'd create another one of these then, I suppose?

CHANGELOG.md
```diff
@@ -0,0 +1,427 @@
...
+### Removed
+
+- Remove only current request from state when url unresponsive by @jost-s
+- Remove backoff and replace with unresponsive function in peer meta store by @jost-s
+
+## First-time Contributors
+
+* @szabgab made their first contribution in [#244](https://github.com/holochain/kitsune2/pull/244)
+
+## [0.2.7] - 2025-05-23
```
ThetaSinner, line 62: No 0.2.8 between 0.2.7 and 0.3.0?

Looking above I'd actually expect everything except "Update Cargo.toml use repository instead of homepage" to already be released. I really like the changelog but I'm thinking that it should go up to 0.2.8 and then we don't do a release just yet? As we've already released the breaking changes on 0.2.x, the reason to bump to 0.3.x has probably passed? Perhaps we shouldn't break Holochain dev versions and should be releasing 0.2.X-dev.Y versions until we are ready to release Kitsune2 0.2...

ThetaSinner (changes requested):
See comment about 0.3.x vs 0.2.8

cdunster:
These changes are already released under `v0.2.8`, so I am closing this PR in favour of #248, which just adds the changelog.

PR #250, opened 2025-06-20
chore(release): bump to v0.1.9

cdunster:


ThetaSinner (approved):
🎉

ThetaSinner:
> Wait though, isn't the tx5 update supposed to go into this release?

Only into K2 0.2.x for now. This version of K2 uses tx5 0.3.x which means I'd have to backport and release the tx5 changes. I don't think we have to for 0.5, I'm most interested in fixing the dependencies that are causing people problems and otherwise minimizing disruption

PR #253, opened 2025-06-20
Improve flaky two new agents sync test

ThetaSinner:


ThetaSinner:
Two test failures, neither of them are the one I was targeting...

jost-s (approved):

PR #254, opened 2025-06-24
ci: add job to generate changelog preview with git-cliff

cdunster:
PR to add a comment to this and future PRs to show what will be added to the generated changelog, including what version bump will happen due to these changes. See [the comment below](https://github.com/holochain/kitsune2/pull/254#issuecomment-3000465692) for an example.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -132,6 +132,10 @@ jobs:
       - name: Build
         run: docker buildx build -f docker/kitsune2_test_auth_hook_server/Dockerfile .
 
+  changelog-preview-comment:
+      name: Add comment of changelog preview
+      uses: holochain/actions/.github/workflows/changelog-preview-comment.yml@v1.2.0
+
   ci_pass:
     if: ${{ always() }}
     runs-on: "ubuntu-latest"


.github/workflows/test.yaml
```diff
@@ -132,6 +132,43 @@ jobs:
...
+            --output=PR_CHANGELOG.md
+
+      - name: Add changelog preview as comment
+        run: |
+          gh pr comment ${{github.event.pull_request.html_url }} \
+            --body="$(printf 'The following will be added to the CHANGELOG.md\n---\n\n'; cat PR_CHANGELOG.md)" \
+            --edit-last \
+            --create-if-none
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ThetaSinner, line 170:
```suggestion
      - name: Add changelog preview as comment
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          gh pr comment ${{github.event.pull_request.html_url }} \
            --body="$(printf 'The following will be added to the CHANGELOG.md\n---\n\n'; cat PR_CHANGELOG.md)" \
            --edit-last \
            --create-if-none
```

ThetaSinner: I find it slightly easier to digest when the env is provided before the script but just a personal preference probably :)

ThetaSinner (approved):
Very cool!

Should this go into `holochain/actions` as a reusable workflow and be a workflow call here?

PR #255, opened 2025-06-24
Improve gossip test coverage

ThetaSinner:


crates/gossip/src/gossip.rs
```diff
@@ -712,274 +712,6 @@ mod test {
         assert_eq!(op_1, received_ops[0]);
     }
 
-    #[tokio::test]
-    async fn respect_size_limit_for_new_ops() {
```
ThetaSinner, line 716: These 3 tests are important but were relying on being able to run one gossip round and then do assertions. That's no longer reliable with the faster initiation. Replacing these with unit tests

jost-s (approved):
Tests look great, legible and clear. Seeing how directly this can be tested underlines that the gossip protocol is well encapsulated.

Generally I prefer to use variables defined at the beginning of the test for things like budget and op size and then calculate in the assertions with these variables, e.g.

```rust
    let op_size = 128;
    let budget = 7;
    let number_of_ops = 5;
    for i in 0u8..number_of_ops {
        let op = MemoryOp::new(Timestamp::now(), vec![i; op_size]);
        ops.push(op);
    }
    assert_eq!(number_of_ops, sent_ops.len(), "Should have sent all 5 ops");

    assert_eq!(
        budget - number_of_ops * op_size,
        remaining_budget,
        "Should have remaining budget for 2 more ops"
    );
```

Then I don't need to double-check that the assertions use the right numbers. Not requesting changes based on that.

ThetaSinner:
> Then I don't need to double-check that the assertions use the right numbers. Not requesting changes based on that.

Agreed that's a nice pattern for tests this long, I thought it was worth the changes https://github.com/holochain/kitsune2/pull/255/commits/a9301e7404b2855b8ef1141d8c3970409b1e203e

jost-s (approved):
So approved!

PR #273, opened 2025-07-25
refactor: Rename space to space_id where applicable

matthme:
Renames `space` to `space_id` where applicable (and where it doesn't introduce breaking changes to the bootstrap server) to reduce ambiguity.


crates/api/proto/gen/kitsune2.fetch.rs
```diff
@@ -1,6 +1,6 @@
 // This file is @generated by prost-build.
 /// A fetch request.
-#[derive(Clone, PartialEq, ::prost::Message)]
+#[derive(Clone, PartialEq, Eq, Hash, ::prost::Message)]
```
matthme, line 3: Interesting that this got added now, I don't know why.

crates/bootstrap_srv/src/parse.rs
```diff
@@ -5,7 +5,7 @@ pub struct ParsedEntry {
     pub agent: ed25519_dalek::VerifyingKey,
 
     /// space
-    pub space: bytes::Bytes,
+    pub space_id: bytes::Bytes,
```
matthme, line 8: This `ParsedEntry` type seems to be stored and parsed from temp files. Might this be a problem temporarily when bumping the server version on a deployed instance? If we think it's problematic I can revert it.

crates/bootstrap_srv/src/server.rs
```diff
@@ -236,17 +236,17 @@ impl Handler<'_> {
     /// Respond to a request for the agent infos within a space.
     fn handle_boot_get(
         &mut self,
-        space: bytes::Bytes,
+        space_id: bytes::Bytes,
```
matthme, line 239: Only in-memory variable name changes in this file.

crates/api/proto/gen/kitsune2.fetch.rs
ThetaSinner: Probably because it's a new version of the library and we haven't regenerated the Rust code. It was on the TODO list at some point to check that these generated files are up to date on CI. Not sure if that ever made it to an issue.

crates/kitsune2_showcase/src/app.rs
```diff
@@ -143,7 +143,7 @@ impl App {
         let kitsune = builder.build().await?;
         kitsune.register_handler(h).await?;
         let transport = kitsune.transport().await?;
-        let space = kitsune.space(space).await?;
```
matthme, line 146: Particularly nasty - shadowing with a different type.

ThetaSinner (approved):

PR #277, opened 2025-07-27
Make `Kitsune2MemoryOpStore` generic over its op type

maackle:
I'm going to try integrating kitsune2 into a project as a standalone, with no Holochain. So far so good, but here is a change I found helpful.

`Kitsune2MemoryOpStore` is useful for testing other kitsune hosts, but it is written to be used with `MemoryOp` which is clearly for testing purposes -- its `OpId` implementation is intentionally not robust, making it unsuitable for real-world situations. Aside from that, host implementers will have written their own custom op types, and having to convert to `MemoryOp` isn't great. A better pattern is that any reusable OpStore implementation should be abstract over its op data.

I defined a trait that has the key elements necessary for working with your mem op store implementation: `created_at` timestamp, op id calculation, and serialization. This lets users more quickly get up and running with `Kitsune2MemoryOpStore`, which as far as I know is the *only* working OpStore implementation in the world aside from Holochain's. Without this, your first attempts at hooking up your own custom op type likely result in confusing deserialization errors, and in my case, discovering that I got the same OpId for every op due to the first 32 bytes being the same.

If you don't want to use this that's fine, but I will be using this in my branch.

### DIFF
--- a/crates/core/src/factories/core_bootstrap.rs
+++ b/crates/core/src/factories/core_bootstrap.rs
@@ -78,20 +78,24 @@ impl BootstrapFactory for CoreBootstrapFactory {
     }
 
     fn validate_config(&self, config: &Config) -> K2Result<()> {
-        const ERR: &str = "invalid bootstrap server_url";
-
         let config: CoreBootstrapModConfig = config.get_module_config()?;
 
+        let err = format!(
+            "invalid bootstrap server_url: {:?}",
+            config.core_bootstrap.server_url
+        );
+        let err = err.as_str();
+
         let url = url::Url::parse(&config.core_bootstrap.server_url)
-            .map_err(|e| K2Error::other_src(ERR, e))?;
+            .map_err(|e| K2Error::other_src(err, e))?;
 
         if url.cannot_be_a_base() {
-            return Err(K2Error::other(ERR));
+            return Err(K2Error::other(err));
         }
 
         match url.scheme() {
             "http" | "https" => Ok(()),
-            _ => Err(K2Error::other(ERR)),
+            _ => Err(K2Error::other(err)),
         }
     }
 
--- a/crates/core/src/factories/core_fetch/test.rs
+++ b/crates/core/src/factories/core_fetch/test.rs
@@ -4,7 +4,7 @@ mod outgoing_request_queue;
 
 #[cfg(test)]
 pub(crate) mod test_utils {
-    use crate::factories::MemoryOp;
+    use crate::factories::TestMemoryOp;
     use kitsune2_api::{Timestamp, Url};
     use rand::RngCore;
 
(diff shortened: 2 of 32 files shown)


maackle:
I will also sneak a totally unrelated question in here: is there a way to register a handler that gets called whenever new ops have been received via publish/gossip/fetch? Or must one simply poll the op store?

maackle:
> Is there something else about the memory op store that is useful, beyond getting started quicker?

No, but the value of getting started quicker can't be understated. I'm trying to get a demo of an app ready in two weeks, and it doesn't have to be robust, it just has to have a p2p experience. Getting something working with a mem op store ASAP lets me have all of this without investing in building my own op store:

- see if kitsune is actually going to be a good fit for this app
- see if kitsune is going to perform well enough for my demo
- start figuring out the p2p patterns I need to use for my app
- have confidence that any problems are at least probably not due to the op store
- have a demo sooner

> not persisting ops is not usable outside tests, there is also no op location logic used in tests with the memory op store

My plan was to set up an always-online node that just syncs with other people. Good enough for a demo, and even lets me test initial sync over and over. Also I don't have any need for location, all Spaces will be full-sync for a long time, and I'm not doing anything like holochain's authorities pattern.

> Kitsune2 could obviously use more documentation about which trait implementations from the core module are intended to be re-used, and which ones the host is supposed to implement. The op store is one the host is expected to implement.

It seems to me that in the ideal world, the host shouldn't be expected to *implement* an op store, but rather initially to *select* one from a list of quality first- and third-party op store implementations. Maybe implementing their own if needed. Most framework users are accustomed to picking a pre-made "backend" or "driver", which this basically is.

Regardless of whether something is *designed* to be reused or not, people who are trying to get stuff done are certainly going to attempt to reuse and repurpose them anyway (especially when there are no other guides), so why not go the extra mile and make them reusable? :)

----------

I'm definitely not going to invest in building my own op store at this time, especially since I have one that already works for me. Depending how this goes I may or may not invest further in making a persistent op store, if one doesn't already exist by then. If you're not already thinking this way, I would encourage you to think of every op store as reusable, as I think that will be important for actual adoption of Kitsune.

Last question, would you recommend I use the windtunnel op store you linked to instead of the one I'm currently using?

maackle:
> I have already written one that uses the Fjall key-value database, which is why I'm defending the position that having a persistent store is practical outside of Holochain

Of course my ideal is to have a persistent op store, and I imagine it would be the ideal for just about anyone else. I'm only using the mem one because it's available and it works now.

> I really don't want people to use the memory op store in a binary. I was thinking this PR was about using it in tests but with the op type being defined as something other than the memory op.

Yes, most normal use cases would want a persistent store, including mine. Still, I wouldn't go so far as to say there are no valid reasons for using a mem op store in production. Off the top of my head: imagine IoT devices that don't have persistent storage of their own and/or are extremely limited in memory, and want to do some ephemeral syncing of small amounts of data and don't need to install a database for that. p2p syncing of data is a very general pattern and I can't imagine anticipating all real world use cases up front.

Or even in the course of development of a custom op store, it's quite useful to have a simple-as-possible working reference implementation, both as guidance and also as something testable, so you can A/B test to find if the problem is in your own implementation. The more complex the reference implementation, the less useful it is.

> I would like to have more example and/or ready-to-use code for Kitsune2 but we haven't had time to do that.

When you do please let me know! Until then I'll keep using the only op store that exists.

No further reply needed unless it's about wanting to use this, otherwise feel free to close.

PR #279, opened 2025-07-28
chore: Allow a dedicated override for `OpId`

ThetaSinner:
Part of https://github.com/holochain/holochain/issues/5163

I'm pretty sure that most implementations on top of Kitsune2 will actually need different logic for op ids than other ids. I'm trying to make the API change compatible so it can be backported right back to 0.1 of K2 and used in Holochain 0.5.

Note that because of the statics, there are no tests for this code and I've not attempted to add any as part of this PR.

### DIFF
--- a/crates/api/src/id.rs
+++ b/crates/api/src/id.rs
@@ -24,7 +24,7 @@ macro_rules! imp_from {
     };
 }
 
-/// The function signature for Id loc derivation.
+/// The function signature for [`Id`] location derivation.
 pub type LocCb = fn(&bytes::Bytes) -> u32;
 
 fn default_loc(b: &bytes::Bytes) -> u32 {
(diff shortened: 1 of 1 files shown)


crates/api/src/id.rs
```diff
@@ -60,34 +60,35 @@ imp_from!(Id, bytes::Bytes, b => Id(b));
...
 impl Id {
-    /// Get the location u32 based off this Id.
+    /// Get the location u32 based off this [`Id`].
     ///
     /// This is accomplished by directly xor-ing every successive 4 bytes
     /// in the hash. It is okay if the hash len is not a multiple of 4,
     /// it will stop with the ending byte of the hash.
     ///
     /// The remaining 4 bytes are then interpreted as a little-endian u32.
-    //
```
ThetaSinner, line 70: Notes from the original design of this interface, no longer required.

cdunster (approved):

PR #292, opened 2025-08-06
test: when a tie-breaker check results in a tie

cdunster:
There are tests for when a tie-breaker check results in a win and a loss, but not for the unlikely case that the tie-breaker results in a tie. This PR adds that test.

There is #316 to explicitly define the behaviour of a tie in a tie-breaker check, so this PR is only to test the current behaviour.

## Summary by CodeRabbit

* **Tests**
  * Added a test covering the equal tie-breaker scenario when two peers initiate a gossip session: verifies the local side accepts and clears its initiated state, and that an unsolicited Accept from the peer produces the expected error.
  * Refined test setup and imports to simplify dependencies and focus the scenario.

### DIFF
--- a/crates/gossip/src/respond/initiate.rs
+++ b/crates/gossip/src/respond/initiate.rs
@@ -591,6 +591,104 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn resolve_tie_break_tie() {
+        enable_tracing();
+
+        let mut harness = RespondTestHarness::create().await;
+
+        let remote_agent = harness.create_agent(DhtArc::Empty).await;
+
+        // Initiate a session with the remote agent
+        let initiated = harness
+            .gossip
+            .initiate_gossip(remote_agent.url.clone().unwrap())
+            .await
+            .unwrap();
+        assert!(initiated);
+
+        // Wait for us to send the initiate message
+        let response = harness.wait_for_sent_response().await;
+        let initiate = match response {
+            GossipMessage::Initiate(initiate) => initiate,
+            other => panic!("Expected initiate message, got: {:?}", other),
+        };
+
+        assert!(initiate.tie_breaker > 0, "Expected tie breaker to be set");
+
+        // Receive an initiate message. It doesn't need to be valid for this test, it just needs to
+        // have the same value for the tie-breaker.
+        let arc_set = ArcSet::new(vec![DhtArc::FULL]).unwrap();
+        harness
+            .gossip
+            .respond_to_msg(
+                remote_agent.url.clone().unwrap(),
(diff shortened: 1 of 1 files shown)


crates/gossip/src/respond/initiate.rs
```diff
@@ -593,6 +593,67 @@ mod tests {
...
+            .unwrap();
+
+        // Check that we accepted the session
+        let response = harness.wait_for_sent_response().await;
+        match response {
+            GossipMessage::Accept(accept) => accept,
+            other => panic!("Expected accept message, got: {:?}", other),
+        };
+
+        // Check that we removed our initiated state
```
ThetaSinner, line 649: Also assert that there is now an accepted state with the other peer? If we cleared our state, wouldn't we accept the incoming session?

crates/gossip/src/respond/initiate.rs
```diff
@@ -593,6 +590,103 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn resolve_tie_break_tie() {
```
ThetaSinner, line 594:
```suggestion
    async fn tie_breaker_conflict_cancels_gossip_session() {
```

The flow here is that both sides initiate at the same time with the same tie breaker. Then our side spots that our tie_breaker is *not* greater than the other side's tie_breaker and so we drop our initiate and send back an accept. We can actually expect that the other side would do the same thing so this test is really saying that if both sides send an initiate, at the same time, with the same tie_breaker, then both sides will cancel their session and try to accept the other one which is likely going to result in the session failing to initiate either way right?

ThetaSinner (commented):
Perhaps I'm still not quite getting this and there are multiple cases based on timing here. I think the cases where we've accepted and still send an initiate are super rare because that's a local race condition. If both sides initiate at the same time it seems to me like both drop their session and send an unsolicited accept which will result in the session being cancelled?

crates/gossip/src/respond/initiate.rs
cdunster: I can change the name; I was trying to keep it consistent with the other two tests `resolve_tie_break_win` and `resolve_tie_break_lose`.

As for the logic then yeah, it's all very timing-specific and incredibly unlikely that two agents will initiate at the same time **and** choose the same random number for the `tie_breaker` which is why I'm not too worried about "fixing" it and just created #316 just so that we're aware. I do think it's something that we should test, though, so we can hopefully catch if the logic changes.

If it does happen, then I think that both agents will accept and wait for the other agent to do something, and they will remain in this state until the gossip-round timeout elapses, which is set to 15 seconds, then they will try again.

ThetaSinner (approved):

PR #303, opened 2025-08-14
fix: Close connections when marking a peer unresponsive

ThetaSinner:
I expected the transport to just deal with this but it seems that it can take a bit of time for the transport to notice that something is wrong even if you get an error that causes you to mark the peer as unresponsive.

## Summary by CodeRabbit

- New Features
  - None

- Bug Fixes
  - Ensure peer endpoints are closed after failed sends to unresponsive/nonexistent peers, preventing stale connections and improving stability and reported network stats.

- Tests
  - Enabled tracing for diagnostics, switched to dynamic test URLs, and added assertions to confirm no active connections remain after failure scenarios.

### DIFF
--- a/crates/transport_tx5/src/test.rs
+++ b/crates/transport_tx5/src/test.rs
@@ -3,6 +3,7 @@
 use crate::harness::{MockTxHandler, Tx5TransportTestHarness};
 
 use super::*;
+use kitsune2_test_utils::enable_tracing;
 use kitsune2_test_utils::space::TEST_SPACE_ID;
 use std::collections::HashSet;
 use std::sync::Mutex;
@@ -347,6 +348,8 @@ async fn preflight_send_recv() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn nonexistent_peer_marked_unresponsive() {
+    enable_tracing();
+
     // set the tx5 timeout to 5 seconds to keep the test reasonably short
     let test = Tx5TransportTestHarness::new(None, Some(3)).await;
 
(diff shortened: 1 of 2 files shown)


crates/transport_tx5/src/test.rs
```diff
@@ -371,9 +374,10 @@ async fn nonexistent_peer_marked_unresponsive() {
         tx_handler1.clone(),
     );
 
-    let faulty_url = Url::from_str(
-        "ws://127.0.0.1:40813/VtK2IOCncQM6LbWkvhB_CYwajQzw6Dii-Oc-0IRtHmc",
-    )
+    let faulty_url = Url::from_str(format!(
```
ThetaSinner, line 377: If we don't use the right signal URL then yes you get an error back but it never attempts to make a connection because it fails to set up the signal connection. Not really what the test is supposed to be checking.

matthme (approved):

crates/transport_tx5/src/test.rs
matthme: Hmm...yes and slightly yes. The code that marks peer unresponsive also doesn't distinguish these two cases, right? So the test was testing as far as the actual code can distinguish but not what the code may ideally want to be able to distinguish. But I agree, it's certainly better with a valid signal url.

ThetaSinner: Yeah that's fair, the test as it was tested that any error would result in a peer getting marked unreachable. It was fine in that respect. It wasn't useful to me with checking the effect on the connection status without this change.

PR #308, opened 2025-08-14
Improve flaky tests

ThetaSinner:
## Summary by CodeRabbit

* **Tests**
  * Increased timeouts in sync tests to improve stability in slower environments.
  * Swapped the order of waiting for full synchronization between peers to reduce flakiness.
  * Extended the gossip wait window (from ~5s to ~15s) while keeping check intervals unchanged.

### DIFF
--- a/crates/gossip/tests/initial_sync.rs
+++ b/crates/gossip/tests/initial_sync.rs
@@ -57,11 +57,11 @@ async fn two_new_agents_sync() {
         .await;
 
     // Then both agents should have reached full arc.
-    harness_1
-        .wait_for_full_arc_for_all(Duration::from_secs(5))
-        .await;
     harness_2
-        .wait_for_full_arc_for_all(Duration::from_secs(5))
+        .wait_for_full_arc_for_all(Duration::from_secs(10))
+        .await;
+    harness_1
+        .wait_for_full_arc_for_all(Duration::from_secs(10))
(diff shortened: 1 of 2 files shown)


matthme (approved):

crates/gossip/tests/initial_sync.rs
```diff
@@ -57,11 +57,11 @@ async fn two_new_agents_sync() {
         .await;
 
     // Then both agents should have reached full arc.
-    harness_1
-        .wait_for_full_arc_for_all(Duration::from_secs(5))
-        .await;
     harness_2
```
ThetaSinner, line 60: In theory, harness 2 should be ready first because it comes online second and will boostrap and initiate right away. That's not guaranteed but it's what I'd expect most of the time. 

Either way, increase the time these get to be ready. Locally this test reliably finishes in 2.1s...

crates/kitsune2/tests/integration.rs
```diff
@@ -171,7 +171,7 @@ async fn two_node_gossip() {
         .unwrap();
 
     // Wait for gossip to exchange all ops.
-    iter_check!(5000, 500, {
+    iter_check!(15000, 500, {
```
ThetaSinner, line 174: Same here, locally this test finishes in a couple of seconds.

PR #330, opened 2025-09-05
MemTransport to complete preflight before sending messages

ThetaSinner:
Fix an issue in the `MemTransport` that allowed notify and module messages to be sent before the preflight exchange was complete

## Summary by CodeRabbit

- Bug Fixes
  - Prevented messages from sending before a connection completes its handshake, reducing intermittent failures.
  - Added a 5s readiness timeout with clearer error reporting on failed handshakes or disconnects.
  - Improved handling of initial connection messages for more reliable startup.

- Refactor
  - Enforced readiness gating in the in-memory transport so send/disconnect paths wait for a successful handshake.

- Tests
  - Enabled tracing in targeted tests and added checks to ensure preflight (handshake) occurs before regular messages.

### DIFF
--- a/crates/core/src/factories/mem_transport/test.rs
+++ b/crates/core/src/factories/mem_transport/test.rs
@@ -10,6 +10,8 @@ enum Track {
     Disconnect(Url, Option<String>),
     SpaceRecv(Url, SpaceId, bytes::Bytes),
     ModRecv(Url, SpaceId, String, bytes::Bytes),
+    PreflightSend,
+    PreflightRecv,
 }
 
 type G = Box<dyn Fn(Url) -> K2Result<bytes::Bytes> + 'static + Send + Sync>;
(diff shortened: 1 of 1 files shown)


crates/core/src/factories/core_space/test.rs
```diff
@@ -93,6 +95,8 @@ async fn space_local_agent_join_leave() {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn space_notify_send_recv() {
+    enable_tracing();
```
matthme, line 98: Do you want to keep this tracing? I don't bother about having tracing in tests in general but it seems to be the only change in this file.

matthme (approved):
Nice solution!

crates/core/src/factories/core_space/test.rs
ThetaSinner: I do, ideally it'd be enabled in all tests. I enabled tracing in two tests that I was debugging while making this change and it was really useful to be able to see what they're doing.

PR #338, opened 2025-10-01
Remove re insertion logic from fetching ops

jost-s:
Fetch requests for ops used to be re-inserted into the fetch request queue, until the incoming op would end these retries. This PR removes this re-insertion logic and simplifies the fetch request queue.

The flow used to be to add fetch requests to a state object, send the fetch requests and then re-insert these requests back into the fetch request queue after a delay. When the op would come in, all requests for this op still remaining in the state object were removed.

This is now simplified and every fetch request is sent once. Requests are not re-inserted into the queue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Immediately purge failed outgoing requests to unresponsive peers and ensure state updates occur before enqueueing to prevent stuck queues and race conditions.
  - Enhanced structured error logging for enqueue failures.

- **Refactor**
  - Removed delayed re-insert/backoff behavior and simplified outgoing-request ordering and handling; one configuration option was removed (breaking change).

- **Documentation**
  - Clarified outgoing fetch flow and failure handling to reflect immediate purge behavior.

- **Tests**
  - Tests moved to default configuration and rely on notification/drain-driven synchronization instead of timing sleeps.

### DIFF
--- a/crates/core/src/factories/core_fetch.rs
+++ b/crates/core/src/factories/core_fetch.rs
@@ -5,7 +5,6 @@ use std::sync::MutexGuard;
 use std::{
     collections::HashSet,
     sync::{Arc, Mutex},
-    time::Duration,
 };
 use tokio::{
     sync::mpsc::{channel, Receiver, Sender},
@@ -32,11 +31,6 @@ mod config {
         /// Default: 2.
         #[cfg_attr(feature = "schema", schemars(default))]
         pub parallel_request_count: u8,
-        /// Delay before re-inserting ops to request back into the outgoing request queue.
-        ///
-        /// Default: 30 s.
-        #[cfg_attr(feature = "schema", schemars(default))]
-        pub re_insert_outgoing_request_delay_ms: u32,
     }
 
     impl Default for CoreFetchConfig {
@@ -44,7 +38,6 @@ mod config {
         fn default() -> Self {
             Self {
                 parallel_request_count: 2,
-                re_insert_outgoing_request_delay_ms: 30000,
             }
         }
     }
@@ -186,6 +179,7 @@ impl Fetch for CoreFetch {
                         .map(|op_id| (op_id.clone(), source.clone())),
                 );
             }
+
             // Insert requests into fetch queue.
             for op_id in new_op_ids {
                 if let Err(err) =
@@ -256,12 +250,10 @@ impl CoreFetch {
             let request_task =
                 tokio::task::spawn(CoreFetch::outgoing_request_task(
                     state.clone(),
-                    outgoing_request_tx.clone(),
                     outgoing_request_rx.clone(),
                     space_id.clone(),
                     peer_meta_store.clone(),
                     Arc::downgrade(&transport),
-                    config.re_insert_outgoing_request_delay_ms,
                 ));
             tasks.push(request_task);
         }
@@ -308,12 +300,10 @@ impl CoreFetch {
 
     async fn outgoing_request_task(
         state: Arc<Mutex<State>>,
-        outgoing_request_tx: Sender<OutgoingRequest>,
         outgoing_request_rx: Arc<tokio::sync::Mutex<Receiver<OutgoingRequest>>>,
         space_id: SpaceId,
         peer_meta_store: DynPeerMetaStore,
         transport: WeakDynTransport,
-        re_insert_outgoing_request_delay: u32,
     ) {
         while let Some((op_id, peer_url)) =
             outgoing_request_rx.lock().await.recv().await
@@ -348,8 +338,7 @@ impl CoreFetch {
             // Do nothing if op id is no longer in the set of requests to send.
             //
             // If the peer URL is unresponsive, the current request will have been removed
-            // from state and no request will be sent and the request
-            // will not be re-inserted into the queue.
+            // from state and no request will be sent.
             {
                 let lock = state.lock().expect("poisoned");
                 if !lock.requests.contains(&(op_id.clone(), peer_url.clone())) {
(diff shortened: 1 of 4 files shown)


crates/api/src/fetch.rs
```diff
@@ -142,11 +141,8 @@ pub type DynFetchFactory = Arc<dyn FetchFactory>;
 /// Summary of the fetch state.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct FetchStateSummary {
-    /// The op ids that are currently being fetched.
-    ///
-    /// Each op id is associated with one or more peer URL from which the op data could be
-    /// requested.
-    pub pending_requests: HashMap<OpId, Vec<Url>>,
+    /// The number of op ids that are currently being fetched.
+    pub pending_request_count: usize,
```
jost-s, line 145: Which op ids are being fetched is no longer kept in state. I propose to only return the count of op ids that are being fetched. Otherwise I'd bring back the requests to the state.

ThetaSinner (dismissed):
This diff is rather difficult to follow because of code changing indentation where logic is being removed. As far as I can tell, this change is good.

crates/api/src/fetch.rs
matthme: I wonder if "pending request [...]" is still the correct name for this field and whether we should think about the value that's left in it.
If I understand the changes of this PR correctly, then this count is now only tracking what's currently in the outgoing request tokio channel but not requested from other peers yet. As soon as a request is being sent to another peer, the count is decreased so requests that are pending in the sense that we're still expecting a response in the incoming response channel are not represented by this number.

jost-s:
> This diff is rather difficult to follow because of code changing indentation where logic is being removed. As far as I can tell, this change is good.

Yes, the outgoing_request task is hard to make sense of. Overall it comes out simpler =D

crates/api/src/fetch.rs
matthme: > how many ops are discovered but not there yet

That this is interesting I agree but that's not what I see the current PR actually doing if it reduces the counter immediately after sending a request.

matthme (dismissed):
I just want to make sure this PR doesn't get merged before [this conversation](https://github.com/holochain/kitsune2/pull/338/files#r2399655953) isn't resolved because I don't currently see how this PR is tracking the same thing as before.

If we decide that it's okay to track something conceptually different then that's okay but I'm not sure we're clear about that yet (or I'm not understanding the code correctly).

jost-s:
> I just want to make sure this PR doesn't get merged before [this conversation](https://github.com/holochain/kitsune2/pull/338/files#r2399655953) isn't resolved because I don't currently see how this PR is tracking the same thing as before.
> 
> If we decide that it's okay to track something conceptually different then that's okay but I'm not sure we're clear about that yet (or I'm not understanding the code correctly).

PRs can't get merged with unresolved conversations in this repo.

crates/api/src/fetch.rs
jost-s: Yes, you're right, now it tracks something different. It needs to be tracked to notify listeners when the fetch request queue is drained.

matthme (commented):
Dismissing my "change requested" because I think I have brought my point across and it's now a weighing of trade-offs.

crates/api/src/fetch.rs
```diff
@@ -142,11 +141,8 @@ pub type DynFetchFactory = Arc<dyn FetchFactory>;
 /// Summary of the fetch state.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct FetchStateSummary {
-    /// The op ids that are currently being fetched.
-    ///
-    /// Each op id is associated with one or more peer URL from which the op data could be
-    /// requested.
-    pub pending_requests: HashMap<OpId, Vec<Url>>,
+    /// The number of op ids that are currently being fetched.
```
matthme, line 144:
```suggestion
    /// The number of op ids currently in the queue to be fetched.
```

crates/core/src/factories/core_fetch/test/outgoing_request_queue.rs
```diff
@@ -106,67 +103,10 @@ fn make_mock_transport(
     Arc::new(mock_transport)
 }
 
-#[tokio::test(flavor = "multi_thread")]
-async fn outgoing_request_queue() {
```
jost-s, line 110: Without the re-insertion logic, this test case is covered in the following happy path test.

crates/core/src/factories/core_fetch.rs
```diff
@@ -176,24 +169,34 @@ impl Fetch for CoreFetch {
...
+                        ?err,
+                        "could not insert fetch request into fetch queue"
                     );
+                    // Remove request from state.
+                    self.state
+                        .lock()
+                        .unwrap()
+                        .requests
+                        .remove(&(op_id, source.clone()));
                 }
```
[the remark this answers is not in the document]
jost-s: @coderabbitai Sharp as always - fixed!

matthme (approved):
Looks good to me now. I made a suggestion for an additional test if it's testable with reasonable effort.

crates/core/src/factories/core_fetch.rs
```diff
@@ -176,24 +169,32 @@ impl Fetch for CoreFetch {
...
+                    .outgoing_request_tx
+                    .send((op_id.clone(), source.clone()))
+                    .await
                 {
                     tracing::warn!(
-                        "could not insert fetch request into fetch queue: {err}"
+                        ?err,
+                        "could not insert fetch request into fetch queue"
                     );
+                    // Remove request from state.
```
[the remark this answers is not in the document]
jost-s: Done https://github.com/holochain/kitsune2/pull/338/commits/e25b7c60ea946ba0a1db3bfb9c8667bc1c611813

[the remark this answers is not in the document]
jost-s: Now that I look at this more closely, I think we can actually remove that part of the error handling and the test. `send` can only fail when the receiver is dropped, and that can only happen when the outgoing_request_task is dropped. Once that's dropped, the whole instance of the fetch module is unsalvageable.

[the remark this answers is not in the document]
mattyg: That makes sense to me, but then the tracing message above should be an error.

mattyg (approved):

mattyg (approved):

PR #349, opened 2025-10-10
refactor: remove tx5 test-utils dependency from kitsune crate

matthme:
Addresses the concern raised [in this comment](https://github.com/holochain/kitsune2/pull/320/files#r2334122648) by moving the sbd server logic into the blocks tests itself rather than importing the Tx5TestHarness.

## Summary by CodeRabbit

- Tests
  - Migrated test setups to a factory-based builder pattern with a shared helper for creating test environments.
  - Made tests more resilient by gracefully handling disconnects and adding retry/check logic to reduce CI flakes.
  - Consolidated peer creation and reuse across multiple message and block test scenarios.

- Chores
  - Removed an obsolete test-only dependency and its test feature configuration to streamline test builds.

- Documentation
  - No user-facing behavior changes.

### DIFF
--- a/crates/kitsune2/tests/blocks.rs
+++ b/crates/kitsune2/tests/blocks.rs
@@ -867,10 +867,25 @@ async fn incoming_notify_messages_from_blocked_peers_are_dropped() {
     // Alice sends a space message that should go through normally and establish
     // a connection with Bob
     let payload = Bytes::from("Hello world");
-    transport_alice
-        .send_space_notify(peer_url_bob.clone(), TEST_SPACE_ID, payload.clone())
-        .await
-        .unwrap();
+
+    // This send here fails relatively often in CI, presumably due to a race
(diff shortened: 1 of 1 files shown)


crates/kitsune2/tests/blocks.rs
```diff
@@ -147,6 +150,39 @@ impl TxHandler for TestTxHandler {
     }
 }
```
matthme: Runs an sbd server and creates a builder using the tx5 transport factory and with the appropriate signal server url. It returns the sbd server instance as well, despite never explicitly being used, since dropping it would stop the server.

jost-s (dismissed):

ThetaSinner (dismissed):

mattyg (dismissed):

jost-s (approved):

PR #353, opened 2025-10-31
feat: include peer_unresponsive timestamp in GossipStateSummary

mattyg:
Adds `peer_unresponsive` to GossipStateSummary, which is exposed in holochain's `dump_network_metrics` call. This is helpful in diagnosing behavior with unresponsive peers.

Putting this field in `GossipStateSummary` isn't really the right place as it's not specifically related to gossip behavior. I'm not sure if that's worth addressing here.


mattyg:
I"m thinking it makes sense to pull out the summary of peer meta store now, rather than kick it down the road. I will revise this PR to do so.

mattyg:
Closing as stale

PR #354, opened 2025-11-01
refactor: convert Timestamp to bytes directly when storing in PeerMetaStore for set_unresponsive created timestamp

mattyg:
Part of resolving https://github.com/holochain/holochain/issues/5376

Previously, `set_unresponsive` and `get_unresponsive` were using serde_json to convert a `Timestamp` to byte array and back.

This allowed for a bug downstream, where holochain attempted to decode the byte array using rmp_serde.

This PR implements `From` and `TryFrom` so there is a canonical way for Timestamp to be converted directly to `Bytes` and back. These are now used in `set_unresponsive` and `get_unresponsive`.

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal timestamp storage and conversion mechanisms for improved efficiency.


mattyg:
We're passing a bunch of other types directly as PeerMetaStore values that we don't control. Maybe it isn't the best approach to create wrappers for them.

PR #357, opened 2025-11-03
test: Improve flaky `two_node_gossip` test

ThetaSinner:
## Summary by CodeRabbit

* **Tests**
  * Improved reliability and observability of gossip synchronization by increasing timing tolerances, adjusting initiation pacing/jitter, extending wait limits, and removing a platform-specific startup delay.
* **Chores**
  * Unified and improved progress logging to structured, consistent output for better test diagnostics.

### DIFF
--- a/crates/kitsune2/tests/integration.rs
+++ b/crates/kitsune2/tests/integration.rs
@@ -82,9 +82,9 @@ async fn make_kitsune_node(
         .config
         .set_module_config(&K2GossipModConfig {
             k2_gossip: K2GossipConfig {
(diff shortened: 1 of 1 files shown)


crates/kitsune2/tests/integration.rs
```diff
@@ -187,12 +183,12 @@ async fn two_node_gossip() {
         {
             break;
         } else {
-            println!(
+            tracing::info!(
```
ThetaSinner: Use tracing rather than printing so I can see the progress it's making if this fails on CI.

PR #370, opened 2025-11-11
Fix/reduce hold on state lock

mattyg:
Reduce hold on state lock that was causing fetch requests to timeout.

## Summary by CodeRabbit

* **Chores**
  * Package version bumped to 0.3.1

* **Refactor**
  * Concurrency and state-handling rewritten to reduce lock scope, limit lock duration, and ensure notifications occur after state changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.3.0"
+version = "0.3.1"
 authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
 homepage = "https://www.holochain.org/"
 repository = "https://github.com/holochain/kitsune2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1636,7 +1636,7 @@ dependencies = [
 
 [[package]]
 name = "kitsune2"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "bytes",
  "jsonschema",
@@ -1655,7 +1655,7 @@ dependencies = [
 
 [[package]]
 name = "kitsune2_api"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "base64",
  "bytes",
@@ -1674,7 +1674,7 @@ dependencies = [
 
 [[package]]
 name = "kitsune2_bootstrap_client"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "axum",
  "axum-server",
@@ -1694,7 +1694,7 @@ dependencies = [
 
 [[package]]
 name = "kitsune2_bootstrap_srv"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "async-channel",
  "axum",
@@ -1724,7 +1724,7 @@ dependencies = [
 
 [[package]]
 name = "kitsune2_core"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "bytes",
  "ed25519-dalek",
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -16,7 +16,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.3.0"
+version = "0.3.1"
```
mattyg: This was never merged from `release-0.3` back into `main`. Will move to another PR.

crates/core/src/factories/core_fetch.rs
```diff
@@ -173,12 +172,15 @@ impl Fetch for CoreFetch {
...
-            );
+            {
+                let requests = &mut self.state.lock().unwrap().requests;
+                requests.extend(
+                    new_op_ids
+                        .clone()
+                        .into_iter()
+                        .map(|op_id| (op_id.clone(), source.clone())),
+                );
+            }
```
mattyg: This is functionally equivalent I believe but makes reading it clearer that we're holding a lock for this block only.

crates/core/src/factories/core_fetch.rs
```diff
@@ -379,19 +385,25 @@ impl CoreFetch {
...
+
                 // Remove all requests to that peer and notify if drained.
-                let mut lock = state.lock().expect("poisoned");
-                lock.requests.retain(|(_, a)| *a != peer_url);
-                Self::notify_listeners_if_queue_drained(lock);
+                {
+                    let mut lock = state.lock().expect("poisoned");
+                    lock.requests.retain(|(_, a)| *a != peer_url);
+                }
+                Self::notify_listeners_if_queue_drained(state.clone());
```
jost-s: This is not exactly the same, because there are multiple of these receivers processing outgoing fetch requests. It may be fine to change to this behavior, but it could also mean that we're seeing a change in gossip behavior in that it's not triggered as often during initial sync.
I'd prefer to keep the changes at a minimum while we're tying to find the issues.

crates/core/src/factories/core_fetch.rs
```diff
@@ -379,19 +385,23 @@ impl CoreFetch {
...
 
-    fn notify_listeners_if_queue_drained(mut state: MutexGuard<State>) {
+    fn notify_listeners_if_queue_drained(state: Arc<Mutex<State>>) {
         // Check if the fetch queue is drained.
-        if state.requests.is_empty() {
+        let mut lock = state.lock().expect("Poisoned");
+        if lock.requests.is_empty() {
             // Notify all listeners that the fetch queue is drained.
-            for notify in state.notify_when_drained_senders.drain(..) {
+            for notify in lock.notify_when_drained_senders.drain(..) {
```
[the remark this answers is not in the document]
matthme: I guess then we'd need to acquire the lock again in each iteration to remove the sender from the `Vec`. I don't know what's the trade-off here about acquiring a lock frequently vs. holding it for a while and get the messages sent more quickly instead.

crates/core/src/factories/core_fetch.rs
```diff
@@ -379,19 +386,25 @@ impl CoreFetch {
...
+                    let mut lock = state.lock().expect("poisoned");
+                    lock.requests.retain(|(_, a)| *a != peer_url);
+                }
+                Self::notify_listeners_if_queue_drained(state.clone());
             }
         }
     }
 
-    fn notify_listeners_if_queue_drained(mut state: MutexGuard<State>) {
+    fn notify_listeners_if_queue_drained(state: Arc<Mutex<State>>) {
```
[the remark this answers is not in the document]
ThetaSinner: The exact same amount of code is executing here though, because the lock must be held to drain the senders. This is just slightly less efficient with the two stage code above because it means acquiring the lock twice.

mattyg:
Closing as this does not seem related to the issue.

PR #392, opened 2025-11-26
Tidy up iroh testing

jost-s:
* Added unit tests for frame encoding and decoding
* Wrote (copied from tx5) test harness for integration tests and updated tests
* Moved transport integration tests to `tests/integration`, along with the test harness

### DIFF
--- a/crates/transport_iroh/src/tests/frame.rs
+++ b/crates/transport_iroh/src/tests/frame.rs
@@ -1,5 +1,39 @@
-use super::{decode_frame, FrameType, FRAME_HEADER_LEN, MAX_FRAME_BYTES};
+use super::{
+    decode_frame, encode_frame_header, FrameType, FRAME_HEADER_LEN,
+    MAX_FRAME_BYTES,
+};
 use bytes::Bytes;
+
+#[test]
+fn encode_frame_header_valid_peer_url() {
+    let ty = FrameType::PeerUrl;
+    let data_len = 100;
+    let header = encode_frame_header(ty, data_len).unwrap();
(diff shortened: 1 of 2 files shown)


crates/transport_iroh/src/tests/frame.rs
jost-s: New unit tests

crates/transport_iroh/src/tests/mod.rs
```diff
@@ -49,387 +46,3 @@ fn validate_plain_relay_url() {
...
-        Box::pin(async { Ok(()) })
-    }
-
-    fn are_all_agents_at_url_blocked(&self, _peer_url: &Url) -> K2Result<bool> {
-        Ok(false)
-    }
-}
-
-#[tokio::test]
-async fn connect_two_endpoints() {
```
jost-s: The two integration tests here are just adapted to use the harness and moved to the `integration.rs` file.

crates/transport_iroh/tests/harness.rs
```diff
@@ -0,0 +1,190 @@
...
+                Url::from_str("http://url.not.set:0/0").unwrap(),
+            )),
+        }
+    }
+}
+
+impl TxBaseHandler for MockTxHandler {
+    fn new_listening_address(&self, url: Url) -> BoxFut<'static, ()> {
+        (self.new_listening_address)(url.clone());
+        *self.current_url.lock().unwrap() = url;
```
jost-s: The handler has a `current_url` field which gets updated by default when a new address is assigned.

PR #411, opened 2025-12-16
Use persistent streams in iroh connections

### LINKED ISSUE #384: Implement connection cleanup and disconnect for iroh transport
## Follow-up work for iroh transport

As discussed in PR #382, the iroh transport implementation needs proper connection lifecycle management:

### Tasks

1. **Prune stale connections from the connections map** - Currently, closed connections remain in the map indefinitely, causing `get_connected_peers` to report stale peers.

2. **Implement proper `TxImp::disconnect` behavior** - The current implementation is a no-op. It should close the underlying iroh connection and clean up the connections map.

3. **Ensure `get_connected_peers` reflects actual connection state** - Verify that the returned list only includes peers with active connections.

4. Test that connection errors close the connection and leave the connections map tidy.

### Context

This work is needed to ensure correct peer lifecycle reporting and prevent issues with reconnection attempts to peers with stale connection contexts.

### References

- Original PR: #382
- Discussion: https://github.com/holochain/kitsune2/pull/382#discussion_r2554484569
- Requested by: @jost-s

### LINKED ISSUE #403: Use persistent uni-directional streams per peer connection instead of per message
Currently for every message a new uni-directional stream is created and destroyed, one for each side of a peer connection. Persist the streams instead in memory with the connection context. That'll ensure message ordering, which at the moment is not guaranteed.

jost-s:
resolves #403 
resolves #384 

## Summary by CodeRabbit

* **New Features**
  * Persistent uni‑directional streams with on‑demand send streams, explicit preflight handshake, remote URL accessor, get_connected_peers, and disconnect notifications.
  * Connection context lifecycle with automatic reader teardown and connection exposure.

* **Bug Fixes**
  * Atomic preflight state, handshake timeouts, stronger header validation, safer send/receive cleanup, and improved reconnect/disconnect handling.

* **Tests**
  * Expanded unit and integration tests covering header/preflight decoding and many reconnection scenarios.

### DIFF
--- a/crates/transport_iroh/src/connection_context.rs
+++ b/crates/transport_iroh/src/connection_context.rs
@@ -1,20 +1,31 @@
-use iroh::endpoint::{Connection, ConnectionType};
-use kitsune2_api::{TxImpHnd, Url};
+use crate::{
+    decode_frame_header, decode_frame_preflight,
+    frame::{encode_frame, Frame},
+    Connections, FrameType, FRAME_HEADER_LEN,
+};
+use bytes::Bytes;
+use iroh::endpoint::{Connection, ConnectionType, RecvStream, SendStream};
+use kitsune2_api::{K2Error, K2Result, TxImpHnd, Url};
 use n0_watcher::Watcher;
 use std::{
     fmt,
     sync::{
-        atomic::{AtomicU64, Ordering},
+        atomic::{AtomicBool, AtomicU64, Ordering},
         Arc, Mutex, RwLock,
     },
+    time::Duration,
 };
+use tokio::{sync::MutexGuard, task::AbortHandle};
+use tracing::{debug, error, info, trace};
 
 pub(super) struct ConnectionContext {
     handler: Arc<TxImpHnd>,
     connection: Arc<Connection>,
+    connection_reader_abort_handle: Mutex<Option<AbortHandle>>,
+    send_stream: tokio::sync::Mutex<Option<SendStream>>,
     remote_url: RwLock<Option<Url>>,
-    preflight_sent: RwLock<bool>,
-    preflight_received: RwLock<bool>,
+    preflight_sent: AtomicBool,
+    preflight_received: AtomicBool,
     send_message_count: AtomicU64,
     send_bytes: AtomicU64,
     recv_message_count: AtomicU64,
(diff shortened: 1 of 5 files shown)


crates/transport_iroh/tests/integration.rs
```diff
@@ -193,6 +194,408 @@ async fn send_and_receive_module_message() {
     .expect("message was not received by ep_2");
 }
 
+#[tokio::test]
+async fn reconnect_succeeds_after_connection_lost() {
```
jost-s: Added multiple tests for reconnecting after stream or connection error.

crates/transport_iroh/src/lib.rs
```diff
@@ -373,6 +355,213 @@ impl IrohTransport {
...
+        let (frame_type, data_len) = match decode_frame_header(&header) {
+            Ok(result) => result,
+            Err(err) => {
+                error!(?err, "failed to decode iroh frame header");
+                // If the frame header could not be decoded, continue by reading the next
+                // frame from the stream.
+                continue;
+            }
+        };
+        if frame_type == FrameType::Preflight {
```
jost-s: It is possible now to remove frame types altogether. Frames would just be decoded based on the convention that the first frame of a connection must be the preflight and every subsequent frame must be a data frame.

It's not a big overhead (1 byte bandwidth and little computation) to keep frame types and they are more expressive, but they are not essential any more.

Let me know what you think.

crates/transport_iroh/src/lib.rs
```diff
@@ -373,6 +355,215 @@ impl IrohTransport {
...
+/// # Parameters
+/// - `ctx`: The connection context containing handler and remote URL state.
+/// - `connections`: Shared map of peer URLs to their connection contexts, updated when
+///   the preflight succeeds.
+/// - `local_url`: The local URL for this endpoint, used to respond to preflight messages.
+fn spawn_connection_reader(
+    ctx: Arc<ConnectionContext>,
+    connections: Connections,
+    local_url: Arc<RwLock<Option<Url>>>,
+) {
```
ThetaSinner: No handle returned? Seems like something we should be able to stop as part of closing down a space?

ThetaSinner (commented):
This is looking good to me, but it's a big review to try to follow all of

crates/transport_iroh/src/connection_context.rs
```diff
@@ -140,4 +174,272 @@ impl ConnectionContext {
             None => ConnectionType::None,
         }
     }
+
+    /// Spawns an asynchronous task to continuously read and handle incoming uni-directional
```
jost-s: These connection and stream related functions have been moved from the lib file here. That's a slightly better encapsulation.

crates/transport_iroh/src/lib.rs
```diff
@@ -516,133 +503,3 @@ impl TxImp for IrohTransport {
         })
     }
 }
-
-/// Spawns an asynchronous task to continuously read and handle incoming unidirectional streams
```
jost-s: These function have been moved to `connection_context.rx`.

crates/transport_iroh/src/lib.rs
jost-s: Good catch! Fixed https://github.com/holochain/kitsune2/pull/411/changes/9b87c26911c1f9f591855ab56daa84dcc4d5f41e

veeso (approved):

ThetaSinner (approved):

PR #420, opened 2026-01-01
Chore: move feature tests to cargo make files

jost-s:
Move tests with different features from the root makefile to their respective crates.

Another change is that previously `cargo clippy` was only run from the root. Now it's run for every crate as well, with different features as configured in the makefiles. That prolongs the static run by a few minutes. If that's not desired, I'll revert that change.

## Summary by CodeRabbit

* **Chores**
  * Consolidated CI test steps into single test invocations per platform.
  * Added targeted lint and test variants for specific transport backends and delegated main tasks to focused sub-tasks.
  * Removed redundant test tasks and an unused dependency.

* **New Features**
  * Configuration examples/schema expanded to optionally include transport-specific settings guarded by feature flags.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -66,10 +66,7 @@ jobs:
       - name: Test
         env:
           RUST_LOG: info
-        run: |
-          cargo make test
-          cargo make test-transport-tx5-go-pion
-          cargo make test-transport-iroh
+        run: cargo make test
 
       - name: Examples
         run: cargo run --example schema --features schema
@@ -114,8 +111,6 @@ jobs:
         run: |-
           $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
           cargo make test
-          cargo make test-transport-tx5-go-pion
-          cargo make test-transport-iroh
 
   nix-build:
     needs: static
--- a/crates/kitsune2/Cargo.toml
+++ b/crates/kitsune2/Cargo.toml
@@ -72,5 +72,6 @@ schema = [
   "dep:serde_json",
   "kitsune2_core/schema",
   "kitsune2_gossip/schema",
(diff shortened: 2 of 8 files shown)


.github/workflows/test.yaml
```diff
@@ -66,10 +66,7 @@ jobs:
       - name: Test
         env:
           RUST_LOG: info
-        run: |
-          cargo make test
```
jost-s: No longer necessary, as all tests are now defined in their crates' makefiles.

crates/kitsune2/Cargo.toml
```diff
@@ -72,5 +72,5 @@ schema = [
   "dep:serde_json",
   "kitsune2_core/schema",
   "kitsune2_gossip/schema",
-  "kitsune2_transport_tx5/schema",
+  "kitsune2_transport_tx5?/schema",
```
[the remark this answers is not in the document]
jost-s: kitsune2 won't build with one of tx5 or iroh enabled as transport feature. But I've added the corresponding iroh transport feature. That makes sense now, doesn't it?

crates/kitsune2/Makefile.toml
```diff
@@ -1,19 +1,69 @@
...
+workspace = false
+dependencies = ["clippy-transport-tx5-backend-go-pion", "clippy-transport-iroh"]
+
+[tasks.clippy-transport-tx5-backend-go-pion]
+workspace = false
+command = "cargo"
+args = [
+  "clippy",
+  "--package",
+  "kitsune2",
```
ThetaSinner: Does this still need `--package kitsune2` now it's in the crate's makefile?

jost-s: I thought that yes, but double-checked with verbose output and no, it's not needed. Removed!

mattyg (approved):

PR #437, opened 2026-01-15
fix: race condition - URL not available before relay connection

lucksus:
# Waiting for the relay connection before completing initialization

While debugging ad4m integration with new Iroh transport, I saw calls to `TxImp::url` returning none because they happened to early in some cases. This wasn't the root cause of our issues. I don't know if it's fine without this fix in all cases, but it makes the initialization more deterministic and adds debug logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Transport initialization now waits (up to 30s) for the first listening URL to become available, failing with a clear timeout if not.
  * Added one-time signaling and improved logging around first URL readiness and subsequent address updates.

### DIFF
--- a/crates/transport_iroh/src/lib.rs
+++ b/crates/transport_iroh/src/lib.rs
@@ -218,10 +218,15 @@ impl IrohTransport {
         let connections = Arc::new(RwLock::new(HashMap::new()));
         let connection_locks = Arc::new(Mutex::new(HashMap::new()));
 
+        // Create a oneshot channel to wait for the first URL to be ready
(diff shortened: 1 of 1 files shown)


jost-s:
This could be a useful change. I would simplify the main connection task and pass in the local URL instead of an option.

But in the larger picture, this is something that kitsune should handle better. When a local URL is received from the relay, `new_listening_address` is called. I think kitsune shouldn't act before it doesn't have a listening address. Then this issue wouldn't occur.

Issue #424, opened 2026-01-05
[BUG] Kitsune2 blocking peers that it shouldn't


ThetaSinner:
See the PR https://github.com/holochain/kitsune2/pull/417 which demonstrates that the behavior is different when Kitsune2 instances connect and missing peers are permitted. This should only happen with a malicious remote who is trying to evade checks. So flipping that flag isn't a solution, it just indicates a problem.

Possible direction for investigation: Seen with flaky internet connections, so we might have to assume some operations failed while other succeeded.

AC:
- [ ] The reason for blocks being wrongly applied is identified
- [ ] It is no longer possible for peers to be treated as blocked when they aren't
- [ ] Regression test added, if possible

PR #451, opened 2026-01-27
fix: blocking race condition by preventing sends before local agent joins

### LINKED ISSUE #439: [SPIKE] Review peer-blocking code, identify and fix edge-cases
In https://github.com/holochain/kitsune2/issues/424 and https://github.com/holochain/kitsune2/pull/417 we have discussed and seen cases where messages get blocked when they shouldn't. While switching from tx5 to iroh transport seems to resolve the issue as a blocker for app-level data sharing as seen in ad4m/Flux, it is still not 100% clear what the blocking-code in Kitsune is doing wrong in the case of tx5 potentially having more issues with connection handling.

Also, we are still left with some cases of small numbers of blocked messages, in the beginning of a session between agents. While this not leading to app-level data sharing being interrupted, there is no clear explanation for why this happens.

Please do a full review of the related code, identify potential edge-cases, verify with tests and fix them.

lucksus:
## Summary

  Fixes #439 by preventing messages from being sent before a local agent has joined a space. This eliminates the race condition where empty preflights are sent to peers, causing them to block all subsequent messages from us.

## Problem

  A race condition exists in the space joining process:

  1. Space is created and bootstrap discovery starts
  2. A connection establishes and preflight exchanges before the local agent joins
  3. The preflight contains an empty agent list since no agents have joined yet
  4. When the local agent finally joins, remote peers never receive updated agent info
  5. Remote peers block all incoming messages due to having no access decision (default fail-safe behavior)

  This manifests as "small numbers of blocked messages occurring in the beginning of a session between agents" as described in #439.

## Previous Approach (PR #441)

  The initial fix in PR #441 attempted to solve this by resending preflights to all connected peers after a local agent joins. However, we determined this approach was too hacky, didn't address the root cause and is not an adequate use of pre-flight messages.

 ## New Approach

  This PR implements the solution we discussed yesterday: check if the local agent store is empty before allowing sends, and error if no agents exist.

### Key Changes

  1. Added has_local_agents() method to TxSpaceHandler trait
    - Checks if a space has any local agents joined
    - Default implementation returns true for backward compatibility
  2. Implemented in TxHandlerTranslator (CoreSpace)
    - Queries the local_agent_store to check for agents
    - Returns true if any agents exist, false otherwise
  3. Added validation in send_space_notify() and send_module()
    - Before sending, checks if the space has local agents via the space handler
    - Returns error: "Cannot send message before local agent has joined space"
    - This prevents empty preflights from ever being sent

### Consequences

  - Prevents the root cause: Messages cannot be sent before agents join, so empty preflights are never sent
  - Fast failure with clear errors: Developers get immediate, actionable feedback
  - No workarounds needed: No preflight resending, no changes to access decision logic
  - Maintains existing behavior: Default access decision semantics unchanged

### New Tests

  Added two comprehensive tests in crates/kitsune2/tests/blocks.rs:

  - send_before_local_agent_join_returns_error: Verifies that send_space_notify() fails before agent joins and succeeds after
  - send_module_before_local_agent_join_returns_error: Same verification for send_module()

  Both tests:
  1. Create a space without joining an agent
  2. Verify that sending returns an error with appropriate message
  3. Join a local agent
  4. Verify that sending now succeeds

### Updated Tests

  Modified existing block count tests to add local agents via local_agent_store().add() to satisfy the new requirement while avoiding broadcast side effects that would interfere with test assertions.

### Breaking Changes

  Potential breaking change: Code that attempts to send messages before calling local_agent_join() will now receive an error instead of silently sending empty preflights.

  This is the intended behavior and should be considered a bug fix rather than a breaking change, as the previous behavior was incorrect and led to blocked messages.

  Alternative Approaches Considered

  - ~~Resending preflights after agent join~~ (PR #441 - rejected as too hacky, i.e. abuse of pre-flight concept)
  - ~~Closing connections with empty preflights~~ (discussed but rejected - too disruptive)
  - ~~Changing default access decision behavior~~ (rejected - fail-safe default is correct)

  Related

  - Alternative approach: #441 (superseded by this PR)
  - Related issues: #424, #417

## Summary by CodeRabbit

* **New Features**
  * Preflight now requires a local agent in a space before allowing sends; operations return a clear error until one joins.

* **Bug Fixes**
  * Preflight failures due to missing local agents no longer mark peers as unresponsive, preventing spurious disconnects.

* **Tests**
  * Added tests for sending before a local agent joins and improved test harness/mocks to stabilize flaky scenarios.

### DIFF
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -567,10 +567,10 @@ impl DefaultTransport {
     ) -> K2Result<()> {
         let space_handler =
             self.space_map.lock().unwrap().get(&space_id).cloned();
-        if let Some(handler) = space_handler {
-            if !handler.has_local_agents().await? {
-                return Err(K2Error::NoLocalAgentsDuringPreflight);
-            }
+        if let Some(handler) = space_handler
+            && !handler.has_local_agents().await?
+        {
+            return Err(K2Error::NoLocalAgentsDuringPreflight);
         }
         Ok(())
     }


crates/kitsune2/tests/blocks.rs
```diff
@@ -1561,3 +1603,288 @@ async fn outgoing_module_messages_to_blocked_peers_are_dropped() {
...
+    );
+
+    // Now have Alice join with a local agent
+    let local_agent_alice = Arc::new(Ed25519LocalAgent::default());
+    space_alice
+        .local_agent_join(local_agent_alice.clone())
+        .await
+        .unwrap();
+
+    // Wait for the agent info to be published
```
mattyg:
```suggestion
    // Wait for the agent info to be added to the peer store
```

crates/kitsune2/tests/blocks.rs
```diff
@@ -1561,3 +1603,288 @@ async fn outgoing_module_messages_to_blocked_peers_are_dropped() {
...
+            {
+                tokio::time::sleep(std::time::Duration::from_millis(5)).await;
+            }
+        }
+    })
+    .await
+    .unwrap();
+
+    // Try to send again - should succeed (or at least not fail with "no local agent" error)
+    // Note: We wrap in iter_check because connection establishment may take time
```
mattyg: I don't think this needs to be an iter check. It should connect and send on the first call.

mattyg (approved):
LGTM, but I might still leave #439 open or else create new issues for the additional edge cases others noted in comments

jost-s:
> I think there is partial truth to both of your comments @jost-s and @ThetaSinner . I do think that @jost-s's scenario is hypothetically possible. But that scenario is not affected by the changes in this PR. It's an independent problem. Because this PR only touches _outgoing_ connections as @ThetaSinner is pointing out (the `send_space_notify` and `send_module` methods on the `DefaultTransport` imply an outgoing connection, if there is no existing connection yet). And your scenario @jost-s would be about returning an empty preflight in response to an _incoming_ connection attempt.

Yes, that is true. A won't mark B as unresponsive. But look what I think will happen:
- A joins space
- B opens connection to A and spawns a reader for incoming streams, awaiting the preflight response
- A receives incoming connection from B and spawns a reader for incoming streams too, awaiting B's preflight
- B as the connection initiator sends the preflight
- A receives the incoming preflight and therefore doesn't mark B as unresponsive. It doesn't return its own preflight, because the peer store is empty.
- B's incoming stream reader times out waiting for the preflight and marks A as unresponsive.

https://github.com/holochain/kitsune2/blob/main/crates/transport_iroh/src/connection_context.rs#L258

lucksus:
Correct @ThetaSinner, the failing test seems flaky and so far has only failed on macos and windows. Couldn't yet reproduce locally. Pushed potential fixes but my CI runs keep getting skipped. Is that one of you stopping runs in favor of other workflows?

ThetaSinner:
https://github.com/holochain/kitsune2/pull/451#issuecomment-3814151300

> B opens connection to A and spawns a reader for incoming streams, awaiting the preflight response

With the changes we'd implemented, how would B open a connection to A without a local agent present? The outgoing message would have been dropped before it lead to opening a connection because B doesn't have any local agents.

crates/api/src/transport.rs
```diff
@@ -731,6 +751,17 @@ pub trait TxHandler: TxBaseHandler {
...
+    /// Check if this handler has any local agents joined.
+    ///
+    /// This is used by the transport to prevent sending messages before
+    /// a local agent has joined, which would result in an empty preflight
+    /// being sent to peers, causing them to block all messages from us.
+    ///
+    /// The default implementation returns true (assumes agents exist).
+    fn has_local_agents(&self) -> BoxFut<'_, K2Result<bool>> {
+        Box::pin(async { Ok(true) })
+    }
```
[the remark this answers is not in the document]
lucksus: There are several mock implementations. That would be a relatively verbose change.

ThetaSinner (approved):
Looking good! Thank you for the perseverance on this one, it's been a challenge to figure out the right path to solving the problem. I'm excited to see this in a release now 🤞🏻

veeso (approved):

PR #462, opened 2026-02-11
fix: mark peer unresponsive when you fail to convert their url to an endpoint

### LINKED ISSUE #454: [BUG] tx5 address added to peer store in hc 0.6.1-rc.0
**Describe the bug**
When connecting Holochain 0.6.1-rc.0 to an existing operating network, where some nodes may still be on tx5 it looks like address found in the bootstrap server are getting into the PeerStore and they probably shouldn't be.  The logs become overloaded with errors like these from different parts of K2.

```
{"timestamp":"2026-01-23T17:43:27.826Z","label":"HOLOCHAIN 0.6.1-rc.0","level":"error","message":"2026-01-23T17:43:27.826248Z  WARN kitsune2_gossip::initiate: /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/kitsune2_gossip-0.4.0-dev.2/src/initiate.rs:129: Failed to initiate gossip: Other { ctx: \"failed to convert peer id to endpoint id\", src: Some(invalid length at 42) }"}
{"timestamp":"2026-01-23T17:43:27.842Z","label":"HOLOCHAIN 0.6.1-rc.0","level":"error","message":"2026-01-23T17:43:27.842508Z ERROR holochain::core::ribosome::real_ribosome: crates/holochain/src/core/ribosome/real_ribosome.rs:652: runtime_error=RuntimeError { source: Sys(User(WasmError { file: \"crates/holochain/src/core/ribosome/host_fn/get_links.rs\", line: 76, error: Host(\"failed to convert peer id to endpoint id (src: invalid length at 42)\") })), wasm_trace: [] } zome=Zome { name: ZomeName(\"zAuthorship\"), def: Wasm(WasmZome { wasm_hash: WasmHash(uhCokK7QflhSqzRPVpa92Wneho47CK7lggKn-hhzwgsLqRxQUGjuS), dependencies: [ZomeName(\"authorship_integrity\")] }) } fn_name=FunctionName(\"get_author\")"}
```

These entries continue to appear in the log, so holochain is continually trying to make these connections.

**Expected behavior**
Agents with tx5 addresses in the bootstrap server should be silently dropped and not be added to the Peer table.

**Steps to reproduce**
Easiest way is to use Moss dev version (as zippy for it) and connect to the public group.

AC:
- Mark peers unesponsive when the logged error above occurs.
- Ensure that any other places where a peer is contacted they will be marked unresponsive on error.

mattyg:
Resolves #454

## Summary by CodeRabbit

* **Bug Fixes**
  * Consistently mark peers as unresponsive when endpoint parsing, connection establishment, or preflight/send failures occur; surface contextual errors and close failed connections when appropriate.

* **Tests**
  * Added integration tests asserting peers are marked unresponsive for invalid peer addresses, connection timeouts, and preflight/send failures, with timing checks for unresponsive notifications.

* **Chores**
  * Updated CI cache target name in workflow configuration.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -128,7 +128,7 @@ jobs:
 
       - uses: cachix/cachix-action@v16
         with:
-          name: holochain-wind-tunnel
+          name: holochain-ci
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
       - run: nix build --verbose -L .#bootstrap-srv


crates/transport_iroh/src/lib.rs
```diff
@@ -650,7 +666,22 @@ impl TxImp for IrohTransport {
                         local_url.clone(),
                         &self.config,
                     )
-                    .await?;
+                    .await
+                    {
+                        Err(e) => {
+                            // On connection error, mark the peer unresponsive
```
mattyg: Additionally, mark peer as unresponsive if we fail to establish a connection with them.

mattyg: I could, although there would be 3 places to handle the errors within that function. I don't feel strongly.

mattyg: Updated to handle the 2 errors directly within `create_connection_and_context` that are related to network responses. Also modified the tests to cover each separately.

PR #477, opened 2026-03-11
fix: separate known-peers index from peer store to fix blocking for shared URLs

### LINKED ISSUE #450: Problems and limitations in the current blocking implementation
The current implementation of blocking (as of commit e9c75f2d2356fac02959db2295a20f820c702aef) contains an inherent contradiction in that the implementation on the one hand assumes that the peer store is kept clean of blocked agents so as to not initiate gossip with them (or return them as available peers via helper functions like [get_responsive_remote_agents_near_location](https://github.com/holochain/kitsune2/blob/e9c75f2d2356fac02959db2295a20f820c702aef/crates/core/src/common.rs#L86) used by the kitsune host), while at the same time assuming that agent infos in the peer store can be used to associate peer urls to agent ids of potentially blocked agents. However, if agent infos of blocked agents are not supposed to be in the peer store in the first place (i.e. [not getting inserted](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/api/src/peer_store.rs#L10-L12) as well as [getting removed when blocked](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/api/src/blocks.rs#L29-L30)), then the peer store can't actually meaningfully be used to [associate peer urls to agent agent infos](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/core/src/factories/core_access.rs#L61-L62) and then check [whether the agents of the associated agent infos are blocked](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/core/src/factories/core_access.rs#L86), because the blocked agents won't ever be returned from the peer store in the first place.

The reason this issue didn't show up in tests is that the access decision [gets removed if no agent for a peer url is found in the peer store](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/core/src/factories/core_access.rs#L77-L84). And this happens as part of calling `block()` on the `Blocks` module which [removes the agent from the peer store](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/kitsune2/tests/blocks.rs#L466-L467), thereby invoking the peer update listener that has been registered by the `CoreAccessState`. And if no access decision is found for a peer url, [it is considered blocked](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/core/src/factories/core_space.rs#L277). But this will only work if there is only a single blocked agent trying to connect on a given peer url and zero non-blocked ones. Otherwise [the list of agents won't be empty here](https://github.com/holochain/kitsune2/blob/e72edc14fa2017f9ab9f5608d5562a0ede4580e3/crates/core/src/factories/core_access.rs#L77-L84) in the peer update listener and we won't find a blocked peer in the peer store because the one that is blocked has deliberately not been inserted into the peer store.

This means that we are currently still effectively implementing "only if *all* agents associated with a peer url are blocked, messages from that peer  url get blocked" rather than "if *any* agent associated with a peer url is blocked, messages from that peer url get blocked", as can for example be (and has been) verified by modifying [this blocks test](https://github.com/holochain/kitsune2/blob/e9c75f2d2356fac02959db2295a20f820c702aef/crates/kitsune2/tests/blocks.rs#L920) by adding an additional local agent to Alice's conductor on [this line](https://github.com/holochain/kitsune2/blob/e9c75f2d2356fac02959db2295a20f820c702aef/crates/kitsune2/tests/blocks.rs#L943) alongside the agent that will [get blocked](https://github.com/holochain/kitsune2/blob/e9c75f2d2356fac02959db2295a20f820c702aef/crates/kitsune2/tests/blocks.rs#L991).

Associated [MM thread](https://chat.holochain.org/holo/pl/q633w6ifapdy7npo5qqd5ns8ia).

And a hackmd providing more context about the broader problem space: https://hackmd.io/RpkAEte7TVOsM6cyjDhf-A

synchwire:
Closes #450

## Problem

The blocking implementation had a fundamental design contradiction:

- Blocked agents are excluded from the `PeerStore` (not inserted; removed on block) to prevent gossip with them
- `CorePeerAccessState` used `PeerStore::get_by_url` to map peer URLs → agent IDs in order to decide whether to block incoming connections

These assumptions are mutually incompatible. When a blocked agent is removed from the peer store, the URL lookup can no longer find them. If a non-blocked agent shares the same peer URL, the lookup finds the non-blocked agent, sees no block, and incorrectly grants access — allowing the blocked agent's traffic through.

**Net result:** blocking only worked when *all* agents at a peer URL were blocked, not when *any* were.

## Solution

Introduce a new `KnownPeers` trait in the `api` crate: an append-only index of all ever-seen agent infos, including blocked ones, used purely for URL → agent ID resolution by the access control layer.

- `KnownPeers::record` — records agent infos without any block filtering
- `KnownPeers::get_by_url` — returns all known agents at a URL regardless of block status

`CorePeerAccessState` now uses `DynKnownPeers` for URL lookups instead of `DynPeerStore`. Agent infos are recorded into `KnownPeers` before the block filter runs, so blocked agents remain findable for access control purposes even after being removed from the peer store.

A new integration test verifies the fixed behaviour: a blocked agent sharing a peer URL with a non-blocked agent is correctly blocked.

## Release Notes

* **New Features**
  * Added a known-peers index for independent URL-to-agent-ID resolution that persists even when peers are blocked or filtered from the peer store, enabling more robust peer discovery and access control decisions.

### DIFF
--- a/crates/api/src/known_peers.rs
+++ b/crates/api/src/known_peers.rs
@@ -1,0 +1,54 @@
+//! Known-peers index types.
+//!
+//! The [`KnownPeers`] trait provides an append-only index of all ever-seen
+//! agent infos, including those that are currently blocked.  It is used
+//! purely for URL → agent-ID resolution by the access control layer so that
+//! a blocked agent cannot slip through just because it shares a URL with a
+//! non-blocked agent.
+
+use crate::*;
+use std::sync::Arc;
+
+/// An append-only index of all ever-seen agent infos, including blocked ones.
+///
+/// Unlike [`PeerStore`] this store never removes entries on blocking – its
+/// only purpose is to let the access control layer answer the question
+/// "which agents have we ever seen at this URL?".
+pub trait KnownPeers: 'static + Send + Sync + std::fmt::Debug {
+    /// Record agent infos without any block filtering.
+    ///
+    /// This should be called with every batch of agent infos *before* any
+    /// block filter is applied, so that blocked agents remain discoverable
+    /// by URL.
+    fn record(
+        &self,
+        agent_infos: Vec<Arc<AgentInfoSigned>>,
+    ) -> BoxFut<'_, K2Result<()>>;
+
+    /// Look up all known agent IDs reachable at the given URL, regardless of
+    /// their current block status.
+    fn get_by_url(&self, url: Url) -> BoxFut<'_, K2Result<Vec<AgentId>>>;
+}
+
+/// Trait-object [`KnownPeers`].
+pub type DynKnownPeers = Arc<dyn KnownPeers>;
+
+/// A factory for constructing [`KnownPeers`] instances.
+pub trait KnownPeersFactory: 'static + Send + Sync + std::fmt::Debug {
+    /// Help the builder construct a default config from the chosen
+    /// module factories.
+    fn default_config(&self, config: &mut Config) -> K2Result<()>;
+
+    /// Validate configuration.
+    fn validate_config(&self, config: &Config) -> K2Result<()>;
+
+    /// Construct a [`KnownPeers`] instance.
+    fn create(
+        &self,
+        builder: Arc<Builder>,
+        space_id: SpaceId,
+    ) -> BoxFut<'static, K2Result<DynKnownPeers>>;
+}
+
+/// Trait-object [`KnownPeersFactory`].
+pub type DynKnownPeersFactory = Arc<dyn KnownPeersFactory>;
--- a/crates/api/src/peer_store.rs
+++ b/crates/api/src/peer_store.rs
@@ -81,11 +81,18 @@ pub trait PeerStoreFactory: 'static + Send + Sync + std::fmt::Debug {
     fn validate_config(&self, config: &Config) -> K2Result<()>;
 
     /// Construct a peer store instance.
+    ///
+    /// Implementations **must** call [`KnownPeers::record`] with every
+    /// batch of agent infos *before* applying any block or expiry
+    /// filtering.  This ensures the access control layer can always
+    /// resolve a peer URL back to its agent IDs, even for agents that
+    /// have been blocked and removed from the peer store.
     fn create(
         &self,
         builder: Arc<Builder>,
         space_id: SpaceId,
         blocks: DynBlocks,
+        known_peers: DynKnownPeers,
     ) -> BoxFut<'static, K2Result<DynPeerStore>>;
 }
 
--- a/crates/core/src/factories/core_access.rs
+++ b/crates/core/src/factories/core_access.rs
@@ -1,6 +1,6 @@
 use kitsune2_api::{
-    AccessDecision, BlockTarget, DynBlocks, DynPeerStore, K2Result, PeerAccess,
-    PeerAccessState, Timestamp, Url,
+    AccessDecision, BlockTarget, DynBlocks, DynKnownPeers, K2Result,
+    PeerAccess, PeerAccessState, Timestamp, Url,
 };
 use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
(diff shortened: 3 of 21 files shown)


crates/api/src/known_peers.rs
```diff
@@ -0,0 +1,57 @@
...
+        &self,
+        agent_infos: Vec<Arc<AgentInfoSigned>>,
+    ) -> BoxFut<'_, K2Result<()>>;
+
+    /// Look up all known agents reachable at the given URL, regardless of
+    /// their current block status.
+    fn get_by_url(
+        &self,
+        url: Url,
+    ) -> BoxFut<'_, K2Result<Vec<Arc<AgentInfoSigned>>>>;
```
ThetaSinner: Do we need to have stored the entire `AgentInfoSigned`? They're small but not that small.

crates/api/src/peer_store.rs
```diff
@@ -86,6 +86,7 @@ pub trait PeerStoreFactory: 'static + Send + Sync + std::fmt::Debug {
         builder: Arc<Builder>,
         space_id: SpaceId,
         blocks: DynBlocks,
+        known_peers: DynKnownPeers,
```
ThetaSinner, line 95: The expectations for peer store implementations should be documented. Where we have integrations between modules, we surface them in documentation to make it easier to implement replacement modules as required.

crates/core/src/factories/core_bootstrap/test.rs
```diff
@@ -96,9 +96,15 @@ impl Test {
...
+            .create(builder.clone(), S1.clone())
+            .await
+            .unwrap();
+
         let peer_store = builder
             .peer_store
-            .create(builder.clone(), S1.clone(), blocks)
+            .create(builder.clone(), S1.clone(), blocks, known_peers)
             .await
             .unwrap();
```
ThetaSinner, line 109: Should the space hold a reference to this new module too? For functional testing, I'd expect to be able to retrieve a handle to this module.

crates/core/src/factories/core_access.rs
```diff
@@ -167,3 +175,126 @@ impl PeerAccessState for CorePeerAccessState {
...
+            known_peers,
+        ))
+    }
+
+    /// Regression test: blocking one agent at a URL must block the URL even
+    /// when a non-blocked agent also resides at that URL.
+    ///
+    /// Before the `KnownPeers` fix the blocked agent was removed from the
+    /// `PeerStore`, so `get_by_url` only saw the non-blocked agent and
+    /// returned `Granted` instead of `Blocked`.
```
ThetaSinner: Don't document past states, focus on what the current state of the code is and why.

crates/core/src/factories/core_access.rs
```diff
@@ -167,3 +175,126 @@ impl PeerAccessState for CorePeerAccessState {
...
+        // at the same URL, the URL must be Blocked.
+        let decision = access_state.get_access_decision(url.clone()).unwrap();
+        assert_eq!(
+            decision.map(|d| d.decision),
+            Some(AccessDecision::Blocked),
+            "expected Blocked after agent_1 at the shared URL was blocked"
+        );
+    }
+
+    /// Sanity test: a URL with only non-blocked agents remains Granted.
```
ThetaSinner: Confusing wording, this is just a happy path case. Skip the "sanity" wording and also "stays". Just "is".

crates/core/src/factories/mem_known_peers.rs
```diff
@@ -0,0 +1,101 @@
...
+use tokio::sync::Mutex;
+
+/// A production-ready memory-based known-peers factory.
+///
+/// This stores agent infos in an in-memory hash map by [`AgentId`], indexed
+/// by [`Url`].  Entries are never removed – the whole point of this store is
+/// to retain the URL → agent mapping even after an agent has been blocked and
+/// removed from the [`crate::factories::MemPeerStoreFactory`].
+#[derive(Debug)]
+pub struct MemKnownPeersFactory {}
```
ThetaSinner: If it's a production one, we should name it `CoreKnownPeersFactory` and not `Mem` prefix. We haven't been consistent about this but that's the intended naming pattern.

crates/api/src/known_peers.rs
synchwire: Changed. `get_by_url` now returns `Vec<AgentId>` and the internal storage keeps only `AgentId → Option<Url>` pairs instead of full `AgentInfoSigned` structs.

crates/api/src/peer_store.rs
synchwire: Added documentation on `PeerStoreFactory::create` explaining that implementations must call `KnownPeers::record` with every batch of agent infos before applying block or expiry filtering.

crates/core/src/factories/core_bootstrap/test.rs
synchwire: Added `fn known_peers(&self) -> &DynKnownPeers` to the `Space` trait. `CoreSpace` now stores and exposes the reference.

crates/core/src/factories/core_access.rs
synchwire: Removed. The test doc now just states the invariant being tested.

crates/core/src/factories/core_access.rs
synchwire: Fixed. Now reads: "A URL with only non-blocked agents is Granted."

crates/core/src/factories/mem_known_peers.rs
synchwire: Renamed to `CoreKnownPeersFactory` / `CoreKnownPeers` and moved the file to `core_known_peers.rs`. Updated all references across the codebase.

jost-s (approved):
Looks great!

A concern is that this is an in-memory store that only grows while kitsune is running. We might have to solve that if it becomes an issue.

PR #524, opened 2026-04-16
fix: stabilize iroh relay in bootstrap_srv

### LINKED ISSUE #491: Stabilize the Iroh relay hosted in the `bootstrap_srv`
The experience using the relay hosted in the `bootstrap_srv` crate has been less good than using the public relay that Iroh host. That experience has been conflated with running with authentication enabled.

Use the new Dino Adventure release a v0.2.1-rc.1 to test with Holochain 0.6.1-rc.4 and Kitsune2 0.4.0-rc.6 to diagnose problems with this relay or confirm that it works without authentication enabled.

This could be just a case of testing the relay at <https://dev-test-bootstrap2-iroh.holochain.org> with authentication disabled. The followup issue [HC-655](https://linear.app/hcf/issue/HC-655/stabilize-the-iroh-relay-hosted-in-the-bootstrap-srv-with) will then fix the issue with authentication.

AC:

- [ ] Relay is confirmed working at [https://dev-test-bootstrap2-iroh.holochain.org](<https://dev-test-bootstrap2-iroh.holochain.org>) or remaining issues in Kitsune2 are resolved.

veeso:
## Summary

Stabilize the self-hosted Iroh relay in `bootstrap_srv` by fixing bugs found during investigation, improving observability, and adding benchmarks/probes for ongoing comparison against upstream.

Closes #491

### Changes

- **Handshake timeout** (`iroh_relay_axum.rs`): Wrap handshake+auth in 30s `tokio::time::timeout` to prevent stalled clients from holding WebSocket connections indefinitely.
- **QAD fix** (`transport_iroh/src/lib.rs`): Replace `RelayConfig { url, quic: None }` with `RelayConfig::from(relay_url)` on the authenticated `insert_relay` path, enabling QUIC Address Discovery that was silently disabled.
- **Error formatting** (`iroh_relay_axum.rs`): Replace `Debug` formatting with `Display` in 5 `AxumWebSocketAdapter` error sites for human-readable messages.
- **Relay tracing** (`iroh_relay_axum.rs`): Upgrade `trace!` to `debug!` for handshake/auth, add `info!` for client registration, classify expected errors (timeout, disconnect, denied auth) as `debug!` instead of `warn!`.
- **Criterion benchmarks** (`benches/iroh_relay_bench.rs`): Local relay throughput (1/8/32 KiB), local + external relay roundtrip for direct comparison.
- **Relay probe test** (`tests/iroh_relay_probe.rs`): Ignored opt-in test for one-shot relay delivery verification against any relay URL.

### Benchmark results (from dev machine, Europe/Rome)

**Local relay throughput (reused clients, in-process relay):**

| Payload | Latency  | Throughput  |
|---------|----------|-------------|
| 1 KiB   | ~22 µs   | ~44 MiB/s   |
| 8 KiB   | ~25 µs   | ~318 MiB/s  |
| 32 KiB  | ~36 µs   | ~862 MiB/s  |

**Relay roundtrip comparison (connect + handshake + 1 KiB datagram, fresh clients per iteration):**

| Relay                                    | Roundtrip |
|------------------------------------------|-----------|
| Local (in-process, localhost)             | ~1.9 ms   |
| n0 EU (euc1-1.relay.n0.iroh-canary)      | ~237 ms   |
| Deployed (dev-test-bootstrap2-iroh)       | ~230 ms   |

The deployed Holochain relay performs comparably to n0's EU public relay (~230 ms vs ~237 ms). The ~1.9 ms local baseline confirms that relay overhead is dominated by network latency, not our code.

### Code comparison analysis (self-hosted vs upstream iroh-relay)

| Dimension         | Finding                                                  |
|-------------------|----------------------------------------------------------|
| Rate limiting     | **Gap (medium-high)**: upstream has per-client token-bucket rate limiting; we have none |
| Metrics export    | **Gap (low-medium)**: `Metrics` struct populated but not exported to OTel |
| TLS key export    | **Gap (low)**: `ExportKeyingMaterial` returns `None` (inherent to axum approach) |
| Everything else   | Equivalent — we reuse upstream `Clients`/`Client` actor directly |

Rate limiting and metrics export tracked as follow-up issues.

### Relay validation

- Probe test passed against `dev-test-bootstrap2-iroh.holochain.org` (64 KiB payload, ~5s)
- Criterion benchmark passed against deployed relay (~230ms roundtrip)
- Health endpoints `/ping` (200) and `/generate_204` (204) confirmed responding
- **Note**: App-level validation with Dino Adventure / Holochain was not performed. Relay-protocol-level delivery is confirmed working; request app-level testing from someone with Holochain infra access.

## Test plan

- [x] `cargo make static` passes
- [x] `cargo test -p kitsune2_bootstrap_srv` — 35 tests pass
- [x] `cargo test -p kitsune2_transport_iroh` — all tests pass
- [x] `cargo test -p kitsune2` — 11 integration tests pass
- [x] `cargo bench -p kitsune2_bootstrap_srv` — local benchmarks pass
- [x] External relay benchmarks pass against n0 EU + deployed relay
- [ ] App-level test with Holochain + Dino Adventure against deployed relay (needs infra access)

### DIFF
--- a/crates/bootstrap_srv/benches/iroh_relay_bench.rs
+++ b/crates/bootstrap_srv/benches/iroh_relay_bench.rs
@@ -1,0 +1,297 @@
+//! Criterion benchmarks for relay message delivery.
+//!
+//! Benchmarks use the `iroh_relay::client::ClientBuilder` API directly
+//! (not `iroh::Endpoint`) to measure raw relay throughput without the
+//! overhead of QUIC negotiation, holepunching, or QAD.
+//!
+//! **Local relay** benchmarks always run against an in-process axum relay
+//! server — no network access required.
+//!
+//! **External relay** benchmarks are gated behind the
+//! `KITSUNE2_IROH_RELAY_BENCH_URLS` environment variable (comma-separated
+//! list of relay URLs). If unset, only local benchmarks run.
+//!
+//! # Examples
+//!
+//! ```sh
+//! # Local only:
+//! cargo bench -p kitsune2_bootstrap_srv
+//!
+//! # Compare against public and deployed relays:
+//! KITSUNE2_IROH_RELAY_BENCH_URLS="https://use1-1.relay.n0.iroh-canary.iroh.link./,https://dev-test-bootstrap2-iroh.holochain.org" \
+//!   cargo bench -p kitsune2_bootstrap_srv
+//! ```
+
+use std::{net::Ipv4Addr, time::Duration};
+
+use axum::{Router, routing::get};
+use criterion::{
+    BenchmarkId, Criterion, Throughput, criterion_group, criterion_main,
+};
+use futures::{SinkExt, StreamExt};
+use iroh_base::{RelayUrl, SecretKey};
+use iroh_relay::{
+    client::ClientBuilder,
+    dns::DnsResolver,
+    protos::relay::{ClientToRelayMsg, Datagrams, RelayToClientMsg},
+};
+use kitsune2_bootstrap_srv::iroh_relay_axum::{
+    create_relay_state, relay_handler,
+};
+use tokio::{net::TcpListener, runtime::Runtime};
+
+/// Spawn an in-process axum relay server on a dedicated thread.
+fn spawn_local_relay() -> RelayUrl {
+    let (tx, rx) = std::sync::mpsc::channel();
+    std::thread::spawn(move || {
+        let rt = Runtime::new().unwrap();
+        rt.block_on(async {
+            let state = create_relay_state();
+            let app = Router::new()
+                .route("/relay", get(relay_handler))
+                .with_state(state);
+
+            let listener =
+                TcpListener::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
+            let addr = listener.local_addr().unwrap();
+            tx.send(addr).unwrap();
+            axum::serve(listener, app).await.expect("server error");
+        });
+    });
+    let addr = rx.recv().unwrap();
+    format!("http://{}/relay", addr).parse().unwrap()
+}
+
+type ClientPair = (
+    iroh_relay::client::Client,
+    iroh_base::EndpointId,
+    iroh_relay::client::Client,
+    iroh_base::EndpointId,
+);
+
+/// Connect two relay clients to the given relay URL.
+async fn connect_client_pair(relay_url: &RelayUrl) -> ClientPair {
+    let resolver = DnsResolver::new();
+
+    let a_secret = SecretKey::generate();
+    let a_key = a_secret.public();
+    let client_a =
+        ClientBuilder::new(relay_url.clone(), a_secret, resolver.clone())
+            .connect()
+            .await
+            .unwrap();
+
+    let b_secret = SecretKey::generate();
+    let b_key = b_secret.public();
+    let client_b = ClientBuilder::new(relay_url.clone(), b_secret, resolver)
+        .connect()
+        .await
+        .unwrap();
+
+    (client_a, a_key, client_b, b_key)
+}
+
+fn local_relay_benchmarks(c: &mut Criterion) {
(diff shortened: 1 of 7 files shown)


crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -224,19 +235,34 @@ async fn handle_relay_websocket(
...
+            .await?;
+
+        Ok::<_, Box<dyn std::error::Error + Send + Sync>>(client_key)
+    })
+    .await
+    .map_err(|_| {
+        std::io::Error::new(
+            std::io::ErrorKind::TimedOut,
+            "Relay handshake timed out",
+        )
```
ThetaSinner, line 275: This looks like we're somehow handling the timeout error before the code where the timeout can happen? Should the async block return `Ok(client_key)` and then the mapping of the timeout error happens below after it has happened where the `??` is now?

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -91,7 +92,19 @@ pub async fn relay_handler(
         if let Err(e) =
             handle_relay_websocket(socket, state, client_auth_header).await
         {
-            warn!("Error handling relay WebSocket: {:?}", e);
+            let msg = e.to_string();
+            let is_expected = msg.contains("timed out")
```
ThetaSinner, line 109: Maybe a comment about where these values were sourced from? It's not too serious if they get out of sync because it's just the log level but string matching is generally something to be cautious with so having a hint about how to trace these down in a new version could be useful?

crates/bootstrap_srv/benches/iroh_relay_bench.rs
ThetaSinner, line 1: This looks fine, just a couple of general comments:
- It seems a bit strange to be benchmarking the Iroh relay when it's third party software - but given that we've embedded it, I can see why it makes sense to have something to check it handles load okay.
- These should be getting run regularly to stop them going stale. For this PR, could you double check that the Makefile.toml will build these benchmarks at some point perhaps? Then I'd be fine with a follow up issue, or doing it on this PR, to get CI to run benchmarks across the repo - then we'll run any that are added in future. That shouldn't be too complicated with `cargo-make` I hope.

ThetaSinner (commented):
Looking good, a few minor comments.

I'm wondering if this also closes https://github.com/holochain/kitsune2/issues/526? If so, that could be referenced in the "Closes" of the PR description too

crates/bootstrap_srv/src/iroh_relay_axum.rs
veeso: Uhm, it looks correct to me:

`timeout(...).await` resolves to `Result<Result<T, BoxErr>, Elapsed>`. `.map_err` converts outer `Elapsed` to `io::Error` after timeout fires.

veeso:
@ThetaSinner No, #526 doesn't get closed because it requires further implementations.

crates/bootstrap_srv/benches/iroh_relay_bench.rs
veeso: https://github.com/holochain/kitsune2/issues/530

crates/bootstrap_srv/src/iroh_relay_axum.rs
jost-s: Should this lower case the `msg` first to make the comparison a bit more robust to changes? Spelling looks quite arbitrary in the log messages.

jost-s (dismissed):

crates/bootstrap_srv/src/iroh_relay_axum.rs
ThetaSinner: Got it! Yeah I'm not keeping track of brackets properly with the squished two-pane review. If I expand the display I can see it's right.

ThetaSinner (approved):

PR #527, opened 2026-04-20
chore: update Rust to 1.95

ThetaSinner:


.github/workflows/test.yaml
```diff
@@ -110,6 +110,7 @@ jobs:
       - name: Test
         run: |-
           $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
+          $env:RUST_LOG="info"
```
ThetaSinner, line 113: This has been too noisy for quite a while, especially when tests do actually fail

jost-s (approved):

PR #533, opened 2026-04-30
feat(bootstrap_srv): rate-limit inbound bytes on embedded iroh relay

veeso:
## Summary

Adds per-connection inbound byte rate limiting to the embedded iroh relay server in `bootstrap_srv`. Closes #501.

## How it works

A token bucket lives inside the axum WebSocket adapter for each connection. Every inbound binary frame's size is charged to the bucket *after* the frame is read but *before* the next frame is polled. If a charge overruns, the next read is paced via `tokio::time::sleep_until(deadline)` where `deadline` comes from the bucket itself. No frames are dropped; every frame is delivered, only paced.

The bucket primitive is `iroh_relay::Bucket`, which is `pub(crate)` upstream. It is consumed here via `[patch.crates-io]` pinning the `holochain/iroh` fork (branch `kitsune2/expose-bucket`, SHA `4002ec17`) which flips the visibility and re-exports the type. Same semantics as Iroh's own `RateLimited`, applied at the frame layer instead of the raw byte stream, which is required because axum terminates the WebSocket upgrade before the relay handler runs.

## What's done

- `iroh_relay::Bucket` exposed on the holochain/iroh fork.
- `[patch.crates-io]` pinning all of `iroh-base`, `iroh-dns`, `iroh-relay` from the fork.
- `RelayClientRxRateLimit` config type plumbed from `Config` through `RelayState` into `AxumWebSocketAdapter`.
- `Config::resolve_relay_rate_limit()` normalises the (bps, burst) pair; burst defaults to `bps/10` when only the sustained rate is set. `BootstrapSrv::new` rejects `burst` without `bps`.
- CLI flags `--relay-client-rx-bytes-per-second`, `--relay-client-rx-burst-bytes`.
- Off by default in both `Config::testing()` and `Config::production()`.

### DIFF
--- a/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
+++ b/crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
@@ -120,6 +120,28 @@ pub struct Args {
     /// Default port is 7842 (iroh's default for QAD).
     #[arg(long)]
     pub quic_bind_addr: Option<std::net::SocketAddr>,
+
+    /// Sustained inbound byte rate per relay client connection (bytes/s).
+    ///
+    /// When set, the bootstrap server's embedded iroh relay handler will
+    /// pace inbound WebSocket frames per connection. Use together with
+    /// `--relay-client-rx-burst-bytes` to control burst allowance; if
+    /// burst is omitted it defaults to about one tenth of this value.
+    ///
+    /// Default: unlimited.
+    #[cfg(feature = "iroh-relay")]
+    #[arg(long)]
+    pub relay_client_rx_bytes_per_second: Option<std::num::NonZeroU32>,
+
+    /// Maximum allowed inbound burst per relay client connection (bytes).
+    ///
+    /// Has effect only when `--relay-client-rx-bytes-per-second` is also
+    /// set.
+    ///
+    /// Default: derived as bps/10 when the sustained rate is set.
+    #[cfg(feature = "iroh-relay")]
+    #[arg(long)]
+    pub relay_client_rx_burst_bytes: Option<std::num::NonZeroU32>,
 }
 
 fn main() {
@@ -213,6 +235,12 @@ fn main() {
     {
         if let Some(quic_bind_addr) = args.quic_bind_addr {
             config.quic_bind_addr = Some(quic_bind_addr);
+        }
+        if let Some(bps) = args.relay_client_rx_bytes_per_second {
+            config.relay_client_rx_bytes_per_second = Some(bps);
+        }
+        if let Some(burst) = args.relay_client_rx_burst_bytes {
+            config.relay_client_rx_burst_bytes = Some(burst);
         }
     }
 
--- a/crates/bootstrap_srv/src/iroh_relay_axum.rs
+++ b/crates/bootstrap_srv/src/iroh_relay_axum.rs
@@ -27,11 +27,29 @@ use iroh_relay::{
     protos::{handshake, streams::StreamError},
     server::{
         Access, AccessControl, AllowAll, ClientRequest, DynAccessControl,
-        Metrics, client::Config, clients::Clients, streams::RelayedStream,
+        Metrics,
+        client::Config,
+        clients::Clients,
+        streams::{Bucket, RelayedStream},
     },
 };
+use std::future::Future;
 
 use crate::RelayAllowlist;
+
+/// Per-connection inbound byte rate limit for the relay WebSocket handler.
+///
+/// The limiter applies to every binary frame received from a single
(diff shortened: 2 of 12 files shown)


crates/bootstrap_srv/src/bin/kitsune2-bootstrap-srv.rs
```diff
@@ -120,6 +120,26 @@ pub struct Args {
     /// Default port is 7842 (iroh's default for QAD).
     #[arg(long)]
     pub quic_bind_addr: Option<std::net::SocketAddr>,
+
+    /// Sustained inbound byte rate per relay client connection (bytes/s).
+    ///
+    /// When set, the bootstrap server's embedded iroh relay handler will
+    /// pace inbound WebSocket frames per connection. Use together with
+    /// `--relay-client-rx-burst-bytes` to control burst allowance; if
+    /// burst is omitted it defaults to ~one tenth of this value.
```
jost-s:
```suggestion
    /// burst is omitted it defaults to about one tenth of this value.
```

crates/bootstrap_srv/Cargo.toml
```diff
@@ -48,8 +48,12 @@ sbd-server = { workspace = true, features = ["tungstenite"], optional = true }
 opentelemetry = { workspace = true }
 # iroh relay server-related dependencies
 iroh = { workspace = true, optional = true }
-iroh-base = { workspace = true, optional = true }
-iroh-relay = { workspace = true, features = [
+# iroh-base and iroh-relay are pulled from the holochain/iroh fork to expose
+# the `Bucket` rate-limit primitive required by the embedded relay handler.
+# Scoped to this crate (rather than via a workspace [patch.crates-io]) so that
+# transport_iroh keeps using the vanilla crates.io iroh-base/iroh-relay.
+iroh-base = { git = "https://github.com/holochain/iroh", rev = "4002ec174335f30a2e012674b960822e33ec3079", optional = true }
```
jost-s: The relay server can't be released with a branch dependency. Please release the forked iroh first.

jost-s (dismissed):
Changes look good to me. Two valid rabbit suggestions IMO.

Requesting to release the iroh fork for the bootstrap to be releasable.

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -589,4 +684,174 @@ mod tests {
...
+    async fn rate_limit_state_construction_safe_for_low_rates() {
+        // bytes_per_second = 100 over a 100ms refill window resolves to
+        // a per-period refill of 10 bytes (>0), so Bucket::new accepts
+        // it. Validates that small-but-nonzero rates do not panic.
+        let limit = RelayClientRxRateLimit {
+            bytes_per_second: NonZeroU32::new(100).unwrap(),
+            burst_bytes: NonZeroU32::new(10).unwrap(),
+        };
+        let _ = RateLimitState::new(limit);
+    }
```
ThetaSinner, line 871: Slightly unusual test. If something can error, I'd usually suggest returning an error and asserting `action.is_ok()` or something like that. Is the lack of a panic okay to just inspect off the code here or is the test important?

ThetaSinner (commented):
Code looks good, just one high-level question. The rate limiting is currently optional - if you don't specify flags then the resolve method on the config will return `None`?

Does that behavior match Iroh? Or should we return a default value from the resolve and have the rate limit be non-optional? 

I would lean towards non-optional unless Iroh have done it the same way as this.

veeso:
@ThetaSinner implemented behaviour matches iroh's behaviour: `server::Limits.client_rx` is `Option<ClientRateLimit>` and defaults to `None`, on `ClientRateLimit` the `bytes_per_second` is required while `max_burst_bytes` is `Option`. resolve_relay_rate_limit returning None when no flags are set mirrors Iroh's default, which is `no limiter`.

crates/bootstrap_srv/src/iroh_relay_axum.rs
veeso: it actually doesn't return error; it just panics if you pass limit 0. I removed the `let _` to avoid wrong comprehension of the code though.

ThetaSinner: I understood it doesn't return an error, I was questioning whether the panic should be an error return. That makes for cleaner testing and users can choose how to handle the error

ThetaSinner (approved):

crates/bootstrap_srv/src/iroh_relay_axum.rs
veeso: Panics can be raised on things that get constructed at launch, so I would keep as-is. I've added a test for panic.

PR #535, opened 2026-05-04
ci: run cargo bench across the workspace on push to main

### LINKED ISSUE #530: CI: run cargo bench across the workspace on a regular schedule
## Context

We have benchmarks in the workspace (currently `crates/bootstrap_srv/benches/iroh_relay_bench.rs`) and we expect more to be added over time. Right now nothing in CI actually runs them, so they could silently break or stop being meaningful without anyone noticing.

`cargo make static` was updated to compile benches via `cargo bench --no-run`, which catches build-time bit-rot. That's a good first step but it doesn't tell us anything about the numbers themselves.

## Goal

Add a CI job that runs `cargo bench` across the workspace so we get fresh results and notice if a benchmark starts failing or drifting wildly.

Run the benchmark after the merge to main.

## Notes

- Some benchmarks may be gated behind environment variables (for example `KITSUNE2_IROH_RELAY_BENCH_URLS`); CI should run the unset / local-only path.
- Suggested follow-up to PR #524.

veeso:
## Summary

- Add a `Bench` workflow that runs `cargo bench` across the workspace on every push to `main` (and on manual dispatch), tracks results via `benchmark-action/github-action-benchmark`, and fails the run if any benchmark drifts more than 150% versus the previous run. History is persisted on the `bench-history` branch.
- Fix the `iroh_relay_bench` in `kitsune2_bootstrap_srv` so it can actually run: pass an explicit TLS client config to `ClientBuilder` (iroh-relay 0.98 requires this even when targeting a local `http://` / `ws://` relay) and enable the `tls-aws-lc-rs` + `test-utils` features on the `iroh-relay` dev-dependency.

Closes #530

## Test plan

- [x] Local: `cargo bench -p kitsune2_bootstrap_srv --bench iroh_relay_bench -- --output-format bencher` succeeds and emits bencher-format output.
- [x] CI: `Bench` workflow runs end-to-end on the branch, including the `Store benchmark result` step pushing to `bench-history`.
- [ ] After merge: confirm `Bench` runs automatically on the push to `main`.

### DIFF
--- a/.github/workflows/bench.yaml
+++ b/.github/workflows/bench.yaml
@@ -1,0 +1,78 @@
+name: Bench
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch: {}
+
+permissions:
+  contents: write
+  deployments: write
+
+concurrency:
+  group: bench-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  bench:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Maximize build space (Linux)
+        uses: AdityaGarg8/remove-unwanted-software@v5
+        with:
+          remove-dotnet: "true"
+          remove-android: "true"
+          remove-codeql: "true"
+          remove-docker-images: "true"
+
+      - uses: Swatinem/rust-cache@v2
+
+      # we cannot use cmake 4 until openssl-src is updated
+      - name: cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: "3.31.x"
+
+      - name: Run benchmarks and write results to file
+        env:
+          RUST_LOG: info
+        shell: bash
+        run: |
+          set -euo pipefail
+          : > bench-output.txt
+          cargo metadata --format-version 1 --no-deps \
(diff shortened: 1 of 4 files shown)


.github/workflows/bench.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Maximize build space (Linux)
+        uses: AdityaGarg8/remove-unwanted-software@v5
+        with:
+          remove-dotnet: 'true'
+          remove-android: 'true'
+          remove-codeql: 'true'
+          remove-docker-images: 'true'
+
+      - name: Go Toolchain
```
jost-s: I think Go was only needed for the tx5 Go pion backend. Shouldn't be needed for iroh.

.github/workflows/bench.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+      - uses: Swatinem/rust-cache@v2
+
+      # we cannot use cmake 4 until openssl-src is updated
+      - name: cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.31.x'
+
+      - name: Bench
```
jost-s:
```suggestion
      - name: Run benchmarks and write results to file
```

jost-s (commented):
Please attach an example of the artifact that the workflow stores. I'd like to review the format.

veeso:
@jost-s https://github.com/holochain/kitsune2/tree/bench-history

jost-s (approved):
Looks great! Just the workflows are still failing.

PR #545, opened 2026-05-26
fix(transport_iroh): don't mark peers unresponsive when relay is unreachable

synchwire:
## Summary

When our relay connection drops (e.g. Android doze mode), outbound gossip connections time out after 60 s and falsely mark reachable peers as unresponsive for up to ~15 minutes. The failure is ours, not the peer's.

## Fix

iroh 1.0.0-rc.0's `home_relay_status()` is driven directly by the relay actor's state machine — `Disconnected` fires immediately on TCP drop, unlike `watch_addr` which is subject to net-report hysteresis and won't reflect relay loss.

- **`Endpoint::is_home_relay_connected()`** — new internal trait method backed by `home_relay_status().get().iter().any(|s| s.is_connected())`. Returns `false` at startup, during reconnect, and on drop.
- **`create_connection_and_context`** — guard checks `is_home_relay_connected()` before touching the network; returns early without calling `set_unresponsive` if `false`.
- **`TestBootstrapSrv`** — now calls `Clients::shutdown()` on drop, closing relay WebSocket connections immediately so tests don't have to wait 60 s for QUIC idle timeout.

## Tests

- **Unit** (`relay_disconnect.rs`): `RelayDownEndpoint` with `is_home_relay_connected() = false` and a hanging `connect()`. Guard returns in < 1 ms; zero `set_unresponsive` calls.
- **Integration** (`no_unresponsive_when_relay_drops`): real transport + live relay, server dropped, polls until guard fires, asserts `set_unresponsive` never called.

## Limitation

Peer-side relay loss (relay up, peer sleeping) still causes 60 s timeouts and unresponsive marks — separate problem requiring backoff or OS network signals.

### DIFF
--- a/crates/test_utils/src/bootstrap.rs
+++ b/crates/test_utils/src/bootstrap.rs
@@ -10,7 +10,6 @@ use std::sync::{Arc, Mutex};
 /// iroh relay endpoints at `/relay` and `/ping`.
 pub struct TestBootstrapSrv {
     kill: Option<tokio::sync::oneshot::Sender<()>>,
-    task: tokio::task::JoinHandle<std::io::Result<()>>,
     halt: Arc<std::sync::atomic::AtomicBool>,
     addr: String,
     #[cfg(feature = "relay")]
@@ -19,10 +18,13 @@ pub struct TestBootstrapSrv {
 
 impl Drop for TestBootstrapSrv {
     fn drop(&mut self) {
+        // Sending the kill signal causes the server task to call
+        // clients.shutdown(), which closes all relay WebSocket connections.
+        // This lets connected iroh endpoints detect relay loss immediately
+        // rather than waiting for the QUIC idle timeout.
         if let Some(kill) = self.kill.take() {
             let _ = kill.send(());
         }
-        self.task.abort();
     }
 }
 
@@ -122,13 +124,30 @@ impl TestBootstrapSrv {
             Some(relay_state)
         };
 
-        let task = tokio::task::spawn(std::future::IntoFuture::into_future(
-            serve(l, app).with_graceful_shutdown(kill_r),
-        ));
+        // Clone the relay clients handle so the task can shut down connections
+        // on drop.  Closing the connections lets connected iroh endpoints
+        // detect relay loss promptly (rather than waiting for QUIC idle
+        // timeout), which is essential for tests that simulate relay outages.
+        #[cfg(feature = "relay")]
+        let relay_clients = relay_state.as_ref().map(|s| s.clients.clone());
+
+        tokio::task::spawn(async move {
+            // Stop accepting when the kill signal fires.
+            tokio::select! {
+                _ = kill_r => {},
+                result = serve(l, app) => { let _ = result; }
+            }
+
+            // Gracefully close all relay WebSocket connections so iroh
+            // endpoints see the relay as disconnected immediately.
+            #[cfg(feature = "relay")]
+            if let Some(clients) = relay_clients {
+                clients.shutdown().await;
+            }
+        });
 
         Self {
             kill,
-            task,
             halt,
             addr,
             #[cfg(feature = "relay")]
--- a/crates/transport_iroh/src/lib.rs
+++ b/crates/transport_iroh/src/lib.rs
@@ -238,6 +238,15 @@ pub mod test_utils;
 mod tests;
 
 const ALPN: &[u8] = b"kitsune2/0";
+/// Error message returned when a connection attempt is skipped because the
+/// home relay is not connected.  Exported so integration tests can match it
+/// without depending on a free-form string literal.
+#[cfg(any(test, feature = "test-utils"))]
+pub const RELAY_NOT_CONNECTED_ERR: &str =
+    "relay not connected, skipping to avoid false unresponsive mark";
+#[cfg(not(any(test, feature = "test-utils")))]
+pub(crate) const RELAY_NOT_CONNECTED_ERR: &str =
+    "relay not connected, skipping to avoid false unresponsive mark";
 
 /// IrohTransport configuration types
 pub mod config {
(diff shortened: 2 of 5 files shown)


crates/test_utils/src/bootstrap.rs
```diff
@@ -22,7 +23,10 @@ impl Drop for TestBootstrapSrv {
         if let Some(kill) = self.kill.take() {
             let _ = kill.send(());
         }
-        self.task.abort();
+        // Do not abort the task here: the kill signal above causes the task to
+        // run clients.shutdown(), which closes all relay WebSocket connections
+        // so that connected iroh endpoints can detect the relay loss promptly.
+        // Aborting immediately would cancel that cleanup before it runs.
```
ThetaSinner: You're documenting the code change and not the current state, this should be a comment on the `kill.send` that explains how it calls `clients.shutdown()`.

crates/test_utils/src/bootstrap.rs
```diff
@@ -10,7 +10,8 @@ use std::sync::{Arc, Mutex};
 /// iroh relay endpoints at `/relay` and `/ping`.
 pub struct TestBootstrapSrv {
     kill: Option<tokio::sync::oneshot::Sender<()>>,
-    task: tokio::task::JoinHandle<std::io::Result<()>>,
+    #[allow(dead_code)]
+    task: tokio::task::JoinHandle<()>,
```
ThetaSinner: If unused why not remove?

crates/transport_iroh/src/tests/connect_failure.rs
```diff
@@ -284,6 +288,9 @@ async fn marks_unresponsive_when_outer_connect_timeout_fires() {
         fn id_bytes(&self) -> [u8; 32] {
             [0u8; 32]
         }
+        fn is_home_relay_connected(&self) -> bool {
```
ThetaSinner: Missing newline before new function

crates/transport_iroh/src/tests/relay_disconnect.rs
```diff
@@ -0,0 +1,193 @@
...
+use kitsune2_api::{
+    BoxFut, DefaultTransport, K2Result, TransportStats, TxImp, TxImpHnd, Url,
+};
+use kitsune2_test_utils::space::TEST_SPACE_ID;
+use n0_watcher::Disconnected;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, RwLock};
+use std::time::Duration;
+
+// ─── minimal stubs ──────────────────────────────────────────────────────────
```
ThetaSinner: Shouldn't need to document sections - at most, docs on types where needed or skip if clear enough from type names and context.

crates/transport_iroh/src/tests/relay_disconnect.rs
```diff
@@ -0,0 +1,193 @@
...
+/// Without the guard in `create_connection_and_context`, this hanging connect
+/// would block for `connect_timeout_s` and then falsely mark the peer as
+/// unresponsive.
+#[derive(Debug)]
+struct RelayDownEndpoint;
+
+impl Endpoint for RelayDownEndpoint {
+    fn watch_addr(&self) -> Box<dyn EndpointAddrWatcher> {
+        Box::new(PendingWatcher)
+    }
```
ThetaSinner: Functions should have a newline between them, even in test code

crates/transport_iroh/src/tests/relay_disconnect.rs
ThetaSinner, line 1: How valuable is this code? It could be a single connection attempt in the integration test but this requires several new types to achieve the same thing. I like unit tests but I'm thinking this may not be justified?

crates/transport_iroh/src/lib.rs
```diff
@@ -812,6 +812,21 @@ impl IrohTransport {
...
+        if !self.endpoint.is_home_relay_connected() {
+            warn!(
+                ?remote_url,
+                "skipping outbound connection: relay not connected, \
+                 peer will not be marked unresponsive"
+            );
+            return Err(K2Error::other(
+                "relay not connected, skipping to avoid false unresponsive mark",
+            ));
+        }
```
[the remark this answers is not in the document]
synchwire: Fixed — changed to `debug!`.

crates/test_utils/src/bootstrap.rs
synchwire: Fixed — comment now describes the current drop behaviour rather than the change that introduced it.

crates/test_utils/src/bootstrap.rs
synchwire: Removed the field and changed the spawn to fire-and-forget.

crates/transport_iroh/src/tests/connect_failure.rs
synchwire: Fixed — added blank line.

crates/transport_iroh/src/tests/relay_disconnect.rs
synchwire: Removed — section header comments dropped.

crates/transport_iroh/src/tests/relay_disconnect.rs
synchwire: Fixed — blank lines between all method impls.

crates/transport_iroh/src/tests/relay_disconnect.rs
synchwire: The unit test gives sub-millisecond, deterministic coverage of the guard itself: it verifies the `if !is_home_relay_connected()` branch fires correctly without spinning up a real relay. The integration test validates the full relay-drop → actor-state → guard → no-unresponsive chain but is 3–4 s and timing-sensitive. The two layers catch different failure modes — e.g. the guard being removed vs iroh not reporting state correctly.

Happy to drop the unit test if you'd prefer integration-only coverage. The CodeRabbit comment about stub duplication also applies; consolidating into a shared helper module would reduce the type count if the unit test stays.

synchwire: Done — dropped the unit test and `relay_disconnect.rs` entirely. Integration test covers the end-to-end case; the stubs are gone with it so the duplication concern is resolved too.

jost-s (approved):

PR #547, opened 2026-05-29
feat: add metadata from publish through the fetch queue to the op store

### LINKED ISSUE #539: Carry host-supplied metadata from publish through the fetch queue to the op store
## Problem

Hosts (Holochain) need to attach arbitrary metadata to a publish and receive it alongside the op when it is delivered to the op store. Kitsune2 today has no path for this: `Publish::publish_ops` carries only op ids, the fetch queue tracks only `(OpId, Url)`, and `OpStore::process_incoming_ops` only sees raw op bytes.

The metadata should be opaque bytes — no host-specific generics in the API. Breaking changes are acceptable on `main` under the current dev-release policy.

## Approach

Thread per-op `Option<Bytes>` metadata along publish → fetch queue → op store.

### Wire format

`crates/api/proto/publish.proto`:

```proto
message PublishOps {
  repeated PublishOpEntry entries = 1;
}

message PublishOpEntry {
  bytes op_id = 1;
  optional bytes metadata = 2;
}
```

`crates/api/proto/op_store.proto` — `Op` gains `op_id` so a `FetchResponse` can be paired with metadata cached in the fetch queue. The fetch module is fire-and-forget with no per-request correlation, and only the host can compute op ids from data.

```proto
message Op {
  bytes op_id = 1;
  bytes data = 2;
}
```

### Rust API

```rust
pub struct PublishOp {
    pub op_id: OpId,
    pub metadata: Option<Bytes>,
}

pub struct IncomingOp {
    pub op_id: OpId,        // from the wire; host should validate
    pub op_data: Bytes,
    pub metadata: Option<Bytes>,
}

trait Publish {
    fn publish_ops(&self, ops: Vec<PublishOp>, target: Url) -> BoxFut<'_, K2Result<()>>;
}

trait Fetch {
    fn request_ops(&self, ops: Vec<PublishOp>, source: Url) -> BoxFut<'_, K2Result<()>>;
}

trait OpStore {
    fn process_incoming_ops(
        &self,
        op_list: Vec<IncomingOp>,
    ) -> BoxFuture<'_, K2Result<Vec<OpId>>>;
}
```

Gossip-flow callers of `request_ops` wrap their op ids as `PublishOp { op_id, metadata: None }`. `Report::fetched_op` is unchanged.

### Fetch queue state

`core_fetch::State::requests` becomes `HashMap<(OpId, Url), Option<Bytes>>`. Insertion rule:

- New key: insert as given.
- Existing `None`: upgrade to `Some(_)` if the new value is `Some(_)`.
- Existing `Some(_)`: never overwrite.

On response, the incoming-response task pairs each wire `(op_id, op_data)` with the metadata at `(op_id, peer)` and calls `process_incoming_ops`.

### Size cap

New `CorePublishConfig::max_metadata_bytes` (default 1024). Enforced on outgoing (`Publish::publish_ops` returns `K2Error`) and on incoming (`PublishMessageHandler` drops oversized entries with `warn!`, keeps the rest of the batch).

## Touchpoints

- `crates/api`: `proto/publish.proto`, `proto/op_store.proto`, `src/publish.rs`, `src/fetch.rs`, `src/op_store.rs`.
- `crates/core`: `factories/core_publish.rs` (+ `message_handler.rs`), `factories/core_fetch.rs`, `factories/mem_op_store/*`.
- `crates/gossip`: `request_ops` and `process_incoming_ops` call sites in `respond/*` and `gossip.rs`; `harness/op_store.rs`.
- Tests: new coverage for metadata round-trip, the upgrade/overwrite rules, oversize rejection on both ends, and gossip delivering `None`; one integration test in `crates/kitsune2`.

cdunster:
The PR adds the ability for hosts (such as Holochain) to attach arbitrary metadata to ops when publishing them. This metadata is passed through the wire protocol and the fetch queue, and is then processed with the op data. The metadata has a configurable maximum length where any outgoing publishes with oversized metadata are rejected with an error and any received ops with oversized metadata are dropped with a warning, keeping the rest of the batch.

> [!Warning]
> This contains braking changes to the protobuf definitions

Closes #539.


crates/core/src/factories/core_fetch.rs
```diff
@@ -161,24 +159,38 @@ impl CoreFetch {
...
+                    let key = (op_id.clone(), source.clone());
+                    lock.requests
+                        .entry(key)
+                        .and_modify(|existing| {
+                            if existing.is_none() {
+                                *existing = meta.clone();
+                            }
+                        })
+                        .or_insert(meta);
+                }
```
jost-s, line 188: Maybe a bit simpler
```suggestion
                for op_id in &new_op_ids {
                    let meta = metadata_map.remove(op_id).flatten();
                    let entry = lock.requests.entry((op_id.clone(), source.clone())).or_insert(None);
                    if entry.is_none() {
                        *entry = meta;
                    }
                }
```

jost-s (dismissed):
LGTM

crates/core/src/factories/core_fetch.rs
cdunster: Looks good, I used `or_default()` instead of `or_insert(None)` but it was done in c1ea843219bb.

jost-s (approved):

PR #548, opened 2026-06-03
chore: update iroh dependency to latest rc.1 release

cdunster:
Also make the required changes to be compatible with `v1.0.0-rc.1`

### DIFF
--- a/crates/bootstrap_srv/src/iroh_relay_axum.rs
+++ b/crates/bootstrap_srv/src/iroh_relay_axum.rs
@@ -21,16 +21,17 @@ use std::{
 use tokio::time::timeout;
 use tracing::{debug, info, trace, warn};
 
-use futures::FutureExt;
 use iroh_relay::http::ProtocolVersion;
 use iroh_relay::{
     ExportKeyingMaterial, KeyCache,
     protos::{handshake, streams::StreamError},
     server::{
-        Access, AccessConfig, ClientRequest, Metrics, client::Config,
-        clients::Clients, streams::RelayedStream,
+        Access, AccessControl, AllowAll, ClientRequest, DynAccessControl,
+        Metrics, client::Config, clients::Clients, streams::RelayedStream,
     },
 };
+
+use crate::RelayAllowlist;
 
 /// State required for the relay handler.
 #[derive(Clone, Debug)]
@@ -38,7 +39,7 @@ pub struct RelayState {
     /// Key cache for the relay
     pub key_cache: KeyCache,
     /// Access control configuration
-    pub access: Arc<AccessConfig>,
+    pub access: Arc<dyn DynAccessControl>,
     /// Metrics for the relay server
     pub metrics: Arc<Metrics>,
     /// Write timeout for client connections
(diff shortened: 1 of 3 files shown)


crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -38,7 +37,7 @@ pub struct RelayState {
     /// Key cache for the relay
     pub key_cache: KeyCache,
     /// Access control configuration
-    pub access: Arc<AccessConfig>,
+    pub access: Arc<dyn DynAccessControl>,
```
cdunster, line 42: This is the new way of setting up access control, see: https://github.com/n0-computer/iroh/pull/4276

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -387,11 +388,39 @@ pub async fn captive_portal_handler(
...
+        tracing::debug!(
+            key = %endpoint_id.fmt_short(),
+            allowed,
+            allowlist_size = self.allowlist.len(),
+            "Relay access check"
+        );
+        if allowed {
+            Access::Allow
+        } else {
+            Access::Deny { reason: None }
```
cdunster, line 427: Not sure if we should add a reason here but I can't think of one other than "Access not allowed" which isn't really helpful.

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -387,11 +388,39 @@ pub async fn captive_portal_handler(
...
+    allowlist: crate::RelayAllowlist,
+}
+
+impl std::fmt::Debug for AllowlistAccess {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AllowlistAccess").finish()
+    }
+}
+
+impl AccessControl for AllowlistAccess {
```
cdunster, line 414: We could also just `impl AccessControl for RelayAllowlist` directly which would remove this wrapper struct but then it would move the `on_connect` logic to the `relay_allowlist` module.

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -21,24 +21,25 @@ use std::{
...
+use crate::RelayAllowlist;
+
 /// State required for the relay handler.
 #[derive(Clone, Debug)]
 pub struct RelayState {
     /// Key cache for the relay
     pub key_cache: KeyCache,
     /// Access control configuration
-    pub access: Arc<AccessConfig>,
+    pub access: Arc<dyn DynAccessControl>,
```
ThetaSinner, line 42: Is the type not already a `dyn` form?

ThetaSinner (approved):

crates/bootstrap_srv/src/iroh_relay_axum.rs
cdunster: Confusingly no. It's a `dyn`-compatible trait, see: https://github.com/n0-computer/iroh/blob/9a79782eec54273b34cf0a04c4d627a27a9a2df3/iroh-relay/src/server.rs#L306-L320

Basically it's just a wrapper around the actual trait but it returns boxed futures instead of `impl Future`.

jost-s (approved):

PR #560, opened 2026-06-23
refactor: remove tx5

### LINKED ISSUE #542: Remove tx5 support
Now that we have our tooling working with Iroh and we're committed enough to it, we can remove support for tx5 from Kitsune2. That will happen in the 0.5.x release of Kitsune2 which will ship with Holochain 0.7.x

ThetaSinner:
Closes https://github.com/holochain/kitsune2/issues/542

I ran into a reasonably serious issue as part of this change. The shutdown tests I'd created previously to prove that shutting down a space released all tasks were *only* being run for tx5. That was a leftover from the Iroh migration. It turns out that test was failing because our usage of Iroh really was keeping tasks around. When both sides initiate a connection, only one connection is retained on each side but the reader task stays alive. What that actually means I think, is that we were effectively keeping two Iroh connections open if that happened. Much less likely in real usage over unit tests but it was worth fixing I think.


crates/transport_iroh/src/connection_context.rs
```diff
@@ -219,31 +301,41 @@ impl ConnectionContext {
...
+                            }
+                            Err(err) => {
+                                error!(?err, "Stream closed by remote");
+                                // Don't mark peer as unresponsive for NoLocalAgentsDuringPreflight
+                                // errors - this is a temporary state that will resolve once an
+                                // agent joins.
+                                if matches!(err, K2Error::NoLocalAgentsDuringPreflight) {
+                                    skip_unresponsive = true;
+                                }
+                                break err.to_string();
```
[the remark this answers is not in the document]
ThetaSinner: Addressed in e91379a (fixup targeting 52f54c3). The `NoLocalAgentsDuringPreflight` arm now logs at `debug!` ("Stream closed during preflight; no local agents yet"), and the `else` branch reserves `error!` for genuine preflight failures. This matches our logging conventions — transient, self-resolving conditions should not be logged at `error!`.

Note the nearby `error!("Connection closed by remote")` on the next arm is pre-existing (predates this branch), so I left it out of scope for this PR.

ThetaSinner:
Addressing the CodeRabbit nitpick on `crates/api/src/config.rs` (`test_should_merge_config_override`, multi-module merge coverage) — done in e91379a.

I took a slightly different approach than suggested. Instead of setting a second module on *both* the source and override configs, I added a second module (`OtherModuleModConfig`) to the **source only** and assert it persists unchanged after the merge. The overrides touch only the bootstrap module, so this directly exercises the "override one module without dropping another" semantic. Setting an identical module on both sides would not catch a merge that drops source-only modules, since `merge_overrides` iterates over the *override* keys and never removes keys absent from them.

crates/kitsune2/tests/blocks.rs
```diff
@@ -987,11 +935,6 @@ async fn incoming_notify_messages_from_blocked_peers_are_dropped() {
     // a connection with Bob
     let payload = Bytes::from("Hello world");
 
-    // This send here fails relatively often in CI, presumably due to a race
-    // condition in tx5: https://github.com/holochain/tx5/issues/193
-    // The working hypothesis is that agent info gossip which starts in the background
-    // after the local agent join may lead to an incoming connection being established
-    // more or less simultaneously, thereby evoking the race condition.
     iter_check!(2_000, 500, {
```
cdunster: Do we still need this if it was only failing on Tx5 (based on the removed comments)? Same for the other `iter_check!`'s below.

crates/bootstrap_srv/src/auth.rs
```diff
@@ -1,7 +1,7 @@
 //! Authentication module for the bootstrap server.
 //!
 //! This module provides authentication functionality that's independent
-//! of the relay implementation (SBD or Iroh). It implements the
+//! of the relay implementation. It implements the
 //! authentication hook server specification from:
 //! <https://github.com/holochain/sbd/blob/main/spec-auth.md>
```
cdunster, line 5: Is this link still valid as it mentions SBD?

crates/bootstrap_srv/src/auth.rs
```diff
@@ -1,7 +1,7 @@
 //! Authentication module for the bootstrap server.
 //!
 //! This module provides authentication functionality that's independent
-//! of the relay implementation (SBD or Iroh). It implements the
+//! of the relay implementation. It implements the
 //! authentication hook server specification from:
```
cdunster: Also, if you're modifying this comment anyways then maybe a reformat is good but it's fine if you're not making the change 🙂

cdunster (dismissed):
Small comments.

crates/bootstrap_srv/src/auth.rs
ThetaSinner: It's not ideal, we should consider moving that. I'll open an issue.

The spec is still relevant yes

crates/bootstrap_srv/src/auth.rs
ThetaSinner: Reformatted this section

crates/kitsune2/tests/blocks.rs
ThetaSinner: Good point, removed them all

crates/bootstrap_srv/src/auth.rs
ThetaSinner: @coderabbitai can you open an issue to work out what we should do with this spec? Maybe relocate it to this repository?

jost-s (approved):

PR #571, opened 2026-07-10
feat: stabilize authenticated iroh relay hosted in the bootstrap server

### LINKED ISSUE #492: Stabilize the Iroh relay hosted in the `bootstrap_srv` with authentication enabled
The experience using the relay hosted in the `bootstrap_srv` crate has been less good than using the public relay that Iroh host. That experience has been conflated with running with authentication enabled. Following the completion of https://github.com/holochain/kitsune2/issues/491, the relay is expected to work without authentication. Now it needs to work with authentication.

AC:
- [ ] Instructions are added to dino-adventure-kangaroo for obtaining auth material and configuring the app to use it.
- [ ] Relay is confirmed working at https://dev-test-bootstrap2-iroh-auth.holochain.org or remaining issues in Kitsune2 are resolved.

veeso:
## What changed

Authenticated clients now send their bearer token when connecting to the relay. The server validates it on every connection and keeps tokens used by live connections from expiring.

The transport also calls `PUT /relay/keepalive` every 120 seconds. This lets clients recover after a token becomes stale or the bootstrap server restarts, including relays configured for a single space. The endpoint replaces `PUT /relay/register`, and connecting without a token is no longer supported on this branch. Rejected connections now include the reason "not authorized".

## Why

Fixes #492. Authenticated nodes could lose relay access after token expiry or a server restart, then keep retrying with credentials the server no longer accepted.

## Authentication flow

The client authenticates, attaches the returned token to the relay connection, and starts a periodic keepalive:

```mermaid
sequenceDiagram
    participant Node as Kitsune2 node
    participant Conn as iroh relay connection<br>(lives inside the node)
    participant Srv as Bootstrap server<br>(embedded relay)
    participant Hook as Auth hook service

    Node->>Srv: PUT /authenticate with auth material
    Srv->>Hook: check auth material
    Hook-->>Srv: bearer token
    Srv-->>Node: bearer token
    Node->>Conn: start relay connection, token attached
    Conn->>Srv: connect with Authorization Bearer header
    Srv-->>Conn: allowed, token is valid
    Note over Node,Srv: every 120s the node also calls PUT /relay/keepalive,<br>which keeps its public key on the server's allowlist
```

## Recovery flow

Iroh 1.0.0 cannot replace the token on a running relay connection. The keepalive re-registers the node's public key, so the existing connection can recover even while it still presents the old token:

```mermaid
sequenceDiagram
    participant Node as Kitsune2 node
    participant Conn as iroh relay connection<br>(lives inside the node)
    participant Srv as Bootstrap server<br>(embedded relay)

    Note over Srv: server restarts or token expires,<br>all known tokens are gone
    Conn->>Srv: reconnect with the old token
    Srv-->>Conn: denied
    Note over Conn: iroh 1.0.0 cannot give a running relay<br>connection a new token, so it would<br>retry with the dead token forever
    Node->>Srv: keepalive, PUT /relay/keepalive<br>(re-authenticates first, so it carries a fresh token)
    Note over Srv: node's public key is back on the allowlist
    Conn->>Srv: retry, still with the old token
    Srv-->>Conn: allowed, public key is registered
```

## Notes

- The keepalive remains necessary until iroh supports refreshing the token on a live connection.
- Tests cover token authentication, stale-token recovery, single-space relays, and recovery after a bootstrap server restart.

## Follow-ups

- Document auth setup for dino-adventure-kangaroo.
- Verify the hosted relay after this change is deployed.
- Remove the keepalive endpoint when iroh supports live token refresh.

Refs #492

### DIFF
--- a/crates/bootstrap_client/src/lib.rs
+++ b/crates/bootstrap_client/src/lib.rs
@@ -4,7 +4,7 @@
 
 use base64::Engine;
 use kitsune2_api::{AgentInfoSigned, DynVerifier, K2Error, K2Result, SpaceId};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, PoisonError};
 use url::Url;
 
 /// Determine how we should handle an internal request for authorization
@@ -43,13 +43,25 @@ impl AuthMaterial {
         &self.auth_token
     }
 
+    /// Returns the currently cached auth token, if any.
+    pub fn token(&self) -> Option<String> {
+        self.auth_token
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .clone()
+    }
+
     fn priv_authenticate(
         &self,
         auth_url: &str,
         auth_type: AuthType,
     ) -> K2Result<()> {
         if matches!(auth_type, AuthType::IfUninit)
-            && self.auth_token.lock().unwrap().is_some()
+            && self
+                .auth_token
+                .lock()
+                .unwrap_or_else(PoisonError::into_inner)
+                .is_some()
         {
             return Ok(());
         }
@@ -83,7 +95,11 @@ impl AuthMaterial {
         let auth_token: AuthToken = serde_json::from_str(&token)
             .map_err(|err| K2Error::other_src("Authenticate Failed", err))?;
 
-        *self.auth_token.lock().unwrap() = Some(auth_token.auth_token);
+        *self
+            .auth_token
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner) =
+            Some(auth_token.auth_token);
 
         tracing::debug!("Authentication successful, token acquired");
         Ok(())
@@ -188,8 +204,12 @@ pub fn blocking_put_auth(
         let mut req = ureq::put(put_url);
 
         if let Some(auth_material) = auth_material {
-            let token =
-                auth_material.auth_token.lock().unwrap().clone().unwrap();
+            let token = auth_material
+                .auth_token
+                .lock()
+                .unwrap_or_else(PoisonError::into_inner)
+                .clone()
+                .expect("authenticated token must be cached");
             req = req.header("Authorization", &format!("Bearer {token}"));
         }
 
@@ -216,20 +236,57 @@ pub fn blocking_put_auth(
     res.into()
 }
 
-/// Register an iroh endpoint public key with the relay on the bootstrap server.
+/// Fetch the bearer token for the relay on the bootstrap server.
 ///
-/// After authenticating (which yields a bearer token), this function registers
-/// the 32-byte iroh public key with the server's relay allowlist so that the
-/// endpoint is permitted to connect to the relay.
+/// The returned token should be presented on the relay WebSocket upgrade via
+/// `RelayConfig::with_auth_token`, which sends it as an
+/// `Authorization: Bearer` header that the relay validates at connect time.
 ///
-/// This function should only be called when the server is configured with an
-/// auth hook server. Open relays do not expose the `relay/register` endpoint,
-/// and registration is not required when the relay has no access restrictions.
+/// A cached token is reused without contacting the server. Operations that
+/// receive a 401 response refresh the token internally before retrying.
 ///
 /// Note the `blocking_` prefix. This is a hint to the caller that if the
 /// function is used in an async context, it should be treated as a blocking
 /// operation.
-pub fn blocking_register_relay_key(
+///
+/// # Errors
+/// Returns an error if the authentication request fails, if the key is
+/// pending approval on the hook server, or if the response is malformed.
+pub fn blocking_fetch_relay_token(
+    mut server_url: Url,
+    auth_material: &AuthMaterial,
+) -> K2Result<String> {
+    server_url.set_path("authenticate");
+    auth_material.priv_authenticate(server_url.as_str(), AuthType::IfUninit)?;
+    auth_material.token().ok_or_else(|| {
+        K2Error::other("authentication succeeded but no token was cached")
+    })
+}
+
+/// Keep an iroh endpoint public key registered with the relay on the
+/// bootstrap server.
+///
+/// After authenticating (which yields a bearer token), this function
+/// registers the 32-byte iroh public key with the server's relay allowlist
+/// so that the endpoint stays permitted to connect to the relay.
+///
+/// The allowlist complements the bearer token presented on the relay
+/// WebSocket upgrade (see [`blocking_fetch_relay_token`]): iroh captures
+/// that token once per relay connection actor and cannot refresh it while
+/// the actor lives, so the allowlist — keyed on the handshake-proven public
+/// key — is what re-admits an actor whose token has gone stale. After a
+/// bootstrap server restart, this call first obtains a fresh token and then
+/// repopulates the allowlist. Call it periodically to keep the entry alive.
+///
+/// This function should only be called when the server is configured with an
+/// auth hook server. Open relays do not expose the `relay/keepalive`
+/// endpoint, and the keepalive is not required when the relay has no access
+/// restrictions.
+///
+/// Note the `blocking_` prefix. This is a hint to the caller that if the
+/// function is used in an async context, it should be treated as a blocking
+/// operation.
+pub fn blocking_relay_keepalive(
     mut server_url: Url,
     auth_material: &AuthMaterial,
     key_bytes: &[u8; 32],
@@ -237,23 +294,28 @@ pub fn blocking_register_relay_key(
     tracing::info!(
         server_url = %server_url,
         iroh_key = %base64::prelude::BASE64_URL_SAFE_NO_PAD.encode(key_bytes),
-        "Registering iroh endpoint key with relay service",
+        "Keeping iroh endpoint key registered with relay service",
     );
 
     server_url.set_path("authenticate");
     let auth_url = server_url.as_str().to_string();
     auth_material.priv_authenticate(&auth_url, AuthType::IfUninit)?;
 
-    server_url.set_path("relay/register");
-    let register_url = server_url.as_str().to_string();
+    server_url.set_path("relay/keepalive");
+    let keepalive_url = server_url.as_str().to_string();
 
-    fn priv_register(
-        register_url: &str,
+    fn priv_keepalive(
+        keepalive_url: &str,
         key_bytes: &[u8; 32],
         auth_material: &AuthMaterial,
     ) -> Res<()> {
-        let token = auth_material.auth_token.lock().unwrap().clone().unwrap();
-        ureq::put(register_url)
+        let token = auth_material
+            .auth_token
+            .lock()
+            .unwrap_or_else(PoisonError::into_inner)
+            .clone()
+            .expect("authenticated token must be cached");
+        ureq::put(keepalive_url)
             .header("Content-Type", "application/octet-stream")
             .header("Authorization", &format!("Bearer {token}"))
             .send(key_bytes.as_ref())
@@ -261,22 +323,20 @@ pub fn blocking_register_relay_key(
             .into()
     }
 
-    let mut res = priv_register(&register_url, key_bytes, auth_material);
+    let mut res = priv_keepalive(&keepalive_url, key_bytes, auth_material);
 
     if res.needs_auth() {
-        tracing::debug!(
-            "Relay key registration returned 401, re-authenticating"
-        );
+        tracing::debug!("Relay keepalive returned 401, re-authenticating");
         server_url.set_path("authenticate");
         let auth_url = server_url.as_str().to_string();
         auth_material.priv_authenticate(&auth_url, AuthType::Force)?;
-        res = priv_register(&register_url, key_bytes, auth_material);
+        res = priv_keepalive(&keepalive_url, key_bytes, auth_material);
     }
 
     let result: K2Result<()> = res.into();
     match &result {
-        Ok(()) => tracing::info!("Iroh relay key registration succeeded"),
-        Err(e) => tracing::warn!(?e, "Iroh relay key registration failed"),
+        Ok(()) => tracing::info!("Iroh relay keepalive succeeded"),
+        Err(e) => tracing::warn!(?e, "Iroh relay keepalive failed"),
     }
     result
 }
@@ -330,8 +390,12 @@ pub fn blocking_get_auth(
         let mut req = ureq::get(get_url);
 
         if let Some(auth_material) = auth_material {
-            let token =
-                auth_material.auth_token.lock().unwrap().clone().unwrap();
+            let token = auth_material
+                .auth_token
+                .lock()
+                .unwrap_or_else(PoisonError::into_inner)
+                .clone()
+                .expect("authenticated token must be cached");
             req = req.header("Authorization", &format!("Bearer {token}"));
         }
 
--- a/crates/bootstrap_client/tests/integration.rs
+++ b/crates/bootstrap_client/tests/integration.rs
@@ -1,6 +1,6 @@
 use kitsune2_bootstrap_srv::{BootstrapSrv, Config};
 use kitsune2_core::{Ed25519LocalAgent, Ed25519Verifier};
-use kitsune2_test_utils::enable_tracing;
+use kitsune2_test_utils::{auth::AuthHookServer, enable_tracing};
 use std::sync::Arc;
 use url::Url;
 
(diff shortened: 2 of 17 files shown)


veeso:
# Kangaroo instructions - create an issue once approved

## Using an authenticated bootstrap + relay server

The dev-test authenticated server is at
`https://dev-test-bootstrap2-iroh-auth.holochain.org/`.

Unlike the open dev-test server, this deployment requires every node to
authenticate before it can publish agent info or use the embedded iroh
relay. Authentication happens with opaque "auth material" issued by the
authentication hook service that fronts the bootstrap server.

### Obtain auth material

Auth material is an opaque byte string issued by the authentication hook
service. For the Holochain dev-test deployment, request auth material from
the infra team.

### Configure the app

1. In `kangaroo.config.ts`, point both `bootstrapUrl` and `relayUrl` at the
   authenticated server:

   ```ts
   bootstrapUrl: 'https://dev-test-bootstrap2-iroh-auth.holochain.org/',
   relayUrl: 'https://dev-test-bootstrap2-iroh-auth.holochain.org/',
   ```

2. Provide the auth material to the conductor. Kitsune2's iroh transport
   reads it from the transport module config as base64
   (`authMaterialRelayBase64`), and the bootstrap module uses the same
   material for `PUT /authenticate`. In the conductor config this lives
   under the network / kitsune2 module config section.
   

### How it behaves

- On startup the node calls `PUT /authenticate` with the auth material and
  receives a bearer token. The token is presented on the relay connection
  and on every bootstrap request.
- The node re-registers itself with the server periodically (default every
  120 seconds), so it recovers automatically if the bootstrap server
  restarts or the token expires. No user action is needed after the
  initial configuration.
- If the auth material itself is rejected (HTTP 401) or pending approval
  (HTTP 202), networking will not come up: check with the operator of the
  auth hook service.

crates/bootstrap_srv/src/test/iroh_relay_axum.rs
```diff
@@ -129,16 +129,24 @@ fn use_bootstrap_and_iroh_relay_with_tls() {
...
-    enable_tracing();
-    let _ = rustls::crypto::ring::default_provider().install_default();
+/// How a test client proves itself to the authenticated relay.
+enum AuthFlow {
+    /// Present the bearer token on the relay WebSocket upgrade via
+    /// `RelayConfig::with_auth_token` (the current flow).
+    BearerToken,
+    /// Register the endpoint public key via `PUT /relay/register` and
+    /// connect without a token (the transitional 0.4.x flow).
+    LegacyRegister,
```
jost-s: I wouldn't support this on the main branch.

crates/transport_iroh/src/lib.rs
```diff
@@ -620,13 +697,172 @@ impl IrohTransport {
...
+                        // stay quiet and keep the loop alive.
+                        debug!(?e, "Relay key re-registration failed");
+                    }
+                }
+            }
+        })
+        .abort_handle()
+    }
+
+    /// Spawns a background task that watches the home relay connection and
```
jost-s: I don't know what this is meant to solve. If the relay is persistently down, connections can no longer be established. Connectivity is essentially broken then. It doesn't seem relevant what happens to tokens at that moment.

crates/bootstrap_srv/src/test/iroh_relay_axum.rs
veeso: Removed legacy register, addded keepalive to keep working if the boostrap srv goes down.

crates/transport_iroh/src/lib.rs
veeso: Removed. It was meant to solve a "downtime" when the token expired, but it's probably better to remove it anyway

crates/bootstrap_client/src/lib.rs
```diff
@@ -43,6 +43,11 @@ impl AuthMaterial {
         &self.auth_token
     }
 
+    /// Returns the currently cached auth token, if any.
+    pub fn token(&self) -> Option<String> {
+        self.auth_token.lock().unwrap().clone()
```
ThetaSinner: Doesn't need to be unwrapped as a panic, can recover the lock from the error.

If not then we have a convention of using `.expect("poisoned")` in this repository - maybe that's something we should put in the CLAUDE.md now.

crates/bootstrap_client/tests/integration.rs
```diff
@@ -369,3 +369,73 @@ async fn reauth_on_token_expiration() {
...
+            .build()
+            .unwrap();
+        rt.block_on(async {
+            let listener =
+                tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
+            let _ = addr_tx.send(listener.local_addr().unwrap());
+
+            let app = axum::Router::new().route(
+                "/authenticate",
+                // M-OOBE: Consume the request body before responding so that
```
ThetaSinner: Huh? I can find out what this means by googling, but I don't see how M-OOBE is relevant here?

ThetaSinner: Or actually this comment at all, we're not doing any request handling ourselves at all?

crates/bootstrap_client/src/lib.rs
```diff
@@ -216,67 +221,110 @@ pub fn blocking_put_auth(
     res.into()
 }
 
-/// Register an iroh endpoint public key with the relay on the bootstrap server.
+/// Authenticate with the bootstrap server and return the relay bearer token.
```
ThetaSinner: The previous text was a bit clearer about how the remote is structured. "the relay on the bootstrap server" is clearer than "authenticate with bootstrap and return the relay". If possible, having that structure remain clear would make this easier to follow.

crates/bootstrap_client/src/lib.rs
```diff
@@ -216,67 +221,110 @@ pub fn blocking_put_auth(
...
+/// `RelayConfig::with_auth_token`, which sends it as an
+/// `Authorization: Bearer` header that the relay validates at connect time.
 ///
-/// After authenticating (which yields a bearer token), this function registers
-/// the 32-byte iroh public key with the server's relay allowlist so that the
-/// endpoint is permitted to connect to the relay.
+/// With `force = false` a cached token is reused without contacting the
+/// server; with `force = true` the server is always re-contacted. Force a
+/// refresh after the relay has denied the current token, e.g. because it
+/// idle-expired or the server restarted and lost its token state.
```
ThetaSinner: To follow what `blocking_put_auth` does, should we not internalize that logic within this client? If the response indicates that we need to authenticate then we can just do that. The `force` param isn't currently in the public API I think

crates/bootstrap_client/tests/integration.rs
```diff
@@ -369,3 +369,73 @@ async fn reauth_on_token_expiration() {
...
+
+#[test]
+fn fetch_relay_token_authenticates_and_caches() {
+    enable_tracing();
+
+    let (server_url, hits) = spawn_stub_auth_server();
+
+    let auth = AuthMaterial::new(b"material".to_vec());
+
+    let tok1 = blocking_fetch_relay_token(server_url.clone(), &auth, false)
```
ThetaSinner: This crate depends on bootstrap_srv so would it not make sense to call this auth server through the real bootstrap_srv?

crates/bootstrap_client/tests/integration.rs
```diff
@@ -369,3 +369,73 @@ async fn reauth_on_token_expiration() {
 
     auth_hook_server_task.abort();
 }
+
+/// Spawn a stub auth server counting `/authenticate` hits and returning
+/// `tok-<hit-number>` tokens. Returns the server URL and the hit counter.
+fn spawn_stub_auth_server() -> (Url, Arc<std::sync::atomic::AtomicUsize>) {
```
ThetaSinner: There are several things like this defined now, perhaps this belongs in `kitsune2_test_utils`? If the consolidation is too big for this PR then please can a follow up issue be opened?

crates/bootstrap_srv/src/test/iroh_relay_axum.rs
```diff
@@ -129,16 +129,26 @@ fn use_bootstrap_and_iroh_relay_with_tls() {
     create_two_endpoints_and_assert_message_is_received(true, addr);
 }
 
-#[test]
-fn use_bootstrap_and_iroh_relay_with_auth() {
-    enable_tracing();
-    let _ = rustls::crypto::ring::default_provider().install_default();
+/// How a test client proves itself to the authenticated relay.
+enum AuthFlow {
```
ThetaSinner: Could this be simplified to a `use_stale_token` boolean flag? It'd be less code for the same result I think

crates/bootstrap_srv/src/http.rs
```diff
@@ -456,6 +469,15 @@ fn tokio_thread(
                     None
                 };
 
+            // Tracks which auth token backs each live relay connection so
+            // the prune worker can keep those tokens from idle-expiring.
+            // Only needed when relay access is gated on authentication.
```
ThetaSinner:
```suggestion
```

crates/bootstrap_srv/src/iroh_relay_axum.rs
```diff
@@ -510,52 +509,28 @@ pub fn create_relay_state(
...
+/// valid bearer token on the WebSocket upgrade (`Authorization: Bearer`),
+/// with a recovery fallback to the `PUT /relay/keepalive` allowlist for
+/// clients whose token went stale.
 ///
 /// Use this when the bootstrap server is configured with an authentication
-/// hook server. Clients must call `PUT /authenticate` and then
-/// `PUT /relay/register` before their relay connection will be accepted.
-pub fn create_relay_state_with_allowlist(
+/// hook server. Clients call `PUT /authenticate` to obtain the token and
+/// present it directly when connecting to the relay.
```
ThetaSinner:
```suggestion
/// hook server.
```

Is that detail relevant here?

crates/bootstrap_srv/src/relay_access.rs
```diff
@@ -0,0 +1,311 @@
...
+use iroh_base::EndpointId;
+use iroh_relay::server::{Access, AccessControl, ClientRequest, ConnectionId};
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+
+/// Tracks which bearer token authorized each live relay connection.
+#[derive(Clone, Debug, Default)]
+pub struct RelayConnTracker {
+    // ConnectionId is process-unique (guaranteed upstream), so a flat map
+    // suffices; multiple connections may share one token.
```
ThetaSinner:
```suggestion
    // ConnectionId is process-unique (guaranteed upstream), so it is safe to use as a key here; multiple connections may share one token.
```

crates/bootstrap_srv/src/relay_access.rs
```diff
@@ -0,0 +1,311 @@
...
+    conns: Arc<Mutex<HashMap<ConnectionId, Arc<str>>>>,
+}
+
+impl RelayConnTracker {
+    fn insert(&self, id: ConnectionId, token: Arc<str>) {
+        self.conns.lock().unwrap().insert(id, token);
```
ThetaSinner: Same comment about `unwrap` here and in `remove` as I made above.

crates/bootstrap_srv/src/relay_access.rs
```diff
@@ -0,0 +1,311 @@
...
+                self.conns.insert(request.connection_id(), token);
+                tracing::debug!(
+                    key = %request.endpoint_id().fmt_short(),
+                    "Relay access granted via bearer token"
+                );
+                return Access::Allow;
+            }
+
+            // Recovery: the token is stale (idle-expired or wiped by a
+            // server restart) and iroh cannot swap it on the live relay
```
ThetaSinner: I think we need to be careful about wording throughout. If the token has been wiped from `auth_tracker` by a restart - then the `recovery_allowlist` has also been wiped. I think the only path to recovery from a server restart is to re-authenticate?

crates/transport_iroh/src/lib.rs
```diff
@@ -291,6 +291,25 @@ pub mod config {
...
         #[cfg_attr(feature = "schema", schemars(skip))]
         pub auth_material_relay_base64: Option<String>,
+
+        /// Interval in seconds of the keepalive that re-registers the
+        /// endpoint public key with an authenticated relay's bootstrap
+        /// server.
+        ///
+        /// The keepalive keeps the server-side relay allowlist entry
+        /// alive and restores relay access after the bootstrap server
+        /// restarts and loses its in-memory authentication state. Only used
```
ThetaSinner: Another reference to restart, I'm not convinced by that

crates/transport_iroh/src/lib.rs
```diff
@@ -605,12 +640,17 @@ impl IrohTransport {
...
-        // entry is refreshed before the server prunes it. This also
-        // handles the case where the server restarts and loses its
-        // in-memory allowlist.
-        let relay_re_registration_task = relay_registration_params
-            .map(Self::spawn_relay_re_registration_task);
+        // The keepalive keeps the server-side allowlist entry alive, which
+        // admits relay actors that hold a stale bearer token (iroh captures
+        // the token once per relay actor and cannot refresh it while the
+        // actor lives), so relay access recovers after token idle-expiry or
+        // a bootstrap server restart.
```
ThetaSinner: Another restart reference. I'd argue that this information doesn't need to be reproduced here at all. It's already well documented on the API.

crates/transport_iroh/src/lib.rs
```diff
@@ -620,13 +660,98 @@ impl IrohTransport {
...
+    ///
+    /// The bearer token presented on the relay WebSocket upgrade is
+    /// captured once per relay actor inside iroh and cannot be refreshed
+    /// while the actor lives, so a denied actor can retry with a stale
+    /// token indefinitely. The server admits such actors through the
+    /// allowlist by their handshake-proven public key, which makes this
+    /// keepalive the reliable recovery path for token idle-expiry and for
+    /// bootstrap server restarts that wipe the in-memory auth state. The
+    /// keepalive re-authenticates automatically when the server returns
+    /// 401.
```
ThetaSinner: Again, can we compact the docs that are already present elsewhere. This is an internal so we just need to signpost to key types and modules where the information can be found - we don't need to repeat

crates/transport_iroh/src/lib.rs
```diff
@@ -620,13 +660,98 @@ impl IrohTransport {
...
+
+                match Self::relay_keepalive(&params).await {
+                    Ok(()) => {
+                        debug!("Relay keepalive succeeded");
+                    }
+                    Err(e) => {
+                        // A failed keepalive is a tolerated transient
+                        // state (e.g. the bootstrap server is offline):
+                        // stay quiet and keep the loop alive.
+                        debug!(?e, "Relay keepalive failed");
```
ThetaSinner: Is it worth tracking consecutive failures and logging at a higher level? Or I guess if this fails a couple of times in a row then the token would need to be replaced anyway.

crates/transport_iroh/tests/relay_auth.rs
```diff
@@ -189,3 +202,146 @@ async fn authenticated_relay_two_transports_can_communicate() {
     .await
     .expect("message was not received by ep_2 within 30 s");
 }
+
+/// The bootstrap server restarts, wiping its in-memory token state. The
+/// transport's registration keepalive must re-authenticate, re-register
+/// the endpoint key, and restore relay connectivity without intervention.
```
ThetaSinner, line 169: Great, I was going to ask for a test for this if there wasn't one

crates/transport_iroh/tests/relay_auth.rs
```diff
@@ -189,3 +202,146 @@ async fn authenticated_relay_two_transports_can_communicate() {
...
+///
+/// Local-only: restarting the deployed server is not possible from a test.
```
ThetaSinner:
```suggestion
```

ThetaSinner (commented):
This looks good. I just needed more time to absorb why as many changes were needed to integrate with Iroh. The changes do make sense with a more careful look now.

I've left quite a few comments but I think the total change I'm asking for should be really quite small.

crates/transport_iroh/src/lib.rs
```diff
@@ -620,13 +660,98 @@ impl IrohTransport {
...
+        let params = params.clone();
+        tokio::task::spawn_blocking(move || {
+            kitsune2_bootstrap_client::blocking_relay_keepalive(
+                params.server_url.clone(),
+                &params.auth_material,
+                &params.key_bytes,
+            )
+        })
+        .await
+        .map_err(|e| K2Error::other_src("Registration task failed", e))?
```
ThetaSinner, line 677: One more thing I noticed here. This error will be returned and logged at debug but is actually quite different to the inner failure. Perhaps worth ensuring this gets logged at `warn` like the previous code did?

veeso: Now it is logged as warning by the caller.

veeso:
@ThetaSinner All addressed as requested

crates/bootstrap_client/tests/integration.rs
veeso: removed, was still conventional comments being written due to a bug with my skills

crates/bootstrap_client/src/lib.rs
veeso: done

ThetaSinner (approved):

Issue #160, opened 2025-03-27
Add sharding module


ThetaSinner:
Kitsune2 should attempt to determine how many other agents are storing the same data and recommend an appropriate target arc based on the desired redundancy level.

topeuph-ai:
Hi! We build on Holochain ([ValiChord](https://github.com/ValiChord/ValiChord) — commit-reveal verification of research results), and dynamic storage arcs matter to where we want to take it, so we spent some time with this issue. Sharing what we found in case it's useful.

**First: the sockets you've already left made this very approachable.** The `sharding` feature flag with the verified-sync grow side in `storage_arc.rs`, the `set_tgt_storage_arc_hint` seam on `LocalAgent`, and the transport's module-message routing meant a complete controller PoC slotted in behind the existing flag without touching `kitsune2_api` at all. That's unusually good extensibility for a dormant feature.

**We simulated the controller design space before writing Rust** — specifically the oscillation dynamics that led to arcs being fixed in 2021: 200 agents on a 512-sector ring (matching `SECTOR_SIZE` exactly), quantised power-of-two arcs, per-viewer gossip staleness, four disruption scenarios (cold activation, 30% simultaneous node loss, 60% flash crowd, sustained churn) × 312 seeds × 4 controller variants = 4,992 paired simulations, plus a 20-generation evolutionary adversarial search over disruption schedules.

| controller variant | runs with sector loss (of 1,248) |
|---|---|
| naive (react immediately to redundancy estimate) | 1,197 — reproduces the historical oscillation |
| + hysteresis scaled to own measured staleness (grow 1×, shrink 4×) | 300 |
| + desynchronised decision epochs (jitter) | 303 — jitter alone doesn't help, worth knowing since it's the textbook oscillation remedy |
| + two-phase "polite shrink" | **0** (95% upper bound < 0.24%) |

The two-phase polite shrink: announce intent to vacate → wait a multiple of measured staleness → re-check, counting *lower-agent-id* intenders as already gone; the lowest conflicting id proceeds and the others stand down. The asymmetry is essential — symmetric deference livelocks the whole network (we tried it first). Growers count announced vacates as already gone, so they move before a hole opens. The adversarial search broke both damped variants but couldn't orphan a single sector under this one in 20 generations.

[...]

PR #572, opened 2026-07-12
fix(core): mem transport marks peers unresponsive on send failure

topeuph-ai:
`TxImpHnd::set_unresponsive` documents that transports should call it whenever a connection fails to establish, a send fails, or a peer disconnects. The iroh transport honours this; the mem transport returned `Connection Closed` errors without ever marking the peer — so in any test built on the mem transport, a killed node never becomes unresponsive: peer selection keeps offering it, and anything downstream of unresponsive tracking is blind to the death.

This is the standalone fix we offered in [#160 (comment)](https://github.com/holochain/kitsune2/issues/160#issuecomment-4949241059) — it's independent of everything else discussed there, and just brings the mem transport up to the contract in `transport.rs` so failure-mode tests can see node deaths.

The change keeps a handle on the `TxImpHnd` and marks the peer when a send fails (mirroring the iroh transport's behaviour). One existing behaviour is preserved: the error returned to the caller is unchanged.

Tested: full `kitsune2_core` suite passes (99 + 3), no regressions. We found this while writing a multi-node failure test — post-kill, ghost peers stayed "responsive" indefinitely, which quietly invalidates any mem-transport test that kills nodes.

Transparency: developed with AI assistance (Claude), human-directed and reviewed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/core/src/factories/mem_transport.rs
+++ b/crates/core/src/factories/mem_transport.rs
@@ -53,6 +53,9 @@ struct MemTransport {
     task_list: Arc<Mutex<tokio::task::JoinSet<()>>>,
     cmd_send: CmdSend,
     net_stats: Arc<Mutex<TransportStats>>,
+    /// Handler used to notify the wrapping transport of peer state
+    /// changes, e.g. marking a peer unresponsive after a failed send.
+    handler: Arc<TxImpHnd>,
 }
 
 impl Drop for MemTransport {
@@ -64,6 +67,9 @@ impl Drop for MemTransport {
 }
 
 impl MemTransport {
+    /// Construct a new memory transport implementation: registers a fresh
+    /// in-process listener, reports its address to the handler, and spawns
+    /// the connection-listener and command-runner tasks.
     pub async fn create(handler: Arc<TxImpHnd>) -> DynTxImp {
         let mut listener = get_transport_instances().listen();
         let this_url = listener.url();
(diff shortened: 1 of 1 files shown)


topeuph-ai:
@coderabbitai full review

crates/core/src/factories/mem_transport.rs
```diff
@@ -142,17 +149,31 @@ impl TxImp for MemTransport {
...
                 Ok(_) => match result_receiver.await {
                     Ok(result) => result,
                     Err(_) => Err(K2Error::other("Connection Closed")),
                 },
+            };
+            if result.is_err() {
+                // Without this, tests that kill nodes never see those
+                // nodes become unresponsive.
+                let _ =
+                    self.handler.set_unresponsive(peer, Timestamp::now()).await;
```
topeuph-ai, line 174: Self-review note on the approach: all failure paths funnel into `result`, so marking here covers both the real case (the command task reporting a failed send to that peer) and the channel-closed paths. The channel-closed case only occurs during local transport shutdown, where the extra marking is moot. The `set_unresponsive` result is deliberately dropped — the original send error is the one the caller needs, unchanged. Considered marking inside `cmd_task` where the send actually fails instead, but the result already flows back to this oneshot, so handling it here keeps the marking in one place and `cmd_task` untouched.

jost-s:
Thanks for the contribution.

The mem transport is only used in tests. If the feature for marking peers unresponsive hasn't been part of it so far, it's currently not needed there. I prefer to not add it while it's not being used.

PR #588, opened 2026-07-15
feat(transport_iroh): close connections gracefully with close codes

veeso:
## What changed

A connection to another peer can now be closed on purpose, with a reason. The closing side sends a numeric close code and a short reason text as part of the close itself. The other side reads that code and, when the close was intentional, it lets the connection go quietly and passes the reason on to the application instead of treating the peer as unreachable.

Before this change, an intentional close looked exactly like a network failure to the other side. The other peer would mark the closer as unresponsive and avoid it until that state expired, which could stall communication for no good reason. One example: when a space is removed, the node closes its connections, and the peers on the other end wrongly penalised it.

There are three close codes: one for an intentional close, one for the case where two peers opened duplicate connections at the same time and one of them is discarded, and a fallback for anything else. Closes from peers running older versions are still understood through the previous mechanism, so mixed-version networks keep working.

## Why

Closes #496. The issue asks for the ability to close a connection, a defined set of close reasons, and for the other side to be told the close was intentional so it can release the connection without treating it as an error. Upcoming connection management work needs this.

## Notes

The issue suggested looking into the strum crate for mapping codes to reasons. It was evaluated and skipped: there is a single small enum, and a hand-written mapping is less code than the dependency.

The issue also mentioned a possible double disconnect through the reader task. That was already fixed on main by an earlier change (88949e9), so this PR does not need to address it.

Known limitation: a peer running an older version does not understand the new codes, so it will still treat a graceful close from a new peer as a failure. That resolves as peers upgrade.

The cleanup logic that decides how to react when a connection dies was also restructured into a small decision function with its own table of unit tests, plus an end to end test where one peer disconnects and the other observes the reason without penalising it.

### DIFF
--- a/crates/transport_iroh/src/close_code.rs
+++ b/crates/transport_iroh/src/close_code.rs
@@ -1,0 +1,59 @@
+//! Application-level close codes for iroh connections.
+//!
+//! A QUIC application close carries a numeric error code plus reason bytes.
+//! These codes let the remote peer distinguish an intentional close from a
+//! network failure, so it can release the connection quietly.
+
+/// Application close code carried in the QUIC CONNECTION_CLOSE frame.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub(crate) enum CloseCode {
+    /// Unknown close code treated as a connection failure.
+    Unspecified = 0,
+    /// The peer closed the connection intentionally. The close reason bytes
+    /// carry a human-readable reason. Not a failure.
+    Graceful = 1,
+    /// The connection lost simultaneous-open resolution and was replaced by
+    /// a preferred connection to the same peer. Not a failure.
+    Superseded = 2,
+}
+
+impl CloseCode {
+    /// Numeric wire representation of this close code.
+    pub fn as_u8(self) -> u8 {
+        self as u8
+    }
+
+    /// Maps a wire error code to a [`CloseCode`].
+    ///
+    /// Future unknown codes map to [`CloseCode::Unspecified`].
+    pub fn from_wire(code: u64) -> Self {
+        match code {
+            1 => CloseCode::Graceful,
+            2 => CloseCode::Superseded,
+            _ => CloseCode::Unspecified,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn round_trips_known_codes() {
+        for code in [
+            CloseCode::Unspecified,
+            CloseCode::Graceful,
+            CloseCode::Superseded,
+        ] {
+            assert_eq!(CloseCode::from_wire(code.as_u8() as u64), code);
+        }
+    }
+
+    #[test]
+    fn unknown_wire_code_maps_to_unspecified() {
+        assert_eq!(CloseCode::from_wire(3), CloseCode::Unspecified);
+        assert_eq!(CloseCode::from_wire(u64::MAX), CloseCode::Unspecified);
+    }
+}
--- a/crates/transport_iroh/src/connection_context.rs
+++ b/crates/transport_iroh/src/connection_context.rs
@@ -1,5 +1,6 @@
 use crate::IrohTransport;
 use crate::SpaceRelays;
+use crate::close_code::CloseCode;
 use crate::connection::DynConnection;
 #[cfg(feature = "metrics")]
 use crate::metrics::connection_counter_metric;
@@ -22,17 +23,56 @@ use std::{
 use tokio::{sync::MutexGuard, task::AbortHandle};
 use tracing::{debug, error, info, trace, warn};
 
-/// Application close reason sent when a connection is intentionally torn down
-/// because a different, preferred connection to the same peer won
-/// simultaneous-open resolution.
-///
-/// The remote end recognises this reason (via
-/// [`Connection::remote_close_reason`](crate::connection::Connection::remote_close_reason))
-/// and treats the close as a deliberate supersession rather than a peer
-/// disconnect, so it does not mark the peer unresponsive or fire
-/// `peer_disconnect`.
+/// Reason sent when a preferred connection wins simultaneous-open resolution.
 pub(super) const SUPERSEDED_CLOSE_REASON: &[u8] =
     b"superseded by preferred connection";
+
+/// Outcome of the connection reader's accept loop.
+struct ReaderExit {
+    /// Human-readable description of why the loop ended.
+    err: String,
+    /// Whether the reader failure should mark the peer unresponsive.
+    mark_unresponsive: bool,
+}
+
+/// Cleanup action for a stopped connection reader.
+#[derive(Debug, PartialEq, Eq)]
+pub(super) enum ReaderCleanup {
+    /// Reports that the active peer connection failed.
+    PeerGone {
+        /// Whether to mark the peer unresponsive.
+        mark_unresponsive: bool,
+    },
+    /// Reports an intentional remote close to local handlers.
+    Graceful {
+        /// The remote's close reason.
+        reason: String,
+    },
+    /// Closes a superseded or inactive connection without notifying handlers.
+    Quiet,
+}
+
+/// Selects the cleanup action for a stopped reader.
+pub(super) fn classify_exit(
+    was_active: bool,
+    remote_close: Option<(CloseCode, Bytes)>,
+    mark_unresponsive: bool,
+) -> ReaderCleanup {
+    let superseded_by_remote =
+        matches!(remote_close, Some((CloseCode::Superseded, _)));
+
+    if !was_active || superseded_by_remote {
+        return ReaderCleanup::Quiet;
+    }
+
+    if let Some((CloseCode::Graceful, reason)) = remote_close {
+        return ReaderCleanup::Graceful {
+            reason: String::from_utf8_lossy(&reason).to_string(),
+        };
+    }
+
+    ReaderCleanup::PeerGone { mark_unresponsive }
+}
 
 pub(super) struct ConnectionContext {
     handler: Arc<TxImpHnd>,
@@ -240,12 +280,18 @@ impl ConnectionContext {
             // The displaced connection was previously counted as active.
             #[cfg(feature = "metrics")]
             connection_counter_metric().add(-1, &[]);
-            displaced.connection.close(0u8, SUPERSEDED_CLOSE_REASON);
+            displaced
+                .connection
+                .close(CloseCode::Superseded, SUPERSEDED_CLOSE_REASON);
         }
 
         true
     }
 
+    /// Abort the connection reader task, if it is still running.
+    ///
+    /// Used during transport shutdown, where the reader must stop
+    /// immediately without running its exit cleanup.
     pub fn abort_tasks(&self) {
         if let Some(abort_handle) = self
             .connection_reader_abort_handle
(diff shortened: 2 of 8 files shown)


crates/transport_iroh/src/connection_context.rs
```diff
@@ -298,124 +392,171 @@ impl ConnectionContext {
         local_url: Arc<RwLock<Option<Url>>>,
     ) -> AbortHandle {
         tokio::spawn(async move {
-            // Track whether to skip marking peer as unresponsive.
```
veeso, line 301: the logic inside of this task was too complex and subject to 6 levels of nesting.

I have moved the logic into two separate functions.
The first one gathers exit reasons/case, the other applies based on the case.

crates/transport_iroh/src/close_code.rs
```diff
@@ -0,0 +1,61 @@
+//! Application-level close codes for iroh connections.
+//!
+//! A QUIC application close carries a numeric error code plus reason bytes.
+//! These codes let the remote peer distinguish an intentional close from a
+//! network failure, so it can release the connection quietly instead of
+//! marking the peer unresponsive.
```
ThetaSinner:
```suggestion
//! network failure, so it can release the connection quietly.
```

That's specific to what our application may do in some cases - doesn't seem important here.

crates/transport_iroh/src/close_code.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// Application close code carried in the QUIC CONNECTION_CLOSE frame.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub(crate) enum CloseCode {
+    /// Unknown or legacy close. Treated as a connection failure by the
```
ThetaSinner: No need to document "legacy", we don't expect compatibility with previous K2 versions after the release of 0.5.x.

Actually, the way that `from_wire` works, it's covering future unknown codes, rather than legacy ones really.

crates/transport_iroh/src/lib.rs
```diff
@@ -1012,12 +1013,18 @@ impl TxImp for IrohTransport {
     fn disconnect(
         &self,
         peer: Url,
-        _payload: Option<(String, Bytes)>,
+        payload: Option<(String, Bytes)>,
```
ThetaSinner, line 1085: We always discard the `Bytes` part of the tuple, is that needed?

crates/transport_iroh/src/connection_context.rs
```diff
@@ -30,10 +31,94 @@ use tracing::{debug, error, info, trace, warn};
 /// [`Connection::remote_close_reason`](crate::connection::Connection::remote_close_reason))
 /// and treats the close as a deliberate supersession rather than a peer
 /// disconnect, so it does not mark the peer unresponsive or fire
-/// `peer_disconnect`.
+/// `peer_disconnect`. Peers running current code additionally signal this
+/// with [`CloseCode::Superseded`]; the reason-string match is kept for
+/// compatibility with peers that close with code 0.
```
ThetaSinner: I don't think we need to be compatible, we've already made protocol breaking changes during the 0.5.x dev release series

crates/transport_iroh/src/connection_context.rs
```diff
@@ -30,10 +31,94 @@ use tracing::{debug, error, info, trace, warn};
...
 pub(super) const SUPERSEDED_CLOSE_REASON: &[u8] =
     b"superseded by preferred connection";
 
+/// Outcome of the connection reader's accept loop.
+struct ReaderExit {
+    /// Human-readable description of why the loop ended.
+    err: String,
+    /// True when the loop ended on a temporary no-local-agents preflight
+    /// error, which is a timing issue rather than a network problem and
+    /// must not mark the peer unresponsive.
```
ThetaSinner: Hmm, that's not supposed to happen. We're supposed to prevent outbound connections when we have no local peers. We also wouldn't have published an agent info unless we've got an agent for it. So this isn't supposed to happen with outbound or inbound connection initiation. Did you run into a case where it can? That suggests we need a bug fix elsewhere rather than this

crates/transport_iroh/src/connection_context.rs
```diff
@@ -30,10 +31,94 @@ use tracing::{debug, error, info, trace, warn};
...
+/// tear down a surviving connection.
+#[derive(Debug, PartialEq, Eq)]
+pub(super) enum ReaderCleanup {
+    /// The active connection died for a genuine reason: fire
+    /// `peer_disconnect` and, unless the failure is known to be temporary,
+    /// mark the peer unresponsive.
+    PeerGone {
+        /// False when the reader ended on a temporary no-local-agents
+        /// preflight error, which will resolve once an agent joins.
+        mark_unresponsive: bool,
```
ThetaSinner, line 44: More handling of the same thing above, needs discussion or investigation

crates/transport_iroh/src/connection_context.rs
```diff
@@ -30,10 +31,94 @@ use tracing::{debug, error, info, trace, warn};
...
+    ///   preferred or newer connection already replaced it locally).
+    /// - The *remote* closed it as superseded (via
+    ///   [`CloseCode::Superseded`], or [`SUPERSEDED_CLOSE_REASON`] from
+    ///   peers running older code) because it preferred a different
+    ///   connection. This can arrive while this connection is still our
+    ///   active map entry, because the winning connection has not finished
+    ///   its preflight and registered yet. Without honouring that close we
+    ///   would wrongly mark the peer unresponsive (which lasts until the
+    ///   agent info expires) and stall gossip, even though a usable
+    ///   connection is moments away.
```
ThetaSinner: That's a lot of docs for the enum variant, that information should be captured elsewhere. Can this be made more concise?

crates/transport_iroh/src/connection_context.rs
```diff
@@ -30,10 +31,94 @@ use tracing::{debug, error, info, trace, warn};
...
+
+    if let Some((CloseCode::Graceful, reason)) = remote_close {
+        return ReaderCleanup::Graceful {
+            reason: String::from_utf8_lossy(&reason).to_string(),
+        };
+    }
+
+    ReaderCleanup::PeerGone {
+        mark_unresponsive: !skip_unresponsive,
+    }
```
ThetaSinner: Can these variables be made the same so that we don't need to negate here?

crates/transport_iroh/src/connection_context.rs
```diff
@@ -298,124 +392,171 @@ impl ConnectionContext {
...
+    // to be accepted. The connection cannot recover from a failed preflight,
+    // and a new connection must be established.
+    async fn run_reader_loop(
+        ctx: &Arc<Self>,
+        connections: &Connections,
+        local_url: &Arc<RwLock<Option<Url>>>,
+    ) -> ReaderExit {
+        // Track whether to skip marking peer as unresponsive.
+        // This is set to true for temporary errors like NoLocalAgentsDuringPreflight,
+        // which indicate a timing issue rather than a network problem.
```
ThetaSinner: There's quite a bit about this. I suspect the design wasn't properly written down and is recorded in MM but this was intended to be avoided.

crates/transport_iroh/src/tests/simultaneous_open.rs
```diff
@@ -240,26 +47,58 @@ async fn superseded_remote_close_does_not_mark_unresponsive() {
...
+        recorder.disconnects.lock().unwrap().is_empty(),
+        "a connection superseded by the remote must not fire peer_disconnect"
+    );
+}
+
+/// Peers running older code close a superseded connection with code 0 and
+/// only the reason byte string. That legacy close must still be recognised
+/// as a supersession.
+#[tokio::test(flavor = "multi_thread")]
+async fn legacy_superseded_reason_string_does_not_mark_unresponsive() {
```
ThetaSinner: Think we can skip this

crates/transport_iroh/src/tests/support.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    ActiveReader {
+        ctx,
+        accept_gate,
+        close_calls,
+    }
+}
+
+/// Poll `condition` every 10ms until it returns true or `timeout` elapses.
+pub(super) async fn wait_for<F>(mut condition: F, timeout: Duration)
```
ThetaSinner: Is one of these not defined in `kitsune2_test_utils`?

crates/transport_iroh/src/tests/support.rs
ThetaSinner, line 1: As a maybe, since this code was extracted from `simultaneous_open`, does it make sense to make it a sub module of that test?

ThetaSinner (commented):
Main query here is about the handling of no peers during preflight. I think that might be something that we did in the mem transport and tx5 but didn't make it into the iroh transport. I wouldn't expect to need to handle that as this PR does

crates/transport_iroh/src/close_code.rs
veeso: done

crates/transport_iroh/src/close_code.rs
veeso: removed

crates/transport_iroh/src/lib.rs
veeso: it's not needed by iroh. The DefaultTransport creates these bytes for TxImp implementations that send a disconnect data frame, but iroh directly carries those in the QUIC close frame, so it's not necessary.

crates/transport_iroh/src/connection_context.rs
veeso: this already exists on main, it was just split into different function to simplify the code.

crates/transport_iroh/src/connection_context.rs
veeso: this already exists on main, it was just split into different function to simplify the code.

crates/transport_iroh/src/connection_context.rs
veeso: simplified

crates/transport_iroh/src/connection_context.rs
veeso: renamed

crates/transport_iroh/src/connection_context.rs
veeso: Also this behaviour already existed on main, so it was kept as-is for this issue

crates/transport_iroh/src/tests/simultaneous_open.rs
veeso: removed

crates/transport_iroh/src/connection_context.rs
veeso: Backward compatibility removed.

crates/transport_iroh/src/tests/support.rs
veeso: removed, now using kitsune2_test_utils::retry_fn_until_timeout.

crates/transport_iroh/src/tests/support.rs
veeso: No, it's also being used inside of `close.rs`

crates/transport_iroh/src/connection_context.rs
ThetaSinner: Oh I see, okay, fine then

ThetaSinner (approved):

PR #592, opened 2026-07-17
Maint2

ThetaSinner:


crates/core/src/factories/mem_transport.rs
```diff
@@ -595,22 +595,22 @@ impl TransportInstances {
...
-            None => return None,
-            Some(id) => match id.parse() {
+        let to_peer_id: u64 = {
+            let id = to_peer.peer_id()?;
+            match id.parse() {
                 Err(_) => return None,
                 Ok(id) => id,
-            },
+            }
         };
```
veeso: maybe even better:

```rust
let to_peer_id: u64 = to_peer.peer_id()
  .and_then(|id| id.parse().ok())?;
```

veeso (dismissed):

jost-s (dismissed):

crates/core/src/factories/mem_transport.rs
ThetaSinner: That is better, done

veeso (approved):

PR #600, opened 2026-07-27
ci: speed up the Test workflow

ThetaSinner:
The `Test` workflow takes 56-95 minutes per pull request. Two measured causes:

**Everything waited on `static`.** It took 27-28 minutes and every other job declared `needs: static`, so the critical path was `static` (28m) then `test-windows` (28m).

**No pull request could restore a cache.** `test.yaml` ran on `pull_request` only, so every cache it wrote was scoped to `refs/pull/N/merge`. GitHub only lets a pull request restore from its own ref or its base branch, and `main` had no cache for these jobs — so every new branch built cold. Same branch, back to back: 58m34s cold, then 13m43s warm. Repository cache usage was also 11.49 GB against a 10 GB limit, so entries were evicting each other.

## Changes

- **Split `static`** into a `fmt` gate (formatting only, ~0.7s of work, no compilation) plus parallel `clippy` and `doc-check` jobs. Measured with `cargo --unit-graph`, the three configurations share *zero* compilation units — 1090 + 711 + 1087 = 2888, exactly the total distinct count — because each differs in toolchain, feature set, or profile. Splitting them duplicates no work.
- **Dropped `bench-build` from the pull-request path.** It was 1087 units, 910 of them at opt-level 3, forked into all 11 crates — but only `bootstrap_srv` has benchmarks. `bench.yaml` on `main` already builds and runs every bench target.
- **Seeded caches from `main`** via a `push: branches: [main]` trigger, with `save-if` making pull requests restore-only so one authoritative set fits the 10 GB budget.
- **Added a concurrency group.** `cancel-in-progress` is conditional on `pull_request` — cancelling a `main` run would abort the cache-seeding job this depends on.
- **Gated `changelog-preview-comment`** to `pull_request`, since it posts onto a pull request and the workflow now also runs on `main`.

Per-crate test invocation is deliberately kept: measured at 1090 distinct units against 947 for a unified workspace build, so the feature isolation costs 15%, not a multiple. It catches bugs where a crate only compiles because a sibling's dev-dependencies enable a feature.

## Expected

~30 min cold, ~18-25 min warm. Note the payoff starts one merge late — the merge commit's own `main` run restores nothing, because no `main`-scoped cache exists yet, and will take the full cold path.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -4,7 +4,21 @@ on:
   pull_request: {}
 
 jobs:
-  static:
+  fmt:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-make@0.37.24,taplo@0.9.3
+
+      - name: Format
+        run: cargo make static-fmt
+
+  clippy:
+    needs: fmt
     runs-on: ubuntu-latest
 
     steps:
@@ -22,11 +36,33 @@ jobs:
         with:
           cmake-version: '3.31.x'
 
-      - name: Static
-        run: cargo make static
+      - name: Clippy
+        run: cargo make clippy
+
+  doc-check:
+    needs: fmt
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-make@0.37.24
+
+      - uses: Swatinem/rust-cache@v2
+
+      # we cannot use cmake 4 until openssl-src is updated
+      - name: cmake
+        uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.31.x'
+
+      - name: Doc check
+        run: cargo make doc-check
 
   test:
-    needs: static
+    needs: fmt
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest]
@@ -72,7 +108,7 @@ jobs:
         run: cargo run --example schema --features schema
 
   test-windows:
-    needs: static
+    needs: fmt
     strategy:
       matrix:
         os: [windows-latest]
@@ -114,7 +150,7 @@ jobs:
           cargo make test
 
   nix-build:
-    needs: static
+    needs: fmt
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest]
@@ -135,7 +171,7 @@ jobs:
       - run: nix build --verbose -L .#bootstrap-srv
 
   docker-build-bootstrap:
-    needs: static
+    needs: fmt
     runs-on: ubuntu-latest
 
     steps:
@@ -145,7 +181,7 @@ jobs:
         run: docker buildx build -f docker/kitsune2_bootstrap_srv/Dockerfile .
 
   docker-build-bootstrap-auth-test:
-    needs: static
+    needs: fmt
     runs-on: ubuntu-latest
 
     steps:
@@ -162,7 +198,9 @@ jobs:
     if: ${{ always() }}
     runs-on: "ubuntu-latest"
     needs:
-      - static
+      - fmt
+      - clippy
+      - doc-check
       - test
       - test-windows
       - nix-build
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,6 +2,13 @@ name: Test
 
 on:
   pull_request: {}
+  push:
+    branches:
+      - main
+
+concurrency:
+  group: test-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
   fmt:
@@ -29,6 +36,8 @@ jobs:
           tool: cargo-make@0.37.24
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
 
       # we cannot use cmake 4 until openssl-src is updated
       - name: cmake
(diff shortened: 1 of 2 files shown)


Makefile.toml
```diff
@@ -70,6 +70,11 @@ args = ["format"]
 # Helper tasks
 #
 
+[tasks.static-fmt]
+description = "Formatting checks only, with no compilation. Used as a fast CI gate"
```
ThetaSinner:
```suggestion
description = "Formatting checks only, with no compilation"
```

ThetaSinner: Applied in 6e9aff6d (fixup targeting 932194dd).

Agreed — every other description in the file is a single clause with no internal period (`"Run all static checks"`, `"Build all benchmarks without running them so they don't bit-rot"`), so the second sentence was the odd one out. Dropping the "fast CI gate" clause loses nothing: the workflow job that consumes it is named `fmt` and the commit message covers the rationale.

Checked the rest of the branch for the same pattern — this was the only description added, and the `test.yaml` step names (`Format`, `Clippy`, `Doc check`) already match the existing `Test`/`Build`/`Examples` style.

.github/workflows/test.yaml
```diff
@@ -2,9 +2,54 @@ name: Test
...
-  static:
+  fmt:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-make@0.37.24,taplo@0.9.3
```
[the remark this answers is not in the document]
ThetaSinner: Not actioning in this PR — the finding is legitimate but out of scope, and one half of the suggestion would break a working job.

**On `persist-credentials: false`:** the point stands, and nothing in this workflow does git operations after checkout, so it would be safe. But it applies to all nine `actions/checkout` steps in the file, six of which this PR doesn't touch. The three new ones (`fmt`, `clippy`, `doc-check`) are copies of the file's existing pattern. This is a pre-existing posture issue across the repo's workflows, not something this change introduces.

**On repository-level `permissions: contents: read`:** this would break `changelog-preview-comment`. That job calls the reusable workflow `holochain/actions/.github/workflows/changelog-preview-comment.yml@v1.16.0`, which declares no `permissions:` block of its own — I checked the v1.16.0 source, it only declares `on: workflow_call:` with inputs. Reusable workflows without their own `permissions:` inherit the caller's, so a repo-level `contents: read` would strip the write scope it needs to post its comment. Making this work requires adding a scoped `permissions:` to that job and verifying it still posts — real work that deserves its own PR rather than riding along on a performance change.

Worth doing as a dedicated workflow-hardening PR covering all nine checkouts and the permissions model together, where the changelog job's behaviour can actually be verified.

.github/workflows/test.yaml
```diff
@@ -56,6 +103,8 @@ jobs:
           tool: cargo-make@0.37.24
 
       - uses: Swatinem/rust-cache@v2
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' }}
```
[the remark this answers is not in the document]
ThetaSinner: No change — this PR doesn't touch the cachix configuration and doesn't alter the exposure described.

The only edit to `nix-build` in this branch is `needs: static` -> `needs: fmt`. The `cachix/cachix-action@v17` block, including `authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}`, is byte-for-byte unchanged — `git diff c3870c5d..HEAD -- .github/workflows/test.yaml` contains no `cachix` hunk.

On the exposure itself: `nix-build` already ran on every pull request before this change, so PR-time push behaviour is exactly as it was. What this PR adds is `push:` on `main`, which is a *more* trusted context, not a less trusted one. And secrets aren't exposed to `pull_request` runs from forks, so the surface is same-repo pull requests — already the case beforehand.

Whether `holochain-ci` should accept pushes from PR builds is a project policy question. Warming a shared Nix binary cache from PR builds is a common deliberate choice, and changing it would alter cache-population behaviour for everyone. If the team does want PR runs to be pull-only, `skipPush: ${{ github.event_name == 'pull_request' }}` is the change — but it belongs in its own PR with that trade-off stated, not folded into a CI performance change.

.github/workflows/test.yaml
```diff
@@ -2,9 +2,54 @@ name: Test
 
 on:
   pull_request: {}
```
veeso, line 4: is this correct? What does it mean actually?

usually I see always

```yml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - main
```

veeso (commented):
The idea is good, but I'm still not sure why we can't run tests without running `fmt` first. Coulnd't we just run all what can run in parallel?

Ok, it reads it takes 0.7s so it's okay, but yeah, couldn't we just run without any previous task? Also because I think there is some time wasted on the worker spin up if you put the block.

.github/workflows/test.yaml
ThetaSinner: My understanding of what `pull_request: {}` is "a pull request with any base branch".

If do find that helpful because if you stack branches then with a filter for `main` as the base branch, the checks don't run on the stacked branches. Certainly an option and possibly less wasteful.

I can change it if you prefer the more explicit and specific filter?

ThetaSinner:
> not sure why we can't run tests without running fmt first. Couldn't we just run all what can run in parallel?

Absolutely, we could. The reason we ran static checks first was to avoid the expense of running tests if the static checks were going to fail. Actually looking at the timings, I think formatting and clippy could be checked. I assumed clippy was slower too but building the docs and benchmarks is most of the time spent.

I'm okay with not fully optimizing if it means we can save claiming macos runners which we have a limited amount of concurrency for across the org. Yes it takes a bit of time up front, I'm reasonably happy with that.

With the caching changes and structure change, it should take 20 minutes to run CI instead of nearly 60. That's still a bit slower than I'd like but the tests take some time and that's more work to fix.

.github/workflows/test.yaml
veeso: it's okay

veeso (approved):

Issue #263, opened 2025-07-11
Add `Access` module


ThetaSinner:
In order to prevent read access to spaces which the existing participants want to protect, we are going to need to add access control that operates before starting to sync data with a peer.

Note that spaces in Kitsune2 are multi-tenant. Multiple agents can join a space in a single instance. When Kitsune2 communicates with remote peers, it generally doesn't care whether it's one agent or many on the other end. However, for the purposes of access, it is agents that are the actors who can provide authentication. It is also important to understand that if we sync data with one agent then we have no expectation that the data is restricted to that agent. The data is available to all agents on the remote. In which case, the definition of access is that we grant it to an agent but granting access to that agent actually opens up access to the peer where that agent is running.

Add a new module which can actively request authentication and record the results.

In order to discover which agents we need to challenge for authentication, we need to know when new agents are discovered. Adding a subscription method to the peer store would give us this ability. Similarly to the way that we subscribe to "drained" notifications on the fetch module.

The new access module will subscribe to new peers, or more accurately newly inserted peer infos even if we already knew about the agent, and check whether that agent already has access. If the agent doesn't have access then an access challenge is sent to them. This is a module message within the access module.

On receiving a challenge, the access module will respond with access credentials for the agent being challenged. For this to be a safe thing to do, the access credentials *must not* be transferable. Otherwise, it would be possible to send challenge requests to agents who are already on the network and get them to give you their access credentials which you could then use. (Given that these values have always been published to the DHT in Holochain, this was actually always a restriction).

- If the agent fails to respond, the access module will not retry. The retry will happen when the unresponsive agent produces a new agent info.
- If the agent responds then the access credentials are passed to the Kitsune2 host to be validated.
  - If validation passes, the access module records that the agent has access.
  - If validation fails, it is up to the host to decide what to do. The host may choose to block the agent.

When a new agent info shows up for an agent that already has access, then access is not re-checked. The challenge/respond mechanism is bypassed and the agent's access state is not changed.

The access mechanism will be an opt-out mechanism per space. The module builder should have an implementation that will come from K2 core by default and can be overriden by the host. There should also be a core no-op implementation of the module.

Suggested trait for the host to implement for checking access:

```rust
pub trait AgentAccess {
  /// Provide access credentials in response to a challenge.
  fn provide_access(&self, space_id: SpaceId, agent: AgentId) -> BoxFut<'static, K2Result<Bytes>>;

  /// For a space that does use access control, check the response of a challenge.
[...]

PR #615, opened 2026-08-20
docs: add design documents directory and space handshake design

synchwire:
Starts a `docs/design/` directory, and adds the first document in it: a per-space handshake to replace connection preflight as the way peers exchange agent information. Design only, no code changes.

The problem it addresses is that preflight runs once per connection and is not space-aware, so any space that starts after a connection was established never learns the remote's agents for that space, and its traffic is then silently discarded on arrival as if the peer were blocked. Intended to also be the place access control lands later, per #263.

A related host-side problem turned up during the analysis and is filed separately as holochain/holochain#5962.

> Disclosure per AI_POLICY: drafted with AI assistance. The design decisions were made in discussion, not generated.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -1,0 +1,53 @@
+# Design documents
+
+This directory holds design documents for Kitsune2 features.
+
+A design document describes **what** a feature does and **why**, in enough
+detail that two people implementing it independently would produce compatible
+results. It is a living document: keep it current as the feature evolves,
+rather than freezing it at the moment a decision was made.
+
+## Index
+
+| Document | Status | Summary |
+| --- | --- | --- |
+| [Space handshake](space-handshake.md) | Accepted | A per-space handshake that replaces connection preflight as the way peers exchange agent information. |
+
+## Writing a design document
+
+Use a descriptive file name, such as `space-handshake.md`, and add a row to the
+index above. File names are not numbered; the index carries ordering and
+status.
+
+Suggested sections:
+
+- **Status** — Draft, Accepted, or Superseded by another document. Status
+  describes the state of the decision, not the state of the code: an Accepted
+  document may not be implemented yet, or may be only partly implemented.
+- **Problem** — what is broken or missing, and why it matters.
+- **Goals and non-goals** — non-goals must be things a reader could reasonably
+  expect the goals to cover. Anything obviously outside the feature does not
+  belong here; listing it only adds noise.
+- **Design** — the mechanism, described as observable behaviour and the
+  invariants that must hold. Approaches that were considered and rejected go
+  here too, when knowing why something was not done helps an implementer.
+- **Edge cases** — the situations that are easy to get wrong.
+- **Security considerations**.
+- **Compatibility** — what changes for peers already running, particularly on
+  the wire.
+- **Open questions** — only while a document is a Draft. Everything here must
+  be resolved before the design is accepted, so an Accepted document has no
+  open questions left in it.
+
+Some conventions that keep these documents useful:
+
+- Describe behaviour, not code. No references to files, functions, or line
+  numbers; they rot immediately and they tie the design to one implementation
+  of it.
+- No implementation plans, task breakdowns, or descriptions of tests to write.
+  Those belong in issues and pull requests.
+- No release planning. Which branch a feature lands on, and how it is released,
+  is not a property of the feature.
+- State invariants explicitly. "The receiver must have recorded the agents
+  before it processes the next message from that peer" is the kind of sentence
+  that makes two implementations agree.
--- a/docs/design/space-handshake.md
+++ b/docs/design/space-handshake.md
@@ -1,0 +1,346 @@
+# Space handshake
+
+**Status:** Accepted
+
+## Problem
+
+Kitsune2 nodes are multi-tenant. A single node can run many spaces at once, and
+the transport keeps one connection per remote peer URL that all of those spaces
+share.
+
+Agent information is currently exchanged in the connection preflight. The
+preflight runs exactly once, when the connection is established, and it is not
+space-aware — it is produced by the host at the level of the whole node, with
+no knowledge of which spaces the connection will end up carrying.
+
+Blocking is enforced per space, by looking at the agents known to be at a peer
+URL. A peer URL with no known agents in a space is treated as blocked for that
+space, and its messages are dropped silently.
+
+The loss is one-directional. The side that starts a conversation necessarily
+already knows an agent at the peer in that space, because that is how it
+learned the URL, so it is never the side that drops. It is the receiver that
+discards what arrives.
+
+Putting those together: the first space to reach a peer causes a connection to
+be established and a preflight to be exchanged. Every space that starts after
+that point reuses the same connection, so no further preflight happens, and the
+peer never learns that space's agents. Traffic for those spaces is discarded on
+arrival. Nothing reports an error; the sender sees its messages accepted and the
+receiver never sees them.
+
+The condition clears only if the receiving space happens to discover the
+sender's agents some other way, in practice by polling the bootstrap server.
+That can take minutes.
+
+Sharing several spaces with the same peer is a supported case rather than an
+unusual one, so this affects ordinary operation. The observable symptoms are
+gossip rounds that are initiated and never answered, and peers that appear
+reachable but exchange nothing.
+
+**The underlying mismatch is that a connection-scoped, node-level mechanism is
+being used to satisfy a per-space requirement.** Everything below follows from
+moving that responsibility to where it belongs.
+
+## Goals
+
+- Give every space its own handshake with a remote peer, independent of when
+  the underlying connection was established or how long it lives.
+- Ensure that before a space sends anything to a peer, that peer knows at least
+  one agent the space has, so that block enforcement has something to act on and
+  does not fall back to dropping everything.
+- Work correctly when a space starts after a connection already exists, and
+  when a space is torn down and re-created while the connection persists.
+- Preserve block enforcement exactly as it is. The handshake makes enforcement
+  work correctly across multiple spaces; it must not make it more permissive.
+- Live in shared code, so that transport implementations need no knowledge of
+  it.
+- Leave room for access control to be added to the same exchange later.
+
+## Non-goals
+
+- Authentication or access control. The handshake shares agent information; it
+  establishes nothing about whether the peer should be in the space. Deciding
+  that is a separate concern to be added later.
+- Connection management. How connections are established, how failures are
+  handled, and how gossip chooses targets are all unchanged. In particular, the
+  handshake does not address peers that are slow or impossible to connect to.
+
+## Design
+
+### Shape
+
+The handshake is performed by a per-space module. Every space has one. Kitsune2
+provides a default implementation that all nodes can use; a host may substitute
+its own, but is not expected to have to.
+
+It is tracked once between a space and a remote peer URL, not once per agent.
+Kitsune2 identifies remote peers by URL throughout, and the block enforcement it
+feeds is expressed per URL, so the handshake is tracked the same way.
+
+The **handshake message** carries the sending space's signed agent infos. It
+travels within the space as a module message under a reserved module identifier,
+encoded as protobuf so that it can be extended later.
+
+There is no reply. A handshake message is not a request, nothing is returned for
+one, and a sender never waits on one. Both sides send a handshake message, for
+the same reason and under the same rule: a space sends one before it first sends
+anything else to a peer URL. Which side goes first is whichever space wants to
+talk first, and if both want to at once, both send.
+
+This works because the requirement is one-directional. A space needs the peer it
+is about to talk to to know its agents; it needs nothing back in order to send
+safely. The reverse direction is the peer's own concern, and the peer satisfies
+it with its own handshake message when it next has something to send.
+
+Little is lost by having no reply. A space can only address a peer URL it
+discovered, and discovering it means holding an agent info that carries that
+URL, so a sender already knows an agent at the peer it is sending to. The side
+that can be ignorant is the one receiving an unsolicited first message, and that
+is exactly the side a handshake message informs. What a reply would have added
+is a refresh of the sender's possibly stale view of the peer, and the peer's own
+handshake message supplies that as soon as it has something to say.
+
+Agent information is additive, exactly as it is when it arrives from any other
+source. A handshake message supplies agents; it never removes them, and it is
+not a snapshot that supersedes what the receiver already knows. Superseding an
+existing record follows the same rules that apply to agent information
+generally.
+
+A handshake message carrying no agent infos cannot serve its purpose and is
+dropped. So is one that is malformed, oversized, or otherwise not understood.
+
+### Exemption from block enforcement
+
+Messages addressed to the handshake module are exempt from the rule that drops
+traffic to and from a peer URL with no known agents in the space. Without this
+the handshake could never run: it exists precisely to resolve that state.
+
+The exemption is narrow, and two limits define it.
+
+It applies only to the reserved module identifier used by the handshake.
+Everything else addressed to that peer in that space is still dropped until the
+handshake has supplied agents.
+
+It does not extend to peers that are known to be blocked. If any agent known at
+that URL is blocked in this space, handshake messages are dropped along with
+everything else. The exemption covers "we do not know who is there yet", never
+"we know who is there and have refused them".
+
+One path narrows as a result. Agent information can today arrive from a URL with
+a blocked agent at it, because that information travels in the preflight and the
+preflight is always allowed. Once it travels in the handshake instead it no
+longer can, because the exemption stops at URLs known to be blocked and such a
+URL receives nothing. Nothing changes in practice: one blocked agent at a URL
+blocks the whole URL, so learning of another agent there would not have changed
+the decision.
+
+### Handshake state
+
+Each space keeps, in memory, the set of peer URLs it has sent a handshake
+message to.
+
+A URL is marked once the handshake message has been sent successfully, and not
+before. If the send fails, the URL stays unmarked and the space sends a
+handshake message again the next time it has something to send to that peer.
+Nothing else gates the mark, because there is no acknowledgement to wait for.
+
+A successful send is not delivery. It means the transport accepted the message,
+not that the peer received it, so a handshake message can still be lost with the
+connection that was carrying it. The space therefore clears a URL from the set
+when the connection to it is lost, whether or not the loss had anything to do
+with a handshake message, and the next send to that peer sends one again over
+the new connection. Without this, a handshake message lost in transit would
+leave the URL marked and every later message to that peer silently discarded,
+with nothing left to trigger a repair.
+
+Clearing on connection loss is right for the same reason clearing on restart is:
+a peer reached over a new connection may be a peer that restarted and lost
+everything it knew.
+
+Receiving a handshake message does not mark the sender's URL. The two directions
+are independent: what a space records is that it has told a peer about its own
+agents, never what a peer has told it. A space that has received a handshake
+message from a peer it has not yet sent one to still sends its own before it
+first speaks.
+
+**A repeat handshake message for a URL already marked must be accepted and
+processed normally. It is never rejected.** This is the invariant that removes
+the need to treat restarts as a special event. Whichever side lost its state
+simply sends again, and the side that kept its state absorbs the repeat
+harmlessly. Neither side needs to detect that the other restarted, and there is
+only one code path to get right.
+
+The same invariant makes concurrency a non-event. Two sends racing towards the
+same unmarked URL may each send a handshake message; the second is absorbed like
+any other repeat, and the only cost is one redundant message. Suppressing the
+duplicate is permitted but is not required for correctness.
+
+The state is not persisted. It belongs to the space rather than to the
+connection — a space that is torn down and re-created starts with an empty set
+even where the connection survived, which is correct, because it has also lost
+the agents it learned — but it does not outlive a connection either, for the
+reason above.
+
+Note that the state cannot be derived from whether the space knows any agents at
+the URL. A sender always knows some — that is how it found the peer — so a
+derived check would never fire on the side that needs it. The state has to be
+recorded explicitly.
+
+### When the handshake runs
+
+The handshake is lazy. It runs when a space is about to send a message to a peer
+URL it has not sent a handshake message to. Nothing is sent speculatively, and
+no connection is opened for the sake of a handshake alone. Handshake traffic
+therefore scales with how much a node actually communicates, not with how many
+peers it has discovered.
+
+Because sending is the trigger, the check belongs at the point where a space
+hands a message to the transport, rather than inside each module that wants to
+send. Gossip, publish, fetch, agent info broadcast and host notifications are
+then all covered without any of them knowing the handshake exists.
+
+If the handshake message cannot be sent, the triggering send fails. The
+handshake has no retry loop of its own; the callers that trigger it already
+retry on their own schedules, and letting them pace it avoids a second,
+independent source of traffic aimed at a peer that may be unreachable.
+
+### Ordering
+
+The triggering message is sent immediately after the handshake message, on the
+same connection. The sender does not block, and has nothing to block on, because
+nothing comes back.
+
+Two invariants make this safe, and both must hold.
+
+**The sender must issue the handshake message and the triggering message in
+order, not concurrently.** Concurrent sends may be interleaved on the
+connection, and the triggering message arriving first would simply be dropped.
+
+**The receiver must finish applying a handshake message — recording the agents
+and updating whatever state block enforcement consults — before it processes the
+next message from that peer.** A handshake message that is accepted and then
+applied in the background reintroduces exactly the race the handshake exists to
+remove: the message behind it is checked for permission before the agents have
+landed, and is dropped. Accepting a handshake message and applying it must be
+one step from the point of view of message processing.
+
+Both invariants rest on a third, which is a requirement the handshake places on
+the transport rather than on either peer: **messages sent to a peer URL must be
+delivered in the order the space handed them over.** Issuing in order and
+applying in order buy nothing if the path between may reorder. This is the only
+demand the handshake makes of a transport, and it is worth stating plainly
+because nothing else in the design depends on transport behaviour at all.
+
+### Relationship to preflight
+
+Preflight stops carrying agent information. That responsibility moves entirely
+into the handshake message.
+
+Preflight is still exchanged, and it is still exempt from block enforcement, but
+for a different reason than it is today. Today it is exempt because it is the
+only way to learn about new agents at a peer URL. Afterwards it is exempt
+because it is not space-scoped: it is exchanged before any space traffic, so
+there is no space in which a block could be evaluated.
+
+What stays in preflight is connection-level concerns whose correct outcome is to
+reject the whole connection, such as protocol compatibility checks. Because
+those checks happen before any space traffic, the handshake never has to
+consider peers running a known-incompatible protocol.
+
+Two things follow. The host no longer has to assemble agent information across
+spaces for a connection whose eventual use it cannot predict. And the
+requirement that a local agent must already have joined before a preflight can
+be produced disappears, along with the special handling that requirement needs
+today.
+
+### Approaches ruled out
+
+**Treat "no agents known at this URL" as permitted rather than blocked.** This
+removes the symptom by removing the enforcement, and it makes blocking
+unreliable in exactly the situation where it is supposed to work. It also
+leaves the underlying problem — that a space may never learn who it is talking
+to — in place.
+
+**Make the exchange a request and a response.** An explicit reply would confirm
+that the peer applied what it was sent, and would refresh the sender's view of
+the peer at the same time. It costs a second message kind, a rule for matching
+responses to outstanding requests, and a decision about what to do with a
+response that matches nothing. It buys little in return, because a sender
+already knows an agent at any peer it is able to address, and the peer sends its
+own handshake message as soon as it has something to say.
+
+## Edge cases
+
+**A handshake message arrives for a space the receiver does not have.** Drop it.
+This should not normally occur: discovery is space-scoped, so a sender only
+learns a URL in the context of a space both sides have joined. It can happen
+transiently when the sender is acting on an agent info that is still cached
+after the remote removed the space.
+
+**The peer is unreachable.** This surfaces as an ordinary transport failure and
+the existing handling for unresponsive peers applies. The handshake adds
+nothing.
+
+**The handshake message is sent but never applied.** A successful send is not
+proof that the peer recorded the agents: it may have dropped the handshake
+message as malformed or oversized, or not have the space at all. The sender
+marks the URL regardless and is not told. Traffic behind the handshake message
+is then discarded on arrival, which is the original symptom in miniature. It is
+accepted because every case that reaches here is one where that traffic had no
+prospect of being useful anyway — the peer does not have the space, has blocked
+the sender, or cannot parse what the sender emits. Loss in transit is not one of
+those cases, and is not left here: a handshake message that goes down with its
+connection is recovered by clearing the URL when the connection is lost.
+
+**The peer blocks the sender.** The handshake message is dropped, and the
+sender's traffic to that peer continues to be dropped. The sender is not told,
+which is the intended behaviour for blocking.
+
+**A space has no local agents.** It has nothing to put in a handshake message,
+so it does not send one, and the send that would have triggered it fails as it
+does today. A handshake message arriving while a space is in this state is still
+applied, because the sender's agents are worth recording.
+
+**Both sides send a handshake message at the same time.** Both are applied.
+Neither is a reply to the other and neither has to be matched to anything, so
+simultaneity is not a case that needs handling.
+
+**A space is torn down and re-created while the connection persists.** The
+restarted space sends a handshake message again the next time it wants to send,
+and the peer that kept its state accepts the repeat. The gap is in the other
+direction: the peer still believes it has told the restarted space about its
+agents, and the restarted space has lost them, so the peer's traffic is
+discarded on arrival until the restarted space rediscovers it. That is a
+narrower window than the problem this design removes, but it is not zero.
+
+## Security considerations
+
+The handshake module is the first thing an unknown peer can reach in a space,
+so it is the exposed surface and should be treated as such. It should accept
+only well-formed messages and bound both the message size and the number of
+agent infos a single message may carry.
+
+Bounding how often a peer may send handshake messages is a separate concern and
+is not addressed here. Kitsune2 meters no message type per peer today, and
+metering one type alone would not amount to much protection.
+
+Agent information is verified as it is decoded from the wire, before anything
+is recorded, which is where verification happens for agent information from any
+source. Verification is a property of producing the value at all rather than a
+step someone downstream could forget: unverified agent information is not
+something that exists to be passed on, so nothing further along has to re-check
+it.
+
+The exemption does not weaken blocking. A peer URL with a known blocked agent
+in the space receives nothing at all, handshake included. The exemption applies
+only where the space has no information about who is at the URL.
+
+Recording agents from a handshake confers nothing that discovering the same
+agents through bootstrap would not. The handshake changes how a space learns
+about agents, not what learning about them permits.
+
+The handshake does not authenticate. Receiving a handshake message shows only
+that the sender holds agent information for the space, which is public. It is
+the point at which authentication should later be added, and the exchange is
+shaped to allow that, but on its own it establishes no trust.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -6,6 +6,14 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Kitsune2 is a peer-to-peer / DHT communication framework written in Rust, used by Holochain. It is structured as a Cargo workspace of crates that together provide a node implementation, plus a bootstrap server for WAN peer discovery.
 
+## Design documents
+
+`docs/design/` holds design documents for Kitsune2 features. Read the relevant one before working on an area it covers; they explain what a feature does and why, which is context the code alone does not carry. Treat them as a source of information about intended behaviour, alongside the code.
+
+`docs/design/README.md` indexes the documents and records their status. Status describes how settled the design is, not whether it is built: a `Draft` may still change, and an `Accepted` document may not be implemented yet, or only partly. Always check a document against the code rather than assuming the code matches it.
+
+These are living documents. If a change alters behaviour that a design document describes, update the document in the same change. They deliberately avoid references to files, functions, and line numbers, and they do not contain implementation plans or descriptions of tests; keep it that way when editing them.
+
 ## Common commands
 
 This project uses `cargo-make`, configured via `Makefile.toml`, for the standard developer workflow.


docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+Agent information is currently exchanged in the connection preflight. The
+preflight runs exactly once, when the connection is established, and it is not
+space-aware — it is produced by the host at the level of the whole node, with
+no knowledge of which spaces the connection will end up carrying.
+
+Blocking is enforced per space, and it is enforced by looking at the agents
+known to be at a peer URL. A peer URL with no known agents in a space is
+treated as blocked for that space, and messages are dropped silently in both
+directions.
```
ThetaSinner: It's always only in one direction because the initiating side must have known about exactly one agent+peer_url to initiate the connection. It's the receiver that may drop incoming messages.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+Putting those together: the first space to reach a peer causes a connection to
+be established and a preflight to be exchanged. Every space that starts after
+that point reuses the same connection, so no further preflight happens, and it
+never learns the remote's agents for its own space. Those spaces are treated as
+blocked and their traffic disappears. Nothing reports an error; the sender sees
+messages accepted and the receiver never sees them.
+
+The condition clears only if the space happens to discover the remote's agents
+some other way, in practice by polling the bootstrap server. That can take
+minutes, and it does not happen at all when bootstrap is unavailable.
```
ThetaSinner:
```suggestion
minutes, and it does not happen at all when bootstrap if unavailable.
```

Bootstrap would usually be available. We generally assume that's up.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+Running several spaces against the same peer is the normal case rather than an
```
ThetaSinner:
```suggestion
Running several spaces in common the same peer is a supported case rather than an
```

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+unusual one, so this affects ordinary operation. The observable symptoms are
+gossip rounds that are initiated and never answered, and peers that appear
+reachable but exchange nothing.
+
+The underlying mismatch is that a connection-scoped, node-level mechanism is
+being used to satisfy a per-space requirement.
```
ThetaSinner: This is really the key statement - can we make it stand out a bit more?

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+  on and does not fall back to dropping everything.
+- Work correctly when a space starts after a connection already exists, and
+  when a space is torn down and re-created while the connection persists.
+- Live in shared code, so that transport implementations need no knowledge of
+  it.
+- Leave room for access control to be added to the same exchange later.
+
+## Non-goals
+
+- Authentication or access control. The handshake proves only that a peer knows
```
ThetaSinner: Careful with "proves". It shares information, it doesn't actually prove anything I think?

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+  an agent in the space. Deciding whether a peer is *allowed* in the space is a
+  separate concern to be added later.
+- Peer discovery. Learning that a peer URL exists remains the job of bootstrap
+  and gossip. The handshake is what happens once a space already knows a URL
+  and wants to use it.
```
ThetaSinner: This reads as redundant to me?

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+- Connection management. How connections are established, how failures are
+  handled, and how gossip chooses targets are all unchanged.
+- Relaxing block enforcement. The handshake makes enforcement work correctly for
+  multiple spaces; it does not make it more permissive.
```
ThetaSinner: This seems like a goal to me rather than a non-goal, to NOT make it more permissive

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Design
+
+### Shape
+
+The handshake is performed by a per-space module. Every space has one. Kitsune2
+provides a default implementation that all nodes can use; a host may substitute
+its own, but is not expected to have to.
+
+The exchange is between a local space and a remote peer URL. It is not per
```
ThetaSinner: Don't like the phrasing, it's negotiated between spaces once, not per agent - is what you're trying to say.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+the peer — so a derived check would never fire on the initiating side. The
+state has to be recorded explicitly.
+
+### When the handshake runs
+
+The handshake is lazy. It runs when a space is about to send a message to a
+peer URL it has not completed the handshake with. Nothing is sent
+speculatively, and no connection is opened for the sake of a handshake alone.
+Handshake traffic therefore scales with how much a node actually communicates,
+not with how many peers it has heard of.
```
ThetaSinner:
```suggestion
not with how many peers it has discovered.
```

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+and is dropped. Accepting the request and applying it must be one step from the
+point of view of message processing.
+
+### Relationship to preflight
+
+Preflight stops carrying agent information. That responsibility moves entirely
+into the handshake.
+
+Preflight remains for connection-level concerns that are genuinely not
+space-scoped, such as protocol compatibility checks whose correct outcome is to
```
ThetaSinner:
```suggestion
Preflight remains for connection-level concerns that are not
space-scoped, such as protocol compatibility checks whose correct outcome is to
```

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+**Both sides begin a handshake at the same time.** Both requests are processed
+and both responses are sent. The exchange is idempotent, so the duplicate work
+is harmless.
+
+## Security considerations
+
+The hello module is the first thing an unknown peer can reach in a space, so it
+is the exposed surface and should be treated as such. It should accept only
+well-formed messages, bound both the message size and the number of agent infos
+a single message may carry, and verify agent info signatures before recording
```
ThetaSinner: Should it verify agent infos itself or delegate that to the peer store?

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+leaves the underlying problem — that a space may never learn who it is talking
+to — in place.
+
+**Handshake eagerly, on discovery.** Opens connections to peers that may never
+be used, and makes handshake traffic scale with the number of peers known
+rather than with communication actually taking place.
+
+**Derive handshake state from whether the space knows any agents at the URL.**
+Attractive because it needs no new state, but wrong: the initiator always knows
+agents at the URL already, so it would never handshake.
```
ThetaSinner: Drop the section

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Open questions
+
+- Whether the response should be able to signal an explicit rejection, and what
+  a sender should do with one.
```
ThetaSinner: I don't think so, even once we add an access check, a failure should use minimal resources to reject the bad request. Dropping it is as soon as we know it's invalid is the fastest way. The protocol version in the preflight should already prevent known incompatible peers from communicating.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+- Whether the name "hello" is the one to keep, given the module is likely to
+  grow into the place where access is negotiated.
```
ThetaSinner: No, looking at this document, we're refering to "handshake" in many places. Should we drop the hello terminology and just stick with handshake? Including in the design file name

docs/design/README.md
```diff
@@ -0,0 +1,48 @@
...
+- **Goals and non-goals** — the non-goals matter most; they are what keeps a
+  design from quietly growing.
+- **Design** — the mechanism, described as observable behaviour and the
+  invariants that must hold.
+- **Edge cases** — the situations that are easy to get wrong.
+- **Security considerations**.
+- **Compatibility** — what changes for peers already running, particularly on
+  the wire.
+- **Alternatives considered** — and why they were not chosen.
+- **Open questions**.
```
ThetaSinner: These should all be resolved before a design is accepted by PR.

docs/design/README.md
```diff
@@ -0,0 +1,48 @@
...
+- **Problem** — what is broken or missing, and why it matters.
+- **Goals and non-goals** — the non-goals matter most; they are what keeps a
+  design from quietly growing.
+- **Design** — the mechanism, described as observable behaviour and the
+  invariants that must hold.
+- **Edge cases** — the situations that are easy to get wrong.
+- **Security considerations**.
+- **Compatibility** — what changes for peers already running, particularly on
+  the wire.
+- **Alternatives considered** — and why they were not chosen.
```
ThetaSinner: Drop this, it's rarely needed. If it is, it can go in the Design section

docs/design/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+Use a descriptive file name, such as `space-hello-handshake.md`, and add a row
+to the index above. File names are not numbered; the index carries ordering and
+status.
+
+Suggested sections:
+
+- **Status** — Draft, Accepted, or Superseded by another document.
+- **Problem** — what is broken or missing, and why it matters.
+- **Goals and non-goals** — the non-goals matter most; they are what keeps a
```
ThetaSinner: Non-goals should be strictly related to the goals, no completely unrelated information that is clearly out of scope to a reader.

docs/design/space-hello-handshake.md
synchwire: Right, corrected. The rule applies in both directions, but the failure never does: the initiating side already knows an agent at the peer, because that is how it learned the URL, so it is never the side that drops. The document now says the loss is one-directional and that it is the receiver that discards what arrives.

docs/design/space-hello-handshake.md
synchwire: Dropped the claim. Your suggestion had a typo in it so I applied the intent rather than the text — the sentence now just says recovery can take minutes, and asserts nothing about bootstrap being down.

docs/design/space-hello-handshake.md
synchwire: Applied the intent; the literal suggestion did not quite parse. Now reads "Sharing several spaces with the same peer is a supported case rather than an unusual one".

docs/design/space-hello-handshake.md
synchwire: Bolded it and gave it a closing clause so it reads as the conclusion the rest of the document follows from, rather than a trailing remark.

docs/design/space-hello-handshake.md
synchwire: Agreed, it proves nothing. The non-goal now says the handshake shares agent information and establishes nothing about whether the peer should be in the space. Fixed the same overclaim in Security considerations, which said "demonstrates".

docs/design/space-hello-handshake.md
synchwire: Dropped, along with applying the rule you gave on the README about what belongs in non-goals.

docs/design/space-hello-handshake.md
synchwire: Agreed. Moved to Goals as "Preserve block enforcement exactly as it is. The handshake makes enforcement work correctly across multiple spaces; it must not make it more permissive".

docs/design/space-hello-handshake.md
synchwire: Reworded to "The handshake is negotiated once between a space and a remote peer URL, not once per agent".

docs/design/space-hello-handshake.md
synchwire: Applied.

docs/design/space-hello-handshake.md
synchwire: Applied.

docs/design/space-hello-handshake.md
synchwire: The module verifies. Agent information is verified when it is decoded from the wire, and the peer store is only ever handed already-verified values, so it is not the place the check can live. I have made that explicit rather than leaving "verify agent info signatures" sitting in the security section as if it were a free-floating requirement.

docs/design/space-hello-handshake.md
synchwire: Dropped. The reasoning survives as one sentence in the Handshake state section, where it explains why the state has to be recorded explicitly rather than derived — that was your argument originally, so I kept it there. Say if you would rather it went entirely.

docs/design/space-hello-handshake.md
synchwire: Removed the open question and wrote your answer into the design instead: malformed, oversized, unsupported or empty messages are dropped with nothing sent in reply, and the cost of rejecting a bad message should be as close to zero as possible. The preflight section now also notes that the protocol compatibility check keeps known-incompatible peers away from the handshake entirely.

docs/design/space-hello-handshake.md
synchwire: Done. "Handshake" throughout, the module is the handshake module, and the file is now `docs/design/space-handshake.md`. The original commit subject still says "hello"; I will reword it when the fixups are squashed.

docs/design/README.md
synchwire: Reworded: open questions belong only in a Draft, and all of them must be resolved before a design is accepted, so an Accepted document has none left in it. The design document in this PR now has no open questions.

docs/design/README.md
synchwire: Dropped from the template. I also folded the remaining alternatives in the design document into the Design section as "Approaches ruled out", since that is what you said should happen when they are needed. Revert me on that if you meant only the template.

docs/design/README.md
synchwire: Reworded the template: non-goals must be things a reader could reasonably expect the goals to cover, and anything obviously outside the feature does not belong because it only adds noise. Applied it to the design document — peer discovery is gone, and only authentication and connection management remain.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+The exchange is between a local space and a remote peer URL. It is not per
+agent. Kitsune2 identifies remote peers by URL throughout, and the block
+enforcement it feeds is expressed per URL, so the handshake is tracked the same
+way.
+
+Two messages, carried as module messages within the space and encoded as
+protobuf so that they can be extended later:
+
+- A **request**, carrying the sending space's signed agent infos.
+- A **response**, carrying the responding space's signed agent infos.
```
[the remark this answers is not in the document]
synchwire: Not adding a wire schema. Field names, discriminators and versioning belong in the `.proto`, and repeating them here would be stale the moment it changed; these documents deliberately avoid that level of detail.

What was genuinely missing was behaviour, so the design now states the reserved module identifier, the two message kinds, that agent information is additive rather than a snapshot, and that malformed, oversized or empty messages are dropped with nothing sent in reply.

docs/design/space-hello-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+Both directions carry agents. This is deliberately redundant: the initiator
+already knows at least one agent at the remote in that space, because that is
+how it learned the URL in the first place, so strictly only the responder needs
+to be told anything. The exchange is two-way regardless, because the
+initiator's knowledge may be stale — the remote may have re-signed, restarted
+its space, or moved URL since it was discovered — and a two-way exchange
+corrects that instead of depending on the initiator's cache being current.
```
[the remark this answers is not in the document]
synchwire: Valid concern, applied without introducing a correlation value. The initiator now marks a URL complete only on a response to a request it has outstanding, so an unsolicited response cannot complete a handshake that never happened.

A correlation ID would add state for no extra safety here: the peer URL is taken from the connection, not from the message, so only the peer itself can send from that URL.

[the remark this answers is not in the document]
synchwire: Partly valid. Added that agent information is additive and never removes anything, following the same rules that apply to agent information from any other source, and that a message carrying no agent infos is dropped.

Not adding snapshot-and-remove semantics. That is not how agent information behaves anywhere else in Kitsune2, and introducing it here would let a handshake delete records that other sources put there.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+
+Preflight remains for connection-level concerns that are genuinely not
+space-scoped, such as protocol compatibility checks whose correct outcome is to
+reject the whole connection.
+
+Two things follow. The host no longer has to assemble agent information across
+spaces for a connection whose eventual use it cannot predict. And the
+requirement that a local agent must already have joined before a preflight can
+be produced disappears, along with the special handling that requirement needs
+today.
```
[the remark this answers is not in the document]
synchwire: Added as an edge case. A space with no local agents has nothing to put in a handshake, so it does not start one and the send that would have triggered it fails as it does today. An inbound request in that state is still worth recording, but cannot be usefully answered, so the handshake stays incomplete on that side until an agent has joined.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,286 @@
...
+connection. A space that is torn down and re-created starts with an empty set,
+which is the correct behaviour: it has also lost the agents it learned, so it
+does need to handshake again.
+
+**A repeat handshake for a URL already marked complete must be accepted and
+processed normally. It is never rejected.** This is the invariant that removes
+the need to treat restarts as a special event. Whichever side lost its state
+simply repeats the exchange, and the side that kept its state absorbs the
+repeat harmlessly. Neither side needs to detect that the other restarted, and
+there is only one code path to get right.
```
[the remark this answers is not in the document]
synchwire: Correct that the document overclaimed. It said restarts were "covered by the rule that repeat handshakes are always accepted", which is only true when the restarted side is the one that sends next.

Rewritten to say plainly that until the restarted space has something to send, the peer still believes the handshake is complete and its traffic is discarded on arrival, and that this window is narrower than the problem the design removes but is not zero.

Not adding a rule that sends a handshake in response to traffic from an unknown URL: that means answering unauthenticated traffic, and it contradicts the handshake being lazy.

docs/design/space-hello-handshake.md
ThetaSinner: I'm pretty sure the peer store will re-verify and not accept invalid content that it is passed but okay.

synchwire: Checked this properly rather than taking my own word for it, and the outcome you expect does hold, but not by re-verification.

A signed agent info has private fields and only two ways to come into existence: signing one locally, or decoding one, which verifies. There is no constructor that skips the check, so the peer store cannot be handed invalid content in the first place and does not re-verify.

Reworded the security section to say that: verification is a property of producing the value at all rather than a step someone downstream could forget.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,305 @@
...
+The handshake is negotiated once between a space and a remote peer URL, not
+once per agent. Kitsune2 identifies remote peers by URL throughout, and the
+block enforcement it feeds is expressed per URL, so the handshake is tracked
+the same way.
+
+Two messages, carried as module messages within the space under a reserved
+module identifier, and encoded as protobuf so that they can be extended later:
+
+- A **request**, carrying the sending space's signed agent infos.
+- A **response**, carrying the responding space's signed agent infos.
```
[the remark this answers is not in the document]
synchwire: Declining the schema again, for the same reason, and noting that the identical finding was withdrawn in the other thread on this PR after the same explanation.

On the README point: it asks for enough detail that two implementations are *compatible*, and immediately after that it rules out code-level detail. Field names and discriminators live in the `.proto`, which is the authoritative schema; copying them here creates a second source of truth that goes stale on the first change.

On correlation: there is one outstanding request per space and peer URL, so a delayed response cannot complete "the wrong" handshake — any response that arrives is for the same peer in the same space and carries the same thing. I have taken the narrower point from the other thread and stated that accepting a response consumes the outstanding request, so a repeated or delayed copy has nothing left to answer.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,305 @@
...
+
+Each space keeps, in memory, the set of peer URLs it has completed the
+handshake with.
+
+- The initiator marks a URL complete when it receives a response to a request
+  it has outstanding. A response that does not answer an outstanding request is
+  dropped, so that a peer cannot get itself marked complete without having
+  received the initiator's agents.
+- The responder marks a URL complete when it receives a request, because the
+  response it sends back carries its own agents.
```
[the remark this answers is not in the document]
synchwire: Good catch, this was a real contradiction. Fixed by moving where the responder completes.

The responder now marks a URL complete once it has sent its response, not when the request arrives. The request is applied either way, because the sender's agents are worth recording regardless, but the exchange is only complete once the responder's own agents are on their way back. If it has nothing to send, or the send fails, the URL stays incomplete and it handshakes again next time it has something to send. The zero-agent edge case now follows from that rule rather than contradicting it.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,305 @@
...
+**Both sides begin a handshake at the same time.** Both requests are processed
+and both responses are sent. The exchange is idempotent, so the duplicate work
+is harmless.
+
+**A space is torn down and re-created while the connection persists.** The
+restarted space handshakes again the next time it wants to send, and the peer
+that kept its state accepts the repeat. Until that happens, the peer still
+believes the handshake is complete and its traffic is discarded on arrival, so
+recovery waits for the restarted space to have something to send. That is a
+narrower window than the problem this design removes, but it is not zero.
```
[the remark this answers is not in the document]
synchwire: This is the same finding as the one you raised earlier on this PR and then withdrew, so I will hold the same position.

The limitation is documented rather than fixed. A receiver-side trigger on the blocked inbound path means answering traffic from a peer we know nothing about, which is the one thing the exemption is deliberately narrow about, and it contradicts the handshake being lazy. Sender-side detection needs a signal that the current design does not produce.

The first blocked message is discarded, and the caller that sent it retries on its own schedule, which is already the stated failure behaviour throughout.

Whether that residual window is acceptable is a call for the maintainer, not something I want to close by inventing a mechanism.

docs/design/space-hello-handshake.md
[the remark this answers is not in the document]
synchwire: Applied. The state section now says that accepting a response consumes the outstanding request, so a repeated or delayed copy has nothing left to answer and is dropped, as is any response that never had a matching request.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,348 @@
...
+provides a default implementation that all nodes can use; a host may substitute
+its own, but is not expected to have to.
+
+It is tracked once between a space and a remote peer URL, not once per agent.
+Kitsune2 identifies remote peers by URL throughout, and the block enforcement it
+feeds is expressed per URL, so the handshake is tracked the same way.
+
+One message, carried as a module message within the space under a reserved
+module identifier, and encoded as protobuf so that it can be extended later: a
+**hello**, carrying the sending space's signed agent infos.
```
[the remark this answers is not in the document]
synchwire: Valid. Added: the bound is a property of the protocol rather than of either side, so a sender can tell in advance what will be accepted, and when its local agents do not all fit it includes as many as do.

Not adding chunking or state shared across several hellos. A hello exists to make the receiver aware that somebody it does not block is at that URL, which one agent info satisfies; the full membership of the space is not what a hello is for, and the sender's remaining agents reach the peer by the routes they already do. One hello is always one whole message, so there is no partially-sent state to track.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,348 @@
...
+altered the decision.
+
+### Handshake state
+
+Each space keeps, in memory, the set of peer URLs it has sent a hello to.
+
+A URL is marked once the hello has been sent successfully, and not before. If
+the send fails, the URL stays unmarked and the space sends a hello again the
+next time it has something to send to that peer. Nothing else gates the mark,
+because there is no acknowledgement to wait for.
```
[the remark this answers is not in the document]
synchwire: Valid, and the consequence was worse than the edge case admitted. A successful send means the transport accepted the message, not that the peer received it, so a hello could be lost with the connection carrying it while the URL stayed marked — every later message to that peer then silently discarded, with nothing left to trigger a repair.

Fixed by clearing a URL from the sent-hello set when the connection to it is lost, whether or not the loss had anything to do with a hello. The next send to that peer sends one again over the new connection. That is also the right behaviour for a restart, since a peer reached over a new connection may be one that lost everything it knew.

The edge case about a hello that is never applied now says explicitly that loss in transit is not among the cases it tolerates.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,348 @@
...
+concurrently.** Concurrent sends may be interleaved on the connection, and the
+triggering message arriving first would simply be dropped.
+
+**The receiver must finish applying a hello — recording the agents and updating
+whatever state block enforcement consults — before it processes the next message
+from that peer.** A hello that is accepted and then applied in the background
+reintroduces exactly the race the handshake exists to remove: the message behind
+it is checked for permission before the agents have landed, and is dropped.
+Accepting a hello and applying it must be one step from the point of view of
+message processing.
```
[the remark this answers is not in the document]
synchwire: Valid as a gap in the document, though the guarantee does hold in practice today: the iroh transport keeps a single persistent unidirectional stream per connection and holds connections per peer URL, so frames to a URL are FIFO. That is a property of that implementation rather than of the `Transport` contract, and the design was relying on it silently.

Now stated as a third invariant: messages sent to a peer URL must be delivered in the order the space handed them over. It is the only demand the handshake places on a transport.

Not adding an atomic ordered-send operation. Ordered delivery per peer URL is the property that matters, and a dedicated API for it would constrain implementations without providing anything beyond what they already have to guarantee.

docs/design/space-hello-handshake.md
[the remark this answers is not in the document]
synchwire: The document was renamed to `docs/design/space-handshake.md` in a later commit on this branch, which is why the checkout lookup failed.

The additive rule is in the Shape section and still stands after the revision: a hello supplies agents, never removes them, and superseding an existing record follows the same rules that apply to agent information from any other source. A hello carrying no agent infos is dropped.

One thing has changed since this thread: the exchange is no longer a request and a response, just a single one-way hello in each direction, so "repeated exchanges" now means repeated hellos.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+provides a default implementation that all nodes can use; a host may substitute
+its own, but is not expected to have to.
+
+It is tracked once between a space and a remote peer URL, not once per agent.
+Kitsune2 identifies remote peers by URL throughout, and the block enforcement it
+feeds is expressed per URL, so the handshake is tracked the same way.
+
+One message, carried as a module message within the space under a reserved
+module identifier, and encoded as protobuf so that it can be extended later: a
+**hello**, carrying the sending space's signed agent infos.
```
ThetaSinner: Still referring to "hello" here but we've agreed on "handshake" as the terminology to use. It's only really a handshake when both sides send the message but I think the meaning is acceptably clear

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+
+Agent information is additive, exactly as it is when it arrives from any other
+source. A hello supplies agents; it never removes them, and it is not a snapshot
+that supersedes what the receiver already knows. Superseding an existing record
+follows the same rules that apply to agent information generally.
+
+A hello carrying no agent infos cannot serve its purpose and is dropped. So is
+one that is malformed, oversized, or otherwise not understood. Rejecting a bad
+message costs nothing beyond deciding to reject it: there is no reply to
+withhold and no state to unwind.
```
ThetaSinner: If there's no state to unwind, is it useful to state or could that be dropped?

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+
+A space may have more local agents than one hello can carry. The bound is a
+property of the protocol rather than of either side, so a sender can tell in
+advance what will be accepted, and when its agents do not all fit it includes as
+many as do. Nothing is split across several hellos and no partially-sent state
+is tracked: one hello is always one whole message. A subset is enough because
+the receiver needs to know that somebody it does not block is at that URL, not
+the full membership of the space, and the sender's remaining agents reach it by
+the same routes they always did.
```
ThetaSinner: This isn't something we discussed. The handshake message only needs to send local peers so I think we can assume the set doesn't grow too large. If you reach more than 100 local agent infos, you're going to have more serious performance problems than the handshake message size.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+that URL is blocked in this space, hellos are dropped along with everything
+else. The exemption covers "we do not know who is there yet", never "we know who
+is there and have refused them".
+
+One path narrows as a result. Today a peer URL with a blocked agent can still
+deliver agent information, because preflight carries it and preflight is always
+allowed. Afterwards nothing reaches such a URL at all. In practice this changes
+nothing, because a single blocked agent at a URL blocks the whole URL regardless
+of who else is known to be there, so learning of another agent would not have
+altered the decision.
```
ThetaSinner: I don't find this writing to be especially clear. It's trying to describe the difference between stopping communication after a preflight message and stopping after a handshake message. Is that right?

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+
+**Re-send the preflight whenever a space starts.** Preflight is connection
+scoped and not space-aware, so this means re-sending everything about every
+space each time any space starts, and it requires the transport to understand
+spaces. It also does not address a space starting on the far side.
+
+**Make preflight per space inside the transport.** Every transport
+implementation would then have to track handshake state per peer and space and
+implement the exchange itself. Doing it as a module needs no transport changes
+at all, because module messages are already routed generically.
```
ThetaSinner: I didn't consider these, even if you did. Perhaps we should drop them

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+
+**Treat "no agents known at this URL" as permitted rather than blocked.** This
+removes the symptom by removing the enforcement, and it makes blocking
+unreliable in exactly the situation where it is supposed to work. It also
+leaves the underlying problem — that a space may never learn who it is talking
+to — in place.
+
+**Handshake eagerly, on discovery.** Opens connections to peers that may never
+be used, and makes handshake traffic scale with the number of peers known
+rather than with communication actually taking place.
```
ThetaSinner: Didn't consider this either, because it's clearly a bad idea to eagerly connect to every peer on a network.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,379 @@
...
+This is a wire-visible change in two respects: preflight content changes, and a
+new module message type appears within spaces.
+
+It is not interoperable with nodes that do not implement it. Such a node never
+sends a hello and never records the agents in one, so a space on the new node
+has no reliable way to become known to it. Partial success through
+bootstrap discovery is possible but incidental, and should not be relied on.
+
+There is no persisted state associated with the handshake, so there is nothing
+to migrate.
```
ThetaSinner: This is clear to me from the document, this section adds no value for me.

docs/design/space-handshake.md
synchwire: Renamed throughout. The message is now the "handshake message", and "handshake" is kept for the mechanism, the module and the state. There is no "hello" left in the document.

docs/design/space-handshake.md
synchwire: Dropped. It only restated the "there is no reply" point made a few paragraphs above, so it was carrying nothing of its own. The rules about what gets dropped are unchanged.

docs/design/space-handshake.md
synchwire: Removed. That paragraph came from an automated review comment rather than from anything we discussed, and your reasoning is better: a space holding enough local agents to overflow a single message has worse problems than the message size.

The security section still says the receiver should bound message size and the number of agent infos a message may carry. That is a different concern — what an implementation accepts from an unknown peer, rather than what it chooses to send — so I have left it.

docs/design/space-handshake.md
synchwire: Not quite. It is not about preflight versus handshake as the point where communication stops; it is about where agent information travels.

Today that information rides in the preflight, and the preflight is always allowed, so a URL with a blocked agent at it can still deliver agent information. Once it rides in the handshake instead it cannot, because the handshake exemption stops at URLs known to be blocked. The paragraph exists to record that the change closes that path, and that closing it costs nothing, because a single blocked agent blocks the whole URL regardless of who else is known to be there.

Rewritten to say that directly rather than by implication.

docs/design/space-handshake.md
synchwire: Dropped both. Agreed — this section should record what was actually weighed, not alternatives assembled for completeness.

docs/design/space-handshake.md
synchwire: Dropped. With this and the two above gone, the section holds only the two options that were genuinely considered: treating "no agents known at this URL" as permitted, and making the exchange a request and a response.

docs/design/space-handshake.md
synchwire: Section removed.

One thing left alone: the suggested-sections list in `docs/design/README.md` still names Compatibility. It is a suggestion rather than a requirement, and a future design may well have something non-obvious to say about the wire, so it seemed worth keeping there even though this document did not need it. Happy to drop it if you disagree.

docs/design/space-handshake.md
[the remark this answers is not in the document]
synchwire: Reverting this. The paragraph it produced has been removed at the request of a human reviewer: a space's handshake message only carries its own local agents, and a space with enough of those to overflow one message has larger problems than message size. The receiver-side bound on message size and agent-info count remains in the security considerations.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
...
+Preflight is still exchanged, and it is still exempt from block enforcement, but
+for a different reason than it is today. Today it is exempt because it is the
+only way to learn about new agents at a peer URL. Afterwards it is exempt
+because it is not space-scoped: it is exchanged before any space traffic, so
+there is no space in which a block could be evaluated.
+
+What stays in preflight is connection-level concerns whose correct outcome is to
+reject the whole connection, such as protocol compatibility checks. Because
+those checks happen before any space traffic, the handshake never has to
+consider peers running a known-incompatible protocol.
```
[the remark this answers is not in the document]
synchwire: No change. Mixed-version operation is not going to be supported: this is an intentional breaking change, landing as part of the Kitsune2 0.6 releases. Breaking changes are permitted between minor versions and are exactly what a new minor line is for, so a node running an older protocol is not expected to interoperate with one running this design.

That rules out all three suggested options. A version gate and a preflight fallback both exist to keep two protocol versions talking to each other, which is not a property we want here — keeping agent information in the preflight as a fallback would also preserve the exact defect this design removes. Explicit rejection of older peers is closer to the intent, but that is a connection-level protocol compatibility check, and the design already says such checks stay in the preflight and reject the whole connection.

On the test: design documents in this repository deliberately do not describe tests. That belongs in the implementation.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
+# Space handshake
```
mattyg, line 1: To me "handshake" implies a bidirectional protocol. I would prefer a term like "hello", "announce", "introduction", "greeting" - as this is an outgoing message only.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Handshake state
+
+Each space keeps, in memory, the set of peer URLs it has sent a handshake
+message to.
+
+A URL is marked once the handshake message has been sent successfully, and not
+before. If the send fails, the URL stays unmarked and the space sends a
+handshake message again the next time it has something to send to that peer.
+Nothing else gates the mark, because there is no acknowledgement to wait for.
```
mattyg, line 146:
```suggestion
```

seems redundant.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
...
+remove: the message behind it is checked for permission before the agents have
+landed, and is dropped. Accepting a handshake message and applying it must be
+one step from the point of view of message processing.
+
+Both invariants rest on a third, which is a requirement the handshake places on
+the transport rather than on either peer: **messages sent to a peer URL must be
+delivered in the order the space handed them over.** Issuing in order and
+applying in order buy nothing if the path between may reorder. This is the only
+demand the handshake makes of a transport, and it is worth stating plainly
+because nothing else in the design depends on transport behaviour at all.
```
mattyg, line 233: Is this a demand we want or need to make on the transport for all messages? Or do we want to either:
- bundle the actual message inside the 'handshake' message, then both can be processed at once
- Or, add a send_batch for this one case where we need to specify the order strictly for a batch of 2 messages.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
...
+
+## Goals
+
+- Give every space its own handshake with a remote peer, independent of when
+  the underlying connection was established or how long it lives.
+- Ensure that before a space sends anything to a peer, that peer knows at least
+  one agent the space has, so that block enforcement has something to act on and
+  does not fall back to dropping everything.
+- Work correctly when a space starts after a connection already exists, and
+  when a space is torn down and re-created while the connection persists.
```
mattyg, line 53:
```suggestion
- Work correctly when a space starts after a connection already exists
- Recover when a space is torn down and re-created while the connection persists.
```

Seems more accurate since we are explicitly deciding to let the peer keep sending messages that get dropped, until they receive a new handshake.

docs/design/space-handshake.md
mattyg: Actually I think we do want to just make serial processing in the order messages were sent a demand of all transports and all messages.

mattyg:
Just minor tweaks, but I think the approach is sound.

docs/design/space-handshake.md
ThetaSinner: Maybe, and hello is what I called it originally. At the moment, this will only need to be sent by the side initiating a connection but if we want to make the access check efficient then I'd see that getting folded into this message and therefore it will need to be sent back by the accepting side and it becomes  more like a handshake.

docs/design/space-handshake.md
ThetaSinner: Rather than removing this it might be better to write out what this is trying to say implicitly. The mark should really only be applied when we know that the other side has actually received the message. Whether that's a concern at this level depends on the transport that's in use I suppose - but that's what this bit is about I think. Fits with the "handshake or not " discussion above. If it were extended to be a proper handshake then this gap naturally goes away.

docs/design/space-handshake.md
ThetaSinner: That's a great question. I think that in-order delivery demanded of the transport certainly makes this easier to reason about so if that's reasonable for transports other than Iroh, I'd agree with your second comment. It's a protocol break if this does have to change though, or a special case in transport code for a transport that can't guarantee this.

docs/design/space-handshake.md
ThetaSinner: Agreed, that's better stated as you have

docs/design/space-handshake.md
jost-s: How can the sender know that the receiver has received the handshake?

docs/design/space-handshake.md
jost-s: I don't think the third invariant here is relevant to this specification. The trigger message and the handshake message must be sent in order and received in order, but whether or not subsequent messages are sent in order doesn't matter for the space handshake.

Beyond this spec I agree it makes sense to serialize messages as described.

docs/design/space-handshake.md
```diff
@@ -0,0 +1,346 @@
...
+This should not normally occur: discovery is space-scoped, so a sender only
+learns a URL in the context of a space both sides have joined. It can happen
+transiently when the sender is acting on an agent info that is still cached
+after the remote removed the space.
+
+**The peer is unreachable.** This surfaces as an ordinary transport failure and
+the existing handling for unresponsive peers applies. The handshake adds
+nothing.
+
+**The handshake message is sent but never applied.** A successful send is not
```
jost-s: This design is based on unreliable handshake message delivery being an edge case. If this turns out to be a frequent case, the handshake must probably be redesigned as an actual handshake.

docs/design/space-handshake.md
jost-s: I liked "hello message"

jost-s (approved):

PR #621, opened 2026-08-28
fix: Space overrides must not inherit global relay or credentials

ThetaSinner:
Two bugs, one mechanism: `Config::merge_overrides` is a deep per-key merge, so a
space's override silently inherits every field it did not name — the global relay
it never asked for, and credentials issued for servers it is not talking to.

Changes span the iroh transport, `CoreSpaceFactory`, and the core bootstrap module.

## 1. Spaces reconfigure a relay that has not changed

A space overriding nothing is still handed the global config, so
`configure_for_space` re-inserts the relay the transport is already on — from a
config carrying no auth material, which an authenticated relay then refuses.

Nothing looks wrong while this happens: the node holds a valid token and keeps its
allowlist entry alive. It just reaches no peers. A node with a public address never
notices, so a fleet of public nodes stays healthy while every NAT'd client fails.

**Fix:** call the per-space hook only for a space that overrides something, and
treat only a relay that differs from the transport's own as a per-space relay.

## 2. Credentials reach servers they were not issued for

**Relay** — a space naming only its own relay inherits the global
`authMaterialRelayBase64`, and the auth server is derived from the space's relay
URL, so our credential is PUT to a host the space chose, which can replay it
against ours. Fixed by deciding the relay and its auth material as a pair: material
equal to the transport's own was inherited rather than chosen, so it is dropped.

**Bootstrap** — the same hole through the `builder.auth_material_bootstrap`
fallback, with no provenance to read because the global material lives on `Builder`
rather than in the config. `CoreSpaceFactory` clears it when the override names a
different bootstrap server. Material given with no global server of our own was
configured for whatever server the spaces name, so that case keeps it.

Also fixes `unconfigure_for_space` removing the transport's own relay when a space
had named it in order to present auth material of its own.

## Worth a reviewer's eye

- Spaces naming no relay of their own are no longer marked per-space managed, so
  they receive global relay address updates again.
- `TxImp::configure_for_space` no longer fires for spaces that override nothing.
  Signature and default impl unchanged; the trait docs now say so.

## Backport

For a 0.5.1, since 0.5.0 ships in Holochain 0.7.0. No API change, and every hunk
applies to `release-0.5` as-is. `c1d88b9f` (QAD lost on relay re-insert) shares
this root cause and is not on `release-0.5`, so that line likely wants both.

Verified against a production authenticated relay: a NAT'd client that previously
never reached a peer now joins and syncs.

### DIFF
--- a/crates/api/src/transport.rs
+++ b/crates/api/src/transport.rs
@@ -502,9 +502,10 @@ pub trait TxImp: 'static + Send + Sync + std::fmt::Debug {
 
     /// Apply per-space configuration to this transport.
     ///
-    /// Called by the space factory when creating a new space. The transport
-    /// reads any per-space settings it understands from the provided config
-    /// (e.g., a per-space relay URL and auth material) and applies them
+    /// Called by the space factory when creating a space that was given a
+    /// config override, with that override merged over the global config. The
+    /// transport reads any per-space settings it understands from the provided
+    /// config (e.g., a per-space relay URL and auth material) and applies them
     /// internally.
     ///
     /// Default implementation is a no-op.
@@ -602,9 +603,10 @@ pub trait Transport: 'static + Send + Sync + std::fmt::Debug {
 
     /// Apply per-space configuration to this transport.
     ///
-    /// Called by the space factory when creating a new space. The transport
-    /// reads any per-space settings it understands from the provided config
-    /// (e.g., a per-space relay URL and auth material) and applies them
+    /// Called by the space factory when creating a space that was given a
+    /// config override, with that override merged over the global config. The
+    /// transport reads any per-space settings it understands from the provided
+    /// config (e.g., a per-space relay URL and auth material) and applies them
     /// internally.
     ///
     /// Default implementation is a no-op.
--- a/crates/core/src/factories/core_bootstrap/test.rs
+++ b/crates/core/src/factories/core_bootstrap/test.rs
@@ -69,6 +69,36 @@ struct Test {
 
 impl Test {
     pub async fn new(server: &str, auth_material: Option<Vec<u8>>) -> Self {
+        Self::from_builder(Arc::new(Self::builder(server, auth_material))).await
+    }
+
+    /// Build the bootstrap the way `CoreSpaceFactory` does for a space that
+    /// overrides the bootstrap server: the override is merged over the global
+    /// config, and everything the space uses is built from the result.
+    pub async fn new_with_space_override(
+        global_server: &str,
+        space_server: &str,
+    ) -> Self {
+        let space_override = Config::default();
+        space_override
+            .set_module_config(&super::CoreBootstrapModConfig {
+                core_bootstrap: super::CoreBootstrapConfig {
+                    server_url: Some(space_server.into()),
+                    backoff_min_ms: 10,
+                    backoff_max_ms: 10,
+                    ..Default::default()
+                },
+            })
+            .unwrap();
+
+        let builder = Self::builder(global_server, None)
+            .with_config_overrides(space_override)
+            .unwrap();
+
+        Self::from_builder(Arc::new(builder)).await
+    }
+
+    fn builder(server: &str, auth_material: Option<Vec<u8>>) -> Builder {
         let builder = Builder {
             auth_material_bootstrap: auth_material,
             verifier: Arc::new(TestCrypto),
@@ -88,7 +118,10 @@ impl Test {
                 },
             })
             .unwrap();
-        let builder = Arc::new(builder);
+        builder
+    }
+
+    async fn from_builder(builder: Arc<Builder>) -> Self {
         println!("{}", serde_json::to_string(&builder.config).unwrap());
 
         let blocks = builder
@@ -316,4 +349,37 @@ async fn test_should_override_bootstrap_url_per_space() {
     // Creating a space MUST succeed because there we provided bootstrap config for space.
     kitsune.space(kitsune2_test_utils::space::TEST_SPACE_ID, Some(space_config)).await
     .expect("Expected creating a space to succeed after setting bootstrap URL per space");
+}
+
+/// A space that overrides the bootstrap server must push and poll against that
+/// server, not the global one. The existing per-space override test only proves
+/// the override reaches config validation.
+#[tokio::test(flavor = "multi_thread")]
+async fn space_override_targets_the_overridden_bootstrap_server() {
+    let global_srv = TestBootstrapSrv::new(false).await;
+    let space_srv = TestBootstrapSrv::new(false).await;
+
+    let overridden =
+        Test::new_with_space_override(global_srv.addr(), space_srv.addr())
+            .await;
+    let on_space_srv = Test::new(space_srv.addr(), None).await;
+    let on_global_srv = Test::new(global_srv.addr(), None).await;
+
+    let agent = overridden.push_agent().await;
+
+    for _ in 0..5 {
+        if on_space_srv.check_agent(agent.clone()).await.is_ok() {
+            // The poller on the global server has had the same wall clock to
+            // see this agent, and must not have.
+            assert!(
+                on_global_srv.check_agent(agent).await.is_err(),
+                "agent info reached the global bootstrap server"
+            );
+            return;
+        }
+
+        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+    }
+
+    panic!("agent info never reached the overridden bootstrap server");
 }
--- a/crates/core/src/factories/core_space.rs
+++ b/crates/core/src/factories/core_space.rs
@@ -74,6 +74,33 @@ impl CoreSpaceFactory {
     }
 }
 
+fn bootstrap_server_url(config: &Config) -> Option<String> {
+    config
+        .get_module_config::<crate::factories::CoreBootstrapModConfig>()
+        .ok()
+        .and_then(|c| c.core_bootstrap.server_url)
+}
+
+/// Whether this space customized the bootstrap server, naming one other than
+/// the server the global auth material was configured to authenticate against.
+///
+/// A config merge hands a space every field its override did not set, so a space
+/// naming only its own server is still handed the global credential. Presenting
+/// it there would let that server replay it against ours.
+///
+/// Global auth material with no global server of our own was configured for
+/// whatever server the spaces name, so it stays.
+fn bootstrap_server_customized(global: &Config, space: &Config) -> bool {
+    let Some(global_url) = bootstrap_server_url(global) else {
+        return false;
+    };
+    let Some(space_url) = bootstrap_server_url(space) else {
+        return false;
+    };
+
+    space_url.trim_end_matches('/') != global_url.trim_end_matches('/')
+}
+
 impl SpaceFactory for CoreSpaceFactory {
     fn default_config(&self, config: &mut Config) -> K2Result<()> {
         config.set_module_config(&CoreSpaceModConfig::default())
@@ -99,8 +126,22 @@ impl SpaceFactory for CoreSpaceFactory {
             // nothing about a space that chose nothing of its own.
             let builder = match config_override {
                 Some(cfg_override) => {
-                    let builder =
-                        Arc::new(builder.with_config_overrides(cfg_override)?);
+                    let mut space_builder =
+                        builder.with_config_overrides(cfg_override)?;
+
+                    if bootstrap_server_customized(
+                        &builder.config,
+                        &space_builder.config,
+                    ) {
+                        // Cleared even when the override brought its own
+                        // material, which would take precedence anyway:
+                        // `Builder` is public and the bootstrap factory is
+                        // pluggable, so the global credential should not travel
+                        // on a space that named someone else's server at all.
+                        space_builder.auth_material_bootstrap = None;
+                    }
+
+                    let builder = Arc::new(space_builder);
                     tx.configure_for_space(space_id.clone(), &builder.config)
                         .await?;
                     builder
--- a/crates/core/src/factories/core_space.rs
+++ b/crates/core/src/factories/core_space.rs
@@ -93,19 +93,20 @@ impl SpaceFactory for CoreSpaceFactory {
         tx: DynTransport,
     ) -> BoxFut<'static, K2Result<DynSpace>> {
         Box::pin(async move {
+            // Let the transport handle any per-space configuration it
+            // understands (e.g., a per-space relay URL for iroh), but only for a
+            // space that overrides something, because the global config says
+            // nothing about a space that chose nothing of its own.
             let builder = match config_override {
                 Some(cfg_override) => {
-                    Arc::new(builder.with_config_overrides(cfg_override)?)
+                    let builder =
+                        Arc::new(builder.with_config_overrides(cfg_override)?);
+                    tx.configure_for_space(space_id.clone(), &builder.config)
+                        .await?;
+                    builder
                 }
                 None => builder,
             };
-
-            // Let the transport handle any per-space configuration it
-            // understands (e.g., a per-space relay URL for iroh).
-            // This is non-blocking; the transport delivers the URL
-            // asynchronously via the space handler's new_listening_address.
-            tx.configure_for_space(space_id.clone(), &builder.config)
-                .await?;
 
             let builder_config = &builder.config;
             let config: CoreSpaceModConfig =
--- a/crates/core/src/factories/core_space/test.rs
+++ b/crates/core/src/factories/core_space/test.rs
@@ -582,3 +582,192 @@ mod configure_for_space_only_when_overridden {
         assert_eq!(configured, vec![TEST_SPACE_ID]);
     }
 }
+
+/// Global bootstrap auth material authenticates the global bootstrap server. A
+/// config merge hands a space every field its override did not set, so a space
+/// naming its own server is still handed our credential - presenting it there
+/// would let that server replay it against ours.
+mod bootstrap_auth_material_is_not_carried_to_another_server {
+    use super::*;
+    use crate::factories::core_space::bootstrap_server_customized;
+    use crate::factories::{CoreBootstrapConfig, CoreBootstrapModConfig};
+
+    const GLOBAL_SERVER: &str = "http://global-bootstrap.example";
+    const SPACE_SERVER: &str = "http://space-bootstrap.example";
+
+    type Seen = Arc<Mutex<Vec<Option<Vec<u8>>>>>;
+
+    #[derive(Debug)]
+    struct NoopBootstrap;
+    impl Bootstrap for NoopBootstrap {
+        fn put(&self, _info: Arc<AgentInfoSigned>) {}
+    }
+
+    /// Records the auth material the space's builder carries by the time the
+    /// bootstrap module is built from it.
+    #[derive(Debug)]
+    struct RecordingBootstrapFactory {
+        seen: Seen,
+    }
+
+    impl BootstrapFactory for RecordingBootstrapFactory {
+        fn default_config(&self, config: &mut Config) -> K2Result<()> {
+            config.set_module_config(&CoreBootstrapModConfig {
+                core_bootstrap: CoreBootstrapConfig {
+                    server_url: Some(GLOBAL_SERVER.into()),
+                    ..Default::default()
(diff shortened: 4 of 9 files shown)


ThetaSinner:
Added the root-cause fix as a second commit.

`CoreSpaceFactory::create` called the per-space transport hook for every space. With no override it passed the global config, presenting the defaults as that space's own — which a transport cannot distinguish, and which the iroh transport acts on by reconfiguring the relay it is already connected to. Holochain correctly passes no override when an app declares none, so this was kitsune2 substituting the global config and labelling it per-space.

The hook is now called only when an override was actually supplied, with the overridden config.

**Both commits are wanted, not alternatives.** An override that changes only the bootstrap URL still carries the default relay URL, so the hook legitimately fires with a config naming a relay that has not changed — which the first commit's guard catches. The second stops the mistake being made; the first stops it being destructive if it is made anyway.

Tests: two in `core_space`, verified to fail without the change and pass with it, plus one confirming the hook still fires for a genuine override. Full suites green — 104 in `kitsune2_core`, 83 in `kitsune2_transport_iroh`, 16 in `kitsune2_bootstrap_srv`.

Still API and wire compatible: `TxImp::configure_for_space` keeps its signature and default implementation. The behaviour change to note in a changelog is that an implementor assuming the hook fires for every space will no longer see it for spaces that override nothing. Both commits cherry-pick cleanly onto `v0.5.0`.

crates/core/src/factories/core_space.rs
```diff
@@ -93,20 +93,26 @@ impl SpaceFactory for CoreSpaceFactory {
...
         Box::pin(async move {
+            // Let the transport handle any per-space configuration it
+            // understands (e.g., a per-space relay URL for iroh), but only when
+            // this space actually overrides something. Calling it with the
+            // global config would present the defaults as if they were this
+            // space's own, and a transport cannot tell the two apart: the iroh
+            // transport would reconfigure the relay it is already connected to.
+            //
+            // This is non-blocking; the transport delivers the URL
+            // asynchronously via the space handler's new_listening_address.
```
ThetaSinner: Don't say in an essay what you can say in one sentence.

Fine to keep the old comment but don't add too much more

crates/transport_iroh/src/tests/url.rs
```diff
@@ -365,3 +367,62 @@ fn own_url_for_preflight_space_relay_takes_precedence() {
     );
     assert_eq!(result, Some(our_space_url));
 }
+
+/// The case that matters in practice: a space overriding nothing is handed the
+/// global config, so its `relay_url` is the relay already in use. Acting on it
+/// re-inserts that relay from a config with no auth material, which an
+/// authenticated relay then refuses for the life of the process.
```
ThetaSinner:
```suggestion
/// A space overriding nothing is handed the
/// global config, so its `relay_url` is the relay already in use. Acting on it
/// would insert that relay as a custom override. At best, that's pointless, at worst it loses the mapping to auth material and results in a broken configuration.
```

crates/transport_iroh/src/url.rs
```diff
@@ -92,3 +92,36 @@ pub(super) fn endpoint_from_url(url: &Url) -> K2Result<EndpointAddr> {
...
+/// The relay a space should be configured with, or `None` to leave the
+/// transport's relay alone.
+///
+/// A space that overrides no relay is still handed the global config, whose
+/// `relay_url` is the relay the transport already connected to. Acting on that
+/// would re-insert a relay that has not changed, and the global config carries no
+/// auth material, so an authenticated connection would be replaced by one the
+/// relay refuses - leaving the node with a live token, a live allowlist entry,
+/// and no relay. It would also mark the space per-space managed, so global
+/// address updates would stop reaching it.
```
ThetaSinner:
```suggestion
```

crates/core/src/factories/core_space.rs
ThetaSinner: Trimmed to the original comment plus one sentence.

crates/transport_iroh/src/tests/url.rs
ThetaSinner: Applied.

crates/transport_iroh/src/url.rs
ThetaSinner: Dropped. The doc is now three lines: what counts as the space's own relay, and why inherited auth material never travels with it.

That last part is new since your review. The merge is a deep per-key merge, so an override that names only `relayUrl` inherits the global `authMaterialRelayBase64`, and `do_insert_relay` derives the auth server from the space's relay URL — so our credential gets PUT to a host the space chose. `per_space_relay` now decides the relay and its auth material as a pair. Same hole existed on the bootstrap side via the `builder.auth_material_bootstrap` fallback; that one is cleared in `CoreSpaceFactory` since there is no provenance to read. See 5b60b438.

crates/api/src/transport.rs
```diff
@@ -502,10 +502,11 @@ pub trait TxImp: 'static + Send + Sync + std::fmt::Debug {
 
     /// Apply per-space configuration to this transport.
     ///
-    /// Called by the space factory when creating a new space. The transport
-    /// reads any per-space settings it understands from the provided config
-    /// (e.g., a per-space relay URL and auth material) and applies them
-    /// internally.
+    /// Called by the space factory when creating a space that was given a
+    /// config override, with that override merged over the global config. A
+    /// space that overrides nothing does not call this. The transport reads
```
ThetaSinner:
```suggestion
    /// config override, with that override merged over the global config. The transport reads
```

Restating the same thing as the previous sentence

crates/core/src/factories/core_space.rs
```diff
@@ -74,6 +74,33 @@ impl CoreSpaceFactory {
...
+/// Whether this space named a bootstrap server other than the one the global
+/// auth material was configured to authenticate against.
+///
+/// A config merge hands a space every field its override did not set, so a space
+/// naming only its own server is still handed the global credential. Presenting
+/// it there would let that server replay it against ours.
+///
+/// Global auth material with no global server of our own was configured for
+/// whatever server the spaces name, so it stays.
+fn bootstrap_server_replaced(global: &Config, space: &Config) -> bool {
```
ThetaSinner: Rather than replaced "customized per space" or something like that would be more in line with existing terminology?

crates/core/src/factories/core_space.rs
```diff
@@ -93,20 +120,43 @@ impl SpaceFactory for CoreSpaceFactory {
         tx: DynTransport,
     ) -> BoxFut<'static, K2Result<DynSpace>> {
         Box::pin(async move {
+            // Let the transport handle any per-space configuration it
+            // understands (e.g., a per-space relay URL for iroh), but only for a
+            // space that overrides something, because the global config says
+            // nothing about a space that chose nothing of its own.
+            // This is non-blocking; the transport delivers the URL
+            // asynchronously via the space handler's new_listening_address.
```
ThetaSinner: Not clear to me how this is relevant here

crates/core/src/factories/core_space.rs
```diff
@@ -93,20 +120,43 @@ impl SpaceFactory for CoreSpaceFactory {
...
+                            .auth_material_bootstrap
+                            .take()
+                            .is_some()
+                    {
+                        tracing::warn!(
+                            ?space_id,
+                            "Not presenting the global bootstrap auth material \
+                             to a server this space named; set auth material on \
+                             the override to authenticate it"
+                        );
```
ThetaSinner: Not sure about this. What if the override doesn't need authentication?

crates/transport_iroh/src/lib.rs
```diff
@@ -1257,17 +1257,49 @@ impl TxImp for IrohTransport {
...
+                    .is_some_and(|c| c.auth_material_relay_base64.is_some())
+            {
+                warn!(
+                    ?space_id,
+                    relay_url = %url,
+                    "Not presenting the global relay auth material to a relay \
+                     this space named; set auth material on the override to \
+                     authenticate it"
+                );
+            }
```
ThetaSinner: Again, what if the override doesn't require auth material? I think it's fine to just accept the configuration the user provided for the override

crates/api/src/transport.rs
ThetaSinner: Applied — the condition is already in the preceding sentence.

crates/core/src/factories/core_space.rs
ThetaSinner: Renamed to `bootstrap_server_customized`, matching the "customized per space" wording used elsewhere (e.g. c1d88b9f's message and the `relay_qad` module docs). Test name followed.

crates/core/src/factories/core_space.rs
ThetaSinner: Dropped. It described `configure_for_space` being non-blocking, which is a transport implementation detail that has no bearing on the decision this comment is explaining.

crates/core/src/factories/core_space.rs
ThetaSinner: Agreed, removed — both here and the equivalent on the relay side. A space naming a server that needs no authentication is the normal case, so warning on it would fire for most overrides, and the project's logging conventions reserve `warn` for something unexpected or actionable. We just accept the override's configuration now.

crates/transport_iroh/src/lib.rs
ThetaSinner: Removed, same as the bootstrap one. The material is dropped silently and the override is taken as given.

crates/core/src/factories/core_space/test.rs
```diff
@@ -432,3 +432,342 @@ async fn broadcast_new_agent_info_on_resign() {
...
+        fn create(
+            &self,
+            builder: Arc<Builder>,
+            _peer_store: DynPeerStore,
+            _space_id: SpaceId,
+        ) -> BoxFut<'static, K2Result<DynBootstrap>> {
+            self.seen
+                .lock()
+                .unwrap()
+                .push(builder.auth_material_bootstrap.clone());
```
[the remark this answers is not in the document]
ThetaSinner: Applied. Also fixed the two other mutex locks this branch added, in the `configure_for_space_only_when_overridden` module.

crates/transport_iroh/src/url.rs
```diff
@@ -92,3 +92,53 @@ pub(super) fn endpoint_from_url(url: &Url) -> K2Result<EndpointAddr> {
...
+    };
+
+    if !differs && own_auth_material.is_none() {
+        return None;
+    }
+
+    Some(SpaceRelay {
+        url: url.to_string(),
+        auth_material_base64: own_auth_material.map(str::to_string),
+    })
```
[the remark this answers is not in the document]
ThetaSinner: Confirmed and fixed. `unconfigure_for_space` called `remove_relay` on whatever URL the space held, and my auth-only branch is what first lets `space_relays` hold the global relay — so releasing that space would drop the relay everything else is on.

`unconfigure_for_space` now skips `remove_relay` when the URL is the transport's own, via a shared `is_transport_own_relay` helper that `per_space_relay` also uses. The space's keepalive is still aborted either way, since the material went with the space.

Two regression tests in `src/tests/relay_lifecycle.rs`, against a local authenticated relay, asserting on the endpoint's relay map: releasing a space that named our own relay keeps it, and releasing a space that named its own removes it. Both mutation-checked.

Worth noting the first assertion I wrote for this — that the listening address is unchanged — passed with the fix reverted, so it was not testing anything. Asserting on the relay map directly was the only reliable signal; on loopback a lost relay is masked by iroh connecting directly.

crates/transport_iroh/src/tests/race_window.rs
```diff
@@ -0,0 +1,374 @@
...
+    let a_url_slot: Arc<Mutex<Option<Url>>> = Arc::new(Mutex::new(None));
+    let a_addr_seen = Arc::new(tokio::sync::Notify::new());
+    let a_disconnect_seen = Arc::new(AtomicBool::new(false));
+    let a_mock = {
+        let url_slot = a_url_slot.clone();
+        let addr_seen = a_addr_seen.clone();
+        let disconnect_seen = a_disconnect_seen.clone();
+        Arc::new(MockTxHandler {
+            new_listening_address: Arc::new(move |url| {
+                *url_slot.lock().unwrap() = Some(url);
```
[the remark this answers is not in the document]
ThetaSinner: This file was committed to the branch by mistake — `race_window.rs` was an untracked file already in the working tree, unrelated to this change, and a `git add -A` swept it in. It has been removed from the branch and is untracked again, so these three comments no longer apply to anything in the diff.

crates/transport_iroh/src/url.rs
```diff
@@ -92,3 +92,36 @@ pub(super) fn endpoint_from_url(url: &Url) -> K2Result<EndpointAddr> {
...
+                != trim_trailing_slash(transport_relay_url)
+        }
+        // No relay of our own, so anything the space names is its own.
+        None => true,
+    };
+    differs.then(|| space_relay_url.to_string())
+}
+
+fn trim_trailing_slash(url: &str) -> &str {
+    url.strip_suffix('/').unwrap_or(url)
```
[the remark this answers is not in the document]
ThetaSinner: Addressed, though the helper it referred to is gone. The single-slash `trim_trailing_slash` was replaced by `trim_end_matches('/')` inline, and that comparison now lives in `is_transport_own_relay`, shared by `per_space_relay` and `unconfigure_for_space`. Covered by `is_transport_own_relay_ignores_trailing_slashes` and `per_space_relay_ignores_trailing_slash_differences`, both of which include a repeated-slash case.

crates/core/src/factories/core_space.rs
```diff
@@ -93,20 +93,26 @@ impl SpaceFactory for CoreSpaceFactory {
...
             let builder = match config_override {
                 Some(cfg_override) => {
-                    Arc::new(builder.with_config_overrides(cfg_override)?)
+                    let builder =
+                        Arc::new(builder.with_config_overrides(cfg_override)?);
+                    tx.configure_for_space(space_id.clone(), &builder.config)
+                        .await?;
+                    builder
                 }
                 None => builder,
```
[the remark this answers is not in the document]
ThetaSinner: Addressed by revising the contract rather than restoring the call. The docs on both `TxImp::configure_for_space` and `Transport::configure_for_space` now state that the hook fires only for a space given a config override.

Restoring the call for every space would reinstate the bug: the global config says nothing about a space that chose no configuration of its own, and a transport handed it cannot tell the two apart. That an implementor may have relied on the old contract is called out in the PR body as a behaviour change.

crates/core/src/factories/core_space/test.rs
```diff
@@ -432,3 +432,157 @@ async fn broadcast_new_agent_info_on_resign() {
...
+                })
+            })
+        }
+
+        fn configure_for_space(
+            &self,
+            space_id: SpaceId,
+            _config: &Config,
+        ) -> BoxFut<'_, K2Result<()>> {
+            self.configured.lock().unwrap().push(space_id);
```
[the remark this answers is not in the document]
ThetaSinner: Applied here and to every mutex lock this branch added — four in total, across both test modules in `core_space/test.rs`.

crates/transport_iroh/src/tests/race_window.rs
```diff
@@ -0,0 +1,374 @@
...
+            break false;
+        }
+        let res = tx_a
+            .send_space_notify(
+                b_url.clone(),
+                TEST_SPACE_ID,
+                Bytes::from_static(b"loop"),
+            )
+            .await
+            .map_err(|e| e.to_string());
```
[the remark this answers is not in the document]
ThetaSinner: Not applicable — `race_window.rs` was committed to this branch by mistake. It was an untracked file already in the working tree, unrelated to this change, and a `git add -A` swept it in. It has been removed from the branch and is untracked again, so nothing in the diff corresponds to this comment.

crates/transport_iroh/src/tests/race_window.rs
```diff
@@ -0,0 +1,374 @@
...
+        scenario,
+        ?time_to_cleanup,
+        cleanup_observed,
+        total_sends_in_window = total,
+        send_oks,
+        send_errs,
+        ?first_err_at,
+        ?last_err_at,
+        loop_payloads_received,
+        silent_drops = send_oks.saturating_sub(loop_payloads_received),
```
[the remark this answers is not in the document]
ThetaSinner: Not applicable — `race_window.rs` was committed to this branch by mistake. It was an untracked file already in the working tree, unrelated to this change, and a `git add -A` swept it in. It has been removed from the branch and is untracked again, so nothing in the diff corresponds to this comment.

crates/core/src/factories/core_bootstrap/test.rs
```diff
@@ -317,3 +350,36 @@ async fn test_should_override_bootstrap_url_per_space() {
...
+
+    let overridden =
+        Test::new_with_space_override(global_srv.addr(), space_srv.addr())
+            .await;
+    let on_space_srv = Test::new(space_srv.addr(), None).await;
+    let on_global_srv = Test::new(global_srv.addr(), None).await;
+
+    let agent = overridden.push_agent().await;
+
+    for _ in 0..50 {
```
jost-s: Why 50 times?

crates/core/src/factories/core_space.rs
```diff
@@ -93,20 +120,30 @@ impl SpaceFactory for CoreSpaceFactory {
...
                 Some(cfg_override) => {
-                    Arc::new(builder.with_config_overrides(cfg_override)?)
+                    let mut space_builder =
+                        builder.with_config_overrides(cfg_override)?;
+
+                    if bootstrap_server_customized(
+                        &builder.config,
+                        &space_builder.config,
+                    ) {
+                        space_builder.auth_material_bootstrap = None;
```
jost-s, line 141: I don't get why this is set to no auth material.

crates/core/src/factories/core_bootstrap/test.rs
jost-s, line 1: The PR description only mentions relays. Why are the bootstrap tests added here?

crates/core/src/factories/core_space.rs
```diff
@@ -74,6 +74,33 @@ impl CoreSpaceFactory {
...
+        .ok()
+        .and_then(|c| c.core_bootstrap.server_url)
+}
+
+/// Whether this space customized the bootstrap server, naming one other than
+/// the server the global auth material was configured to authenticate against.
+///
+/// A config merge hands a space every field its override did not set, so a space
+/// naming only its own server is still handed the global credential. Presenting
+/// it there would let that server replay it against ours.
```
jost-s, line 89: Say what?

crates/core/src/factories/core_space.rs
```diff
@@ -74,6 +74,33 @@ impl CoreSpaceFactory {
     }
 }
 
+fn bootstrap_server_url(config: &Config) -> Option<String> {
```
jost-s, line 77: Why is the bootstrap server code changed here? I thought it was about relays.

cdunster (dismissed):

crates/core/src/factories/core_bootstrap/test.rs
ThetaSinner: Good point, it passes just fine after one sleep and so 5 is plenty like the other tests

crates/core/src/factories/core_space.rs
ThetaSinner: Added a code comment. This is a global field that we don't read on override and it's important that we don't mix auth material between bootstrap servers

crates/core/src/factories/core_bootstrap/test.rs
ThetaSinner: I found similar problems with the override logic for bootstrap and decided to fix it

crates/core/src/factories/core_space.rs
ThetaSinner: Auth material is not provably tied to an agent like membrane proofs are supposed to be. If we don't make efforts to avoid sending default auth material to overridden servers then it would be possible for an app to capture your default auth material and use it.

crates/core/src/factories/core_space.rs
ThetaSinner: As above, I found the identical logical fault with configuration handling for bootstrap that I came to fix for relays

PR #628, opened 2026-09-03
fix(transport_iroh): preserve sends during simultaneous open

veeso:
## What changed

Two Kitsune2 nodes can end up dialing each other at almost the same moment. When that happens, one of the two connections has to be closed and the other kept, and the node has to be sure which one won before it starts sending real data on it. This branch fixes several ways that decision could still go wrong and cause a message to be silently dropped instead of being sent (or retried) on the connection that actually survived.

## Why

The main fix already on this branch made sending wait for that decision to be resolved before writing data, instead of writing onto a connection that might be about to be torn down. This branch closes gaps that fix itself introduced, and fixes a few more found by going through the whole change looking for anything similar it might have missed:

- The rule for which connection wins sometimes kept the wrong one alive, specifically after a node disconnects and immediately reconnects to the same peer.
- When a connection lost the race, the sender was sometimes told the send had succeeded when it had not, instead of being told to retry on the winning connection.
- A send could still fail with an error instead of retrying, in a couple of narrower timing windows found only by close inspection: while a preflight message was still being written, and while the outgoing data itself was still being written.
- The reverse mistake was also happening in one place: a connection that failed for a real reason (not because it lost the race) was being reported to the other side as if it had merely lost the race, which made a genuine failure look like a harmless retry case.

A new automated test was added at the level where two full nodes talk to each other over the real network stack, so this class of bug gets caught by our own test suite instead of only showing up as flakiness in the separate wind-tunnel test suite ([wind-tunnel#690](https://github.com/holochain/wind-tunnel/issues/690)).

The internal bookkeeping for which connection is currently active per peer was also pulled out into its own, separately tested module, so the tricky part of this logic can be tested on its own without needing a real network connection.

## Notes

One narrow case is still possible and was deliberately left as is: a peer can still drop a message that was already fully sent, if that peer decides at the last moment to keep a different connection instead. The sender has no way to find out, since the send itself succeeded. Fixing this properly would need a change to the network protocol itself (an acknowledgement from the receiving side), which is out of scope for this branch. It is now far less likely to happen than before this fix: in testing, this branch lost about one message per 250 simultaneous-connection attempts, compared to losing a message on the very first attempt every single time before this fix. If the wind-tunnel flakiness mentioned above does not fully go away after this lands, this is the remaining known cause and would need its own follow-up.

## Test plan

- `cargo make static` passes (formatting, lint, docs)
- The full workspace test suite passes
- The three tests that previously failed or covered this bug each pass 10 out of 10 runs in a row
- The new test that reproduces two nodes sending to each other at the same time was confirmed to fail against the old code and pass against the fixed code

### DIFF
--- a/crates/kitsune2/tests/integration.rs
+++ b/crates/kitsune2/tests/integration.rs
@@ -3,7 +3,7 @@ use kitsune2::default_builder;
 use kitsune2_api::{
     BoxFut, Builder, Config, DhtArc, DynKitsune, DynSpace, DynSpaceHandler, Id,
     IncomingOp, K2Result, KitsuneHandler, LocalAgent, OpId, SpaceHandler,
-    SpaceId, Timestamp,
+    SpaceId, Timestamp, Url,
 };
 use kitsune2_core::{
     Ed25519LocalAgent,
@@ -22,7 +22,8 @@ use kitsune2_transport_iroh::{
     IrohTransportFactory,
     config::{IrohTransportConfig, IrohTransportModConfig},
 };
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
 
 fn create_op_list(num_ops: u16) -> (Vec<IncomingOp>, Vec<OpId>) {
     let mut ops = Vec::new();
@@ -36,29 +37,66 @@ fn create_op_list(num_ops: u16) -> (Vec<IncomingOp>, Vec<OpId>) {
     (ops, op_ids)
 }
 
-#[derive(Debug)]
-struct TestKitsuneHandler;
+/// A [KitsuneHandler] that records every notification received by any of
+/// its spaces, so tests can assert on what was actually delivered.
+#[derive(Debug, Default)]
+struct TestKitsuneHandler {
+    received: Arc<Mutex<Vec<Bytes>>>,
+}
 impl KitsuneHandler for TestKitsuneHandler {
     fn create_space(
         &self,
         _space_id: SpaceId,
         _config_override: Option<&Config>,
     ) -> BoxFut<'_, K2Result<DynSpaceHandler>> {
-        Box::pin(async {
-            let space_handler: DynSpaceHandler = Arc::new(TestSpaceHandler);
+        let received = self.received.clone();
+        Box::pin(async move {
+            let space_handler: DynSpaceHandler =
+                Arc::new(TestSpaceHandler { received });
             Ok(space_handler)
         })
     }
 }
+impl TestKitsuneHandler {
+    /// A snapshot of everything this node has been notified of so far.
+    fn received(&self) -> Vec<Bytes> {
+        self.received.lock().expect("poison").clone()
+    }
+}
 
 #[derive(Debug)]
-struct TestSpaceHandler;
-impl SpaceHandler for TestSpaceHandler {}
+struct TestSpaceHandler {
+    received: Arc<Mutex<Vec<Bytes>>>,
+}
+impl SpaceHandler for TestSpaceHandler {
+    fn recv_notify(
+        &self,
+        _from_peer: Url,
+        _space_id: SpaceId,
+        data: Bytes,
+    ) -> K2Result<()> {
+        self.received.lock().expect("poison").push(data);
+        Ok(())
+    }
+}
+
+/// The gossip settings used by every test node except where a test needs
+/// something different, such as suppressing gossip-initiated connections.
+fn default_test_gossip_config() -> K2GossipConfig {
+    K2GossipConfig {
+        initiate_interval_ms: 1000,
+        min_initiate_interval_ms: 100,
+        initiate_jitter_ms: 100,
+        round_timeout_ms: 10_000,
+        ..Default::default()
+    }
+}
 
 async fn make_kitsune_node(
     relay_server_url: &str,
     bootstrap_server_url: &str,
-) -> DynKitsune {
+    gossip_config: K2GossipConfig,
+) -> (DynKitsune, Arc<TestKitsuneHandler>) {
     let kitsune_builder = Builder {
         #[cfg(feature = "transport-iroh")]
         transport: IrohTransportFactory::create(),
@@ -93,24 +131,18 @@ async fn make_kitsune_node(
     kitsune_builder
         .config
         .set_module_config(&K2GossipModConfig {
-            k2_gossip: K2GossipConfig {
-                initiate_interval_ms: 1000,
-                min_initiate_interval_ms: 100,
-                initiate_jitter_ms: 100,
-                round_timeout_ms: 10_000,
-                ..Default::default()
-            },
+            k2_gossip: gossip_config,
         })
         .unwrap();
 
-    let kitsune_handler = Arc::new(TestKitsuneHandler);
+    let kitsune_handler = Arc::new(TestKitsuneHandler::default());
     let kitsune = kitsune_builder.build().await.unwrap();
     kitsune
         .register_handler(kitsune_handler.clone())
         .await
         .unwrap();
 
-    kitsune
+    (kitsune, kitsune_handler)
 }
 
 /// For iroh transport, the relay functionality is integrated into the bootstrap server.
@@ -158,6 +190,18 @@ async fn start_space(kitsune: &DynKitsune) -> DynSpace {
     space
 }
 
+/// The transport URL a node advertises, once the relay has assigned one.
+async fn peer_url_of(space: &DynSpace) -> Url {
+    let mut url = None;
+    iter_check!(10_000, 200, {
+        if let Some(current) = space.current_url() {
+            url = Some(current);
+            break;
+        }
+    });
+    url.expect("the space must learn its own URL")
+}
+
 #[tokio::test]
 async fn two_node_gossip() {
     enable_tracing();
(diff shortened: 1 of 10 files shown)


crates/kitsune2/tests/integration.rs
```diff
@@ -738,3 +798,148 @@ async fn two_spaces_different_relays() {
...
+/// peers dial at once, one of the two connections loses the deterministic
+/// tie-break, and whichever payload was written to the losing connection was
+/// silently dropped.
+///
+/// The race is only hit when both sides' first dial lands in the same narrow
+/// window, so a single pair of nodes is not a reliable reproduction: this
+/// runs several independent pairs, each a fresh "first contact", to make the
+/// race close to certain to be hit at least once if the bug is present.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn simultaneous_space_notify_is_delivered_both_ways() {
```
veeso, line 832: this is the test proving the fix is implemented.

crates/transport_iroh/src/lib.rs
```diff
@@ -971,19 +975,50 @@ impl IrohTransport {
                 max_frame_bytes: self.config.max_frame_bytes,
             });
 
+            // Publish the candidate before sending the preflight, so that a
```
jost-s:
```suggestion
            // Register the candidate before sending the preflight, so that a
```

jost-s (commented):
On a first pass this looks correct and an improvement over what was there. I'll give it a second read.

crates/transport_iroh/src/connection_context.rs
```diff
@@ -310,10 +320,23 @@ impl ConnectionContext {
     /// through the identity-aware cleanup path, which sees that this is not the
     /// active connection and so does not fire `peer_disconnect`.
     pub(super) fn close_quietly(&self) {
```
mattyg: Would it be clearer if this was renamed `close_superseded`?

crates/transport_iroh/src/connection_context.rs
```diff
@@ -506,8 +532,19 @@ impl ConnectionContext {
             }
             ReaderCleanup::Quiet => {
```
mattyg: Should we split this into 2 enum variants? `Superseded` and `Inactive` or similar?

crates/transport_iroh/src/lib.rs
```diff
@@ -1130,100 +1164,108 @@ impl TxImp for IrohTransport {
...
+                        info!(remote = ?remote_url.peer_id(), "Establishing connection to remote");
+                        self.create_connection_and_context(
+                            remote.clone(),
+                            remote_url.clone(),
+                        )
+                        .await?
+                    }
+                };
+
+                match ctx.wait_for_resolution().await {
```
mattyg, line 1203: I think there's an edge case here where we don't yet have the peer url for the winning connection. Then we retry the send which establishes another connection, and go through this potentially another 4 times until the retry limit is exhausted.

If we can listen to changes to the `ConnectionRegistry`, then we can wait for the winning connection to actually have a peer url, or else hit the same timeout and drop it.

So anywhere that we discover a connection is superseded, we would then wait for the winning connection to actually have a peer url, or else reach the timeout.

crates/transport_iroh/src/lib.rs
```diff
@@ -239,6 +241,12 @@ pub mod test_utils;
 mod tests;
 
 const ALPN: &[u8] = b"kitsune2/0";
+/// How many times `send` will re-resolve a connection before giving up.
+///
+/// Each attempt is one full dial-and-preflight round; a peer that keeps losing
+/// simultaneous-open races is either genuinely flapping or being restarted, and
+/// the caller is better served by an error than by an unbounded loop.
+const MAX_SEND_CONNECT_ATTEMPTS: usize = 5;
```
mattyg: This seems like a different change outside the scope of fixing the simultaneous open issue. Do we want to be automatically retrying sends at the transport layer? I'm not sure, but it deserves its own careful thought. It could be a quick way to improve reliability, but likely by papering over underlying bugs.

In this PR I think it covers up that we're not handling a simultaneous-open edge case (see other comment).

crates/transport_iroh/src/tests/connect_failure.rs
```diff
@@ -343,3 +343,333 @@ async fn marks_unresponsive_when_outer_connect_timeout_fires() {
...
+    }
+}
+
+/// A genuine preflight-write failure (the connection's `open_uni` fails, not
+/// a simultaneous-open supersede) must close the connection with the real
+/// failure reason, not `CloseCode::Superseded`. A remote observing
+/// `Superseded` treats it as a signal to retry against the connection that
+/// "won", which is wrong here: this connection just failed on its own.
+#[tokio::test]
+async fn genuine_preflight_write_failure_closes_with_real_reason_not_superseded()
```
mattyg, line 721: There are some edge cases we might want to have test coverage for:

- We picked a winner before we have a peer url for it 
- A connection gets superseded while a frame is being written over it

crates/transport_iroh/src/connection_registry.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// connection's own reader observes the close and exits quietly.
+    fn close_superseded(&self);
+}
+
+/// The live connection for each peer URL, and the rules for which connection
+/// that is.
+#[derive(Debug)]
+pub(crate) struct ConnectionRegistry<E> {
+    entries: Arc<RwLock<HashMap<Url, Arc<E>>>>,
+    changes: watch::Sender<u64>,
```
mattyg, line 177: Should we make this a map so we can listen for changes to a single connection?

crates/transport_iroh/src/lib.rs
```diff
@@ -1074,6 +1103,31 @@ impl IrohTransport {
     }
 }
 
+async fn wait_for_send_replacement(
+    connections: &Connections,
+    remote_url: &Url,
+    superseded: &Arc<ConnectionContext>,
+    wait_timeout: Duration,
```
mattyg: Should this be an instant, so that it shares the same timeout, rather than extending the timeout each time it is called?

crates/transport_iroh/src/connection_registry.rs
```diff
@@ -0,0 +1,372 @@
...
+            match entries.get(peer) {
+                Some(current) if Arc::ptr_eq(current, entry) => {
+                    entries.remove(peer);
+                    true
+                }
+                _ => false,
+            }
+        };
+        if removed {
+            self.notify_changed();
```
mattyg: I don't think we need to listen for when a connection is removed? We just want to know when a new one is added so we can actually send over it, right?

crates/transport_iroh/src/lib.rs
```diff
@@ -1130,100 +1183,112 @@ impl TxImp for IrohTransport {
...
-                    if ctx.register_as_active(&connections, &remote_url) {
-                        ctx
-                    } else {
-                        // Our dial lost the tie-break; discard it (its reader
-                        // then exits quietly) and use the connection that won.
-                        ctx.close_quietly();
-                        connections
-                            .read()
-                            .expect("poisoned")
+            loop {
```
mattyg, line 1202: Ah I thought we would be able to get rid of looping here, but I think I was mistaken. Now that we don't have the retry limit I think it can continue indefinitely if we keep churning connections and each gets a new winner. Maybe we need a new transport config for a `send_timeout` or use the existing `connect_timeout_s`? Is that preferable to the previous fixed loop limit?

crates/transport_iroh/src/connection_registry.rs
veeso: imo it would be more complex without giving much advantage

crates/transport_iroh/src/lib.rs
veeso: fixed, now it receives timeout

crates/transport_iroh/src/connection_registry.rs
veeso: fixed, `remove_if_current` no longer notifies registry waiters.

crates/transport_iroh/src/lib.rs
veeso: fixed without adding further config. The loop is now bounded by the shared 10 second preflight deadline. I dind't add a new `send_timeout` because not necessary, and `connect_timeout_s` is accttually governing quic connection establishment, not this.

crates/transport_iroh/src/lib.rs
```diff
@@ -1074,6 +1103,31 @@ impl IrohTransport {
     }
 }
 
+async fn wait_for_send_replacement(
+    connections: &Connections,
+    remote_url: &Url,
+    superseded: &Arc<ConnectionContext>,
+    deadline: tokio::time::Instant,
+) -> K2Result<Arc<ConnectionContext>> {
+    match tokio::time::timeout_at(
```
mattyg, line 1112: `timeout_at` can complete with the resolved future immediately if available, so if it gets called after the timeout has passed, it would not error. I think this can happen if we have a superseded connection while we are sending the data frame.

I think we just also need to check the timeout manually before this timeout_at.

veeso: already refactored

mattyg: I'm not following -- where do we check the timeout manually before calling timeout_at to avoid this edge case?

veeso: Ah, I see what you mean now. `timeout_at` polls the replacement future first, so an already-available replacement can succeed even if the deadline has passed.

The refactor didn’t address that. I’ll add an explicit deadline check before calling timeout_at.

veeso:
@mattyg fixed timeout request

mattyg (approved):

jost-s (approved):

Issue #638, opened 2026-09-12
validate_accept: "Accept message from wrong peer" is counted as a local error, not peer behavior (plus one unreproducible sighting)


StateofIntent:
**Versions**: `kitsune2_gossip` 0.5.0, `kitsune2_transport_iroh` 0.5.0, `kitsune2-bootstrap-srv` 0.5.1, Holochain 0.7.0. Four conductors on one Linux machine, all using a locally-run `kitsune2-bootstrap-srv` as both bootstrap and relay (`http://127.0.0.1:8892`).

This is **two things**: a code-level question that needs no reproduction, and one field sighting that I could not reproduce and am filing as an observation rather than a bug report. They are separated below so the second does not contaminate the first.

---

## 1. `validate_accept`'s wrong-peer error is attributed to the local node

In `src/respond/accept.rs`, `GossipRoundState::validate_accept` has two adjacent checks that classify their errors differently:

```rust
// accept.rs:284
if self.session_with_peer != from_peer {
    return Err(K2Error::other(format!(
        "Accept message from wrong peer: {} != {}",
        self.session_with_peer, from_peer
    ))
    .into());
}

// accept.rs:293
if self.session_id != accept.session_id {
    return Err(K2GossipError::peer_behavior(format!(
        "Session id mismatch: {:?} != {:?}",
        self.session_id, accept.session_id
    )));
}
```

`handle_gossip_message` in `src/gossip.rs` then splits on the variant:

```rust
// gossip.rs:362
Err(e @ K2GossipError::PeerBehaviorError { .. }) => {
    tracing::error!("Peer behavior error: {:?}", e);
    ... incr_peer_behavior_errors(from_peer) ...
}
Err(e) => {
    tracing::error!("could not respond to gossip message: {:?}", e);   // gossip.rs:378
    ... incr_local_errors(from_peer) ...                                // gossip.rs:383
}
```

So an Accept arriving from a peer other than the one the round was initiated with increments **`local_errors`**, while a session-id mismatch in the same function increments **`peer_behavior_errors`**.

Both look like the same class of fault to me — a message that does not belong to the round it names — so the asymmetry reads as unintentional. If it is unintentional, the practical effect is that a peer sending misdirected Accepts accrues no behavior errors, and the receiving node accrues local ones instead.

[...]

PR #639, opened 2026-09-12
fix: attribute a gossip message from the wrong peer to that peer

StateofIntent:
Rejecting a gossip message that came from a peer other than the one the round was initiated with is currently charged to the receiving node: eight of the nine such checks build the error with `K2Error::other`, so `handle_gossip_message` takes its generic arm and calls `incr_local_errors`. `ring_sector_details_diff.rs` already used `K2GossipError::peer_behavior` and this makes the other eight match it.

It also makes each of those functions internally consistent: the `session_id` mismatch check immediately after each wrong-peer check already uses `peer_behavior` in all eight files, so two consecutive checks on the same message disagreed about whose fault it was.

No test changes were needed. The two tests that assert on this text, `respond::agents::tests::receive_agents_from_wrong_peer` and `respond::busy::tests::busy_from_wrong_peer`, match with `contains`, and `PeerBehaviorError`'s `#[error("Rejected peer behavior - {ctx}")]` keeps the message. `cargo test -p kitsune2_gossip --lib` is 67 passed / 0 failed, and `cargo clippy -p kitsune2_gossip --all-targets` and `cargo fmt` are clean.

I raised this as a question first in #638, where the full site-by-site table is, and offered it in either direction — this is the direction that seemed right to me, but if you would rather `ring_sector_details_diff.rs` change to match the other eight instead, say so and I will swap it for a one-line PR.

**A question for reviewers, and the reason I asked before patching:** `peer_behavior` errors increment a peer's behavior counter, so this changes what the peer meta store records for a peer sending misdirected messages. If that counter feeds anything that eventually acts on a peer, this is a behavioural change and not only a bookkeeping one — I could not tell from reading whether `incr_peer_behavior_errors` has downstream consequences, and you will know. Happy to narrow the change to `accept.rs` alone if the blast radius matters.


StateofIntent:
Good catch from the bot review, and it was right about something that mattered more than "minor": **nothing in the suite pinned the classification**, which is the only thing this branch changes.

Both wrong-peer tests asserted `error.to_string().contains(...)`, and that passes either way — `PeerBehaviorError` renders as `Rejected peer behavior - {ctx}` and the previous error as `K2Error - ...`, but both keep the message in `ctx`, so the substring is present before and after. I had checked that the existing tests would keep passing, and took that as sufficient; it isn't, because a test that passes either way protects nothing.

Addressed in b61a2ac, which adds a `matches!(error, K2GossipError::PeerBehaviorError { .. })` assertion to `receive_agents_from_wrong_peer`, `busy_from_wrong_peer`, and `agents.rs`'s mismatched-session test — the last was already `peer_behavior` before this branch and now says so, so the two adjacent checks are covered symmetrically.

Watched failing rather than assumed: with the production change reverted and the new assertion kept, `receive_agents_from_wrong_peer` fails with

```
Expected a PeerBehaviorError so the dispatcher counts this against the peer
rather than as a local error, got: K2Error(Other { ctx: "Agents message from
wrong peer: ws://test:80/AwUJ... != ws://test:80/F1fG..." })
```

`cargo test -p kitsune2_gossip --lib` is 67 passed / 0 failed, `cargo fmt --check` and `cargo clippy -p kitsune2_gossip --all-targets` clean.

The open question from the PR description still stands and is the one I would most like a maintainer's read on: whether `incr_peer_behavior_errors` feeds anything that acts on a peer, which would make this behavioural rather than bookkeeping. I am happy to narrow the change to `accept.rs` alone, or to flip it to the one-line direction instead, if either is preferable.

ThetaSinner (approved):
That appears to be a valid logic fix. The "true" value is the `from_peer` provided by the network and if they sent us conflicting information in the gossip message then that *is* their mistake. This is rather confusing to read, which is probably why it didn't get written as a peer behavior error in the first place but I can't think of any negative consequences of this change.

jost-s (approved):

PR #642, opened 2026-09-17
fix(core): order known-peer endpoints by advertisement timestamp

light-merlin-dark:
I was testing node recovery against real conductors on 2 physical hosts on my
network. I traced it to two stores that should agree not agreeing.
`CoreKnownPeers` keeps the last endpoint it saw, `MemPeerStore` keeps the newest,
so a cached advertisement in a discovery batch splits them. The symptom presents
as a deliberate block. The solution here orders by advertisement timestamp so the
two agree, including regression tests for both batch orderings.

### The defect

`CoreKnownPeers::Inner` stored `HashMap<AgentId, Option<Url>>` and overwrote the
entry on every `record()` call, so the surviving URL was simply the last one seen
in the batch. Discovery batches can legitimately carry cached advertisements, so
an older advertisement arriving after a newer one replaced the current endpoint
with a superseded one.

`MemPeerStore` orders the same advertisements by `created_at` and keeps the
newest. The two stores could therefore disagree, with the peer store holding the
current endpoint while `CoreKnownPeers` held the old one.

The visible consequence is in access control rather than in peer discovery.
`CorePeerAccessState` removes a decision when `KnownPeers` cannot resolve an
endpoint to an agent, and `CoreSpace` blocks when there is no cached decision.
The peer is then blocked with no deliberate block anywhere. Because `CoreSpace`
increments the same blocked counter either way, this is indistinguishable from an
intentional block when reading the counters.

### The change

Store `(Timestamp, Option<Url>)` and ignore an advertisement that is not newer
than the stored one, matching the peer store's ordering. Equal timestamps keep
the existing entry, which the unit test pins explicitly.

The production diff is confined to `core_known_peers.rs`. The change to
`core_access.rs` is test only.

### Tests

* `core_known_peers.rs`: unit test over equal and newer timestamps, asserting
  which endpoint resolves after each `record()`.
* `core_access.rs`: regression test running **both** batch orderings, newer then
  older and older then newer, asserting the current endpoint retains its access
  decision. This covers the symptom, not only the mechanism.

Verified on this branch, rebased onto `cea2f5b`:

* `cargo fmt --check` clean
* `cargo clippy --workspace --all-targets -- -D warnings`: 0 warnings
* `cargo test -p kitsune2_core`: 114 passed, 0 failed

### Reproduction

Reproduced against published `kitsune2_core` / `kitsune2_api` 0.5.1 components
with real ephemeral signatures, outside the patched tree. Inserting a newer then
an older advertisement for one agent retained the newer endpoint in
`MemPeerStore`, replaced it with the older endpoint in `CoreKnownPeers`, and left
no access decision for the current endpoint.

### Scope

This fixes the endpoint ordering inconsistency only. It is **not** offered as an
explanation for any latency or role convergence behaviour. Those were observed
separately in my own environment and remain unexplained, and I am deliberately
not attaching them to this change.

---

This change was AI assisted, disclosed per `AI_POLICY.md`. The fix and its tests
were produced with an AI coding agent during my own debugging of the failure
above. I understand the change, can explain it, and stand behind it as my own. I
will answer review myself rather than relay.

By Merlin ([rbeckner.com](https://rbeckner.com)) and his AI agent (Codex, GPT-5,
thinking on: high; PR description and commit message: Claude Code, Opus 5).


zippy:
Hey @light-merlin-dark.  Thanks for PR.  I'm curious, what actually put two infos for one agent into a single batch on your network?  Your test does this manually, and I can't see how that would ever happen in the real world.

jost-s:
If an older agent info arrives after a newer agent info, something must be wrong. I see how this is a theoretical inconvenience in a construed case, but fail to understand what the actual scenario is that can lead to this.

mattyg:
@jost-s I think there are probably a bunch of scenarios where this could arise since we're getting agent infos from other peers, not just bootstrap, and those peers might not have the latest agent infos.

I opened a draft PR with an integration test demonstrating it: https://github.com/holochain/kitsune2/pull/643

I'm good with the solution here as the minimal change to fix the issue (though slightly disappointed we have a separate `CoreKnownPeer` store instead of deepening the existing peer store allowing this bug to be possible)

mattyg (approved):

light-merlin-dark:
> Hey @light-merlin-dark. Thanks for PR. I'm curious, what actually put two infos for one agent into a single batch on your network? Your test does this manually, and I can't see how that would ever happen in the real world.

Hey, Zippy. It's a pleasure.

I was directing a lab on my local network to establish feasibility for integrating Holochain into something I'm building. The probe was durability and recovery across two physical hosts: an author writes 30 records plus a small encrypted object, the author stops, a replica restarts keeping its cell, and a fresh cold reader joins and must retrieve all of it.

What the restart produced was two infos for one agent at two different endpoints. At the first exchange involving the cold reader:

```
source       agent        endpoint     created_at (us)
mac-mint     ae51ff7b...  67628ad6...  1789289534164051
cold-reader  ae51ff7b...  b8a355d9...  1789289524914680
```

9.249 seconds apart. The reader was started after the restart and had never seen the pre-restart node, and the author was already stopped, so the only remaining source for the old endpoint is the bootstrap record, which hadn't caught up. On the unmodified run the access decision was removed at `08:33:53.816535Z`, followed by nine consecutive dropped outgoing messages through `08:36:18.296291Z` — 2m25s of silent unreachability, with no later grant and no explicit block recorded. The run failed at the 150-second role barrier.

Your critique of the test is fair, and it doesn't need the batch. `record()` takes a `Vec` and iterates it with no call-boundary awareness, so a two-element batch is exactly two sequential records in list order. I wrote it that way because it makes the sequence deterministic in one call, not because I was claiming that batch arrives on the wire.

But the pair itself — two infos for one agent in circulation at the same time — isn't unusual. kitsune2 mints it on a timer.

**By design, on a schedule.** `CoreSpaceConfig` sets `re_sign_freq_ms: 1000 * 60` and `re_sign_expire_time_ms: 1000 * 60 * 5`, and a signed info gets `expires_at = created_at + Duration::from_secs(60 * 20)`. So `check_agent_infos` re-signs a local agent once its current info is within 5 minutes of expiry — at roughly t+15min — and the previous info stays valid until t+20min. Every agent has two distinct live infos for about 5 minutes out of every 15-minute cycle, each cached in every peer that has seen it. Nothing recalls the old one.

**Immediately, on rebind.** `CoreSpace::new_url()` swaps `current_url` and calls `invoke_cb()` on every local agent, minting an info carrying the new URL while the old one is still in circulation. That is the case my run hit.

**And nothing dedupes them.** `AgentInfoSigned` derives `PartialEq, Eq, Hash` over `agent_info`, `encoded` and `signature`, so two infos for one agent are two distinct values in any collection that gathers them.

That last part is why the batch shape isn't exotic. Holochain's own `SweetConductor::exchange_peer_info` unions `peer_store().get_all()` from every conductor into a `HashMap<Arc<DnaHash>, HashSet<_>>`, then makes a single `.insert(inject_agent_infos)` call per conductor per space. There's no per-agent selection in it — no dedupe by `AgentId`, no `created_at` comparison. It hands the peer store the whole union and relies on the peer store to sort it out. `SweetConductorBatch::exchange_peer_info` runs that in a retry loop (`while !SweetConductor::exchange_peer_info(&self.0).await`, 10ms apart for up to 10s), so after the first pass every conductor holds a copy of everyone's info, and any re-sign or rebind between passes puts the stale copy and the fresh one into the same `Vec`.

I haven't run those tests to failure, and none of them assert on URL-to-agent resolution, so they wouldn't catch this. The narrow point is only that the shape is what the project's own peer-exchange helper produces.

**Which is the actual inconsistency.** `MemPeerStore::insert` forwards the raw list to `known_peers.record()` before any block or expiry filtering, then applies `if a.created_at >= agent.created_at { continue }` to its own store, with a `debug!("Ignoring insert for older agent info")`. `CoreKnownPeers::record` applied none of that — unconditional last-write. Same list, same call, two different selection rules. That branch and that log line in the peer store exist because older-after-newer is expected; this PR only makes the second consumer of the same list obey the rule the first one already enforces.

The tombstone path makes it mandatory rather than incidental. `local_agent_leave` signs a tombstone at `Timestamp::now()` with `url: None`, which by construction is newer than the live info it supersedes — so a cached live info arriving afterwards is older-after-newer every time, and pre-patch it resurrected the dead URL-to-agent mapping.

One caveat on my evidence: the causal chain spans two runs rather than one. The failing unmodified run above predates the advertisement-ordering instrument, so the ordered stale input was captured on later runs.

On the structure, I support @mattyg's framing over mine. `CoreKnownPeers` exists so access control can resolve a URL for an agent the peer store dropped, which is a tombstone requirement rather than a store requirement. A retention flag on `MemPeerStore` would leave one store, one ordering rule and one state, and this class of bug would be unrepresentable. I kept the patch minimal because that seemed out of scope for a PR about an ordering inconsistency, not because I'd defend the current design.

jost-s (approved):
@mattyg 
> @jost-s I think there are probably a bunch of scenarios where this could arise since we're getting agent infos from other peers, not just bootstrap, and those peers might not have the latest agent infos.

Okay, but in many cases it's not a problem if it's not the most recent agent info that's retained. It only makes a difference if the agent's URL has changed. Which is the case in this scenario.

Despite all the words in the issue, the actual problem as it appeared in the real world is still unclear. From what I understand:
- Node A publishes agent info with URL `a`.
- A restarts with a new URL `a'` and publishes it to the bootstrap.
- B comes online and retrieves agent infos from the bootstrap. That would be `a'`.

So that is not it. Since this is about 2 hosts on the same network, the problem would have to be that the second publish to the bootstrap failed and `a'` doesn't get published, or that A's clock is skewed after restart and the timestamp is older than before the restart.

Another more likely scenario involves more nodes on the network:
- A publishes URL `a`
- B comes online and gets `a` from the bootstrap
- A restarts and publishes `a'`
- C comes online and gets `a'` from the bootstrap first, but then gets `a` from B through gossip.

Actually that's not possible, because [gossip filters agents already in the peer store](https://github.com/holochain/kitsune2/blob/main/crates/gossip/src/respond.rs#L209).

Regardless, even though I don't understand how this situation came to be, I agree the fix is correct, the stores must behave identically.

> (though slightly disappointed we have a separate CoreKnownPeer store instead of deepening the existing peer store allowing this bug to be possible)

The alternative would have been to keep blocked agents in the peer store and then change Holochain to filter out blocked agents in every query. We opted for this change in kitsune2.

@light-merlin-dark (or probably rather Claude, given all the side story and unnecessary details)

Please be concise with your comments. It's not helpful to have so much text around a relatively simple problem.

PR #645, opened 2026-09-19
fix(core): Perf Goodies — wake gossip after fetch completion

light-merlin-dark:
## 🍬 Perf Goodies

Hey, good day or evening, everyone. I hope you are doing excellent.

It's my pleasure to contribute this PR, which lets gossip wake when its outstanding fetches finish. In our local arc-growth fixture, **initiation dropped from 3,001 ms to 202 ms** with the existing timers unchanged.

This is part of a morning I spent exploring Kitsune2 with AI coding agents, isolating small changes and testing them against pinned upstream code. The [full lab write-up](https://rbeckner.com/articles/kitsune2-performance-discoveries/) brings together the experiments, results, and trade-offs.

Successful response processing empties the pending map without notifying existing drain listeners. Calling `notify_listeners_if_queue_drained` after removing completed operations restores that signal, just as the error and peer-cleanup paths already do.

This starts the five-part fetch stack. [Focused change and tests](https://github.com/light-merlin-dark/kitsune2/commit/6501d2c116177b35f15b182bfab193a89c02542e). The endpoint-index PR is independent.

### Local proof

At `06555ee` (`0.6.0-dev.1`):

- 5 targeted checks fail on the original code for the missing notification; all 7 new core/gossip tests pass with the patch.
- All 23 existing focused fetch tests pass on both versions. Controls cover partial responses, store errors, multiple listeners, and registration races.
- A fetch held open still reaches the roughly 3-second fallback on both versions; acceptance was independently rerun.

The change adds 1 call to an existing helper. It preserves the configured 200 ms initial delay and 3,000 ms fallback. The measured gain is local gossip initiation, not full synchronization.

Series: [Completion](https://github.com/holochain/kitsune2/pull/645) · [Concurrency](https://github.com/holochain/kitsune2/pull/646) · [Ownership](https://github.com/holochain/kitsune2/pull/647) · [Cleanup](https://github.com/holochain/kitsune2/pull/648) · [Batching](https://github.com/holochain/kitsune2/pull/649) · [Endpoint index](https://github.com/holochain/kitsune2/pull/650).

Run the included regression checks:

```sh
cargo test --locked --release -p kitsune2_core --test k3a_drain_lab
cargo test --locked --release -p kitsune2_gossip --test k3a_drain_initiate
```

### Review follow-up

[...]

PR #646, opened 2026-09-19
fix(core): Perf Goodies — let configured fetch workers send concurrently

light-merlin-dark:
## 🍬 Perf Goodies

Hey, good day or evening, everyone. I hope you are doing excellent.

It's my pleasure to contribute this PR, which makes the configured fetch concurrency available to the workers. With 5 workers configured, our fixture went from **1 send in flight to 5**, while staying within the configured limit.

This is part of a morning I spent exploring Kitsune2 with AI coding agents, isolating small changes and testing them against pinned upstream code. The [full lab write-up](https://rbeckner.com/articles/kitsune2-performance-discoveries/) brings together the experiments, results, and trade-offs.

The receiver lock acquired in the `while let` condition stays alive through the awaited network send. Receiving the next request inside a short block releases that lock before sending, so another worker can collect its request. Queue capacity and worker count stay unchanged.

Fetch stack **2/5**, based on https://github.com/holochain/kitsune2/pull/645. GitHub shows the preceding change too; [this commit isolates the concurrency change and tests](https://github.com/light-merlin-dark/kitsune2/commit/04e616ed023d673f7f63f2ab5587cfb4b80ae50a). Review alongside [the admission-ownership repair](https://github.com/holochain/kitsune2/pull/647) before landing worker concurrency.

### Local proof

At `06555ee` plus the completion-notification repair:

- Paused-clock barriers reproduce 4 overlap failures on the baseline; all 9 lab tests pass with the patch.
- Configurations of 1, 2, and 5 workers reach their expected overlap without exceeding their bounds.
- All 23 existing focused fetch tests pass on both versions; acceptance was independently rerun.

**Review consideration:** overlapping sends make peer-wide failure cleanup able to remove another in-flight request's pending entry. The companion admission-ownership patch narrows that cleanup. This demonstrates worker overlap; real-transport throughput remains unmeasured.

Series: [Completion](https://github.com/holochain/kitsune2/pull/645) · [Concurrency](https://github.com/holochain/kitsune2/pull/646) · [Ownership](https://github.com/holochain/kitsune2/pull/647) · [Cleanup](https://github.com/holochain/kitsune2/pull/648) · [Batching](https://github.com/holochain/kitsune2/pull/649) · [Endpoint index](https://github.com/holochain/kitsune2/pull/650).

Run the included regression checks:

```sh
cargo test --locked --release -p kitsune2_core --test k3b_concurrency_lab
```

### Review follow-up

Inherited the completion-test review fixes and applied the mutex convention and fixture documentation to the concurrency harness. All 9 concurrency tests pass; strict Clippy passes.

The changelog-preview job attempts to fetch the fork branch from `holochain/kitsune2`; that workflow checkout failure is separate from these code changes.

AI agents wrote the patch and test harnesses under my direction and helped draft this description and the commit messages.

Merlin ([rbeckner.com](https://rbeckner.com)) · Lab: ZCode and π, with GPT-6 Astra and GLM-5.3 · Review and refinement: Codex, GPT-6 Astra (thinking: high).


crates/core/src/factories/core_fetch.rs
```diff
@@ -321,9 +321,17 @@ impl CoreFetch {
...
         transport: WeakDynTransport,
     ) {
-        while let Some((op_id, peer_url)) =
-            outgoing_request_rx.lock().await.recv().await
-        {
+        loop {
+            // Receive the next owned request inside a short lock scope, then
+            // process it without holding the shared receiver guard so the
+            // configured parallel workers can overlap their sends.
+            let request = {
```
veeso, line 328: I think this block could just have been 

```rust
let Some((op_id, peer_url)) = {
...
} else { 
  break;
}
```

and then you can remove the let statement below.

veeso (commented):
Thank you very much for your contribution!

It looks good to me, and freeing the lock makes sense there rather than keeping it for the whole cycle.

I want to wait for another review from the team before approving. Also, I'm not sure we need so many tests for this, and the file names are a bit weird.

jost-s (changes requested):
I also think the production code modification is helpful, approved.

Far too many tests. I think the first two in `core_fetch.rs` are valuable, because they test the added line. For the rest of the tests I would need some argument why they have to be added.

light-merlin-dark:
No, after more consideration, I agree. It should be one focused test. We were excited and wanted to submit the labs to prove the gains and trade-offs, but in the process we learned that it's better for the tests to be focused, and that the lab belongs outside the project.

I've reduced this PR to the fix plus one test, `parallel_workers_send_requests_concurrently`, in the existing `outgoing_request_queue.rs`, and taken @veeso's `let ... else` suggestion. It's written as a single expression, because `let ... else` can't end in a block. I've trimmed the rest of the series the same way. Thank you both for the careful review.

light-merlin-dark:
No, after more consideration, I agree. It should be one focused test. We were excited and wanted to submit the labs to prove the gains and trade-offs, but in the process we learned that it's better for the tests to be focused, and that the lab belongs outside the project.

I've reduced this PR to the fix plus one test, `parallel_workers_send_requests_concurrently`, in the existing `outgoing_request_queue.rs`, and taken @veeso's `let ... else` suggestion. It's written as a single expression, because `let ... else` can't end in a block. I've trimmed the rest of the series the same way. Thank you both for the careful review.

